Issue #31, opened 2015-02-10
Effects not working in safari


miguelcobain:
> @miguelcobain When I click a button, I see the background color change within a rounded rect and no animation.

@BryanHunt do you see any difference between http://miguelcobain.github.io/ember-paper/#/button and https://material.angularjs.org/#/demo/material.components.button ?

altrim:
I think @BryanHunt is using Safari. On the latest safari I am checking the effects are not working on buttons and same for switches, checkboxes and radio buttons are broken.

miguelcobain:
Ember-paper or angular material or both?

altrim:
Ember-paper, angular material works fine on safari

miguelcobain:
Ok, thanks.

Don't have my mac around right now. Will try to have a look when i do.

BryanHunt:
Yep, I'm on a Mac using Safari :) Yep, effects seem to be broken.  I see the effects on the Angular button but not the paper button.  IMHO, the wave effect on the materialize site is better than the Angular one.

miguelcobain:
Ok, thanks. Don't know what may be happening. :confused: 
Side note: the discussion should be which effect is closer to specs (I don't know that answer).

BryanHunt:
Just to close out the side note.  The Angular version seems to match the spec.  I'm all for going with the spec.

btecu:
Not just the effects but most of the things seem broken in Safari:

![screen shot 2015-02-10 at 7 23 35 pm](https://cloud.githubusercontent.com/assets/7517745/6139484/60a5c34e-b15a-11e4-8f4f-32def8c6d353.png)

![screen shot 2015-02-10 at 7 23 48 pm](https://cloud.githubusercontent.com/assets/7517745/6139485/66e68d7e-b15a-11e4-86d0-84e06ec8a26d.png)

BryanHunt:
I think the problem is more fundamental than effects.  It looks like the CSS is very broken in general on Safari - there are no headers on the sample application at all.  Looks great under Chrome.

miguelcobain:
I could use some help in finding this bug. What kind of headers are missing?

BryanHunt:
Here is a screenshot of what I get when I open http://miguelcobain.github.io/ember-paper/

![ember-paper_-_the_ember_approach_to_material_design_](https://cloud.githubusercontent.com/assets/504299/6181652/cacba160-b2fb-11e4-841c-ed1e23da0893.png)

[...]

PR #38, opened 2015-02-13
Applying autoprefixer. Fixes miguelcobain/ember-paper#31.

green-arrow:
One of the main issues causing safari to not function properly was the lack of using the `-webkit` vendor prefix for flexbox layouts. Adding autoprefixer and setting it to use the last 2 browsers should ensure that `ember-paper` works properly in Safari as well as IE 10.

See miguelcobain/ember-paper#31 for additional details / discussion.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
     "url": "https://github.com/miguelcobain/ember-paper/issues"
   },
   "devDependencies": {
+    "autoprefixer": "^5.1.0",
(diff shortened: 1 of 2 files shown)


green-arrow:
After looking further into this, I believe that this resolves all of the current issues with Safari, so disregard the 'one of the main issues' comment I made above.

salzhrani:
Woot

PR #39, opened 2015-02-13
update documentation

chenxsan:
I updated my ember-paper package to version 0.0.10, obviously something is broken, there is no more components like `paper-drawer`, the documentation is outdated.


miguelcobain:
It was a breaking change, yes. Thanks.

PR #40, opened 2015-02-13
Autoprefixer

salzhrani:
Most if not all issues on Safari are because vendor prefixes. Autoprefixer takes care of that in the best way possible


miguelcobain:
Duplicate of #38?

salzhrani:
Oops ...

PR #46, opened 2015-02-18
Add feature for Icon sizes and spinners using classes

mhretab:
using classes to create multiple sizes of `paper-icon` and css animation for spinners.

the added classes are:
`.md-lg`  for large icon (1.5em)
`.md-2x` 2x default icon
`.md-3x` 3x default icon
`.md-4x` 4x default icon
`.md-5x` 5x default icon

`.md-spin` clockwise spinner
`.md-spin-reverse` anti-clockwise spinner

### DIFF
--- a/addon/components/paper-icon.js
+++ b/addon/components/paper-icon.js
@@ -4,7 +4,37 @@ export default Ember.Component.extend({
   tagName: 'md-icon',
   classNames: ['paper-icon'],
   classNameBindings: ['iconClass'],
+
+  spin: false,
+  reverseSpin: false,
+
   iconClass: function(){
-    return 'ic-'+this.get('icon');
+    var iconClasses = 'ic-'+this.get('icon');
+
+    if(this.get('spin')){
+      iconClasses +=' md-spin';
+    } else if (this.get('reverseSpin')){
+      iconClasses +=' md-spin-reverse';
+    }
+
+    switch(this.get('size')){
+      case 'lg':
+        iconClasses += ' md-lg';
+        break;
+      case 2:
+        iconClasses += ' md-2x';
+        break;
+      case 3:
+        iconClasses += ' md-3x';
+        break;
+      case 4:
+        iconClasses += ' md-4x';
+        break;
+      case 5:
+        iconClasses += ' md-5x';
+        break;
+    }
+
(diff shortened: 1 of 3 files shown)


mhretab:
I can also add `classNameBindings` if you think that is better. So, instead of
 ```handlebars
{{paper-icon icon="rotate-right" class="md-spin"}}
``` 
it will be something like this 
```handlebars
{{paper-icon icon="rotate-right" spin=true}}
```

miguelcobain:
Yes, please.

Similar to this addon approach: https://github.com/lfridael/ember-cli-font-awesome#options

mhretab:
actually there is a way we can use it without class bindings. The classes can be added to the `icon` like this:
```handlebars
{{paper-icon icon="rotate-right md-spin"}}
```
Also I think that people are used to that, and it will be much easier to migrate to paper-icon from material design icon fonts. Besides that the additional code is no worth it since it will just append the classes and `icon="rotate-right md-lg"` will do the same.

miguelcobain:
With class bindings I can do things like:

```handlebars
{{paper-icon icon="rotate-right" spin=isSpinning}}
```

or 

```handlebars
{{paper-icon icon="rotate-right" size=size}}

Enter size: {{input type="text" value=size}}
```

Basically binding to properties on my controller.

mhretab:
alright :+1: I am on it.

mhretab:
Added class name bindings.

addon/components/paper-icon.js
```diff
@@ -4,7 +4,37 @@ export default Ember.Component.extend({
...
+      case 4:
+        iconClasses += ' md-4x';
+        break;
+      case 5:
+        iconClasses += ' md-5x';
+        break;
+    }
+
+    return iconClasses;
   }.property('icon')
```
miguelcobain: computed property is only depending on `icon`.

I think it makes sense to add separate this logic into different computed properties like `sizeClass` and `spinClass` and then add them do the classNameBindings like `classNameBindings: ['iconClass', 'sizeClass', 'spinClass']`.

mhretab:
yeah.. you are right. I've made a new commit according to that.

miguelcobain:
Looks good. Thanks!

PR #51, opened 2015-03-01
paper-text should pass the 'required' attribute

nightexcessive:
paper-text does not currently pass the `required` attribute to the underlying `input`. This PR resolves that by simply passing that attribute to the `input`.


miguelcobain:
Thanks!

PR #55, opened 2015-03-05
Add animations for showing and hiding backdrop

### LINKED ISSUE #53: fix backdrop out animation
Currently the backdrop disappears instantly instead of fading out animation

alexlafroscia:
Adds the fade-in and fade-out animations to the `paper-backdrop` component.  From what I can tell, it matches the way that it works on the Material Angular website.

Closes #53.

### DIFF
--- a/addon/components/paper-backdrop.js
+++ b/addon/components/paper-backdrop.js
@@ -2,9 +2,18 @@ import Ember from 'ember';
 
 export default Ember.Component.extend({
   tagName: 'md-backdrop',
-  classNames:['paper-backdrop','md-opaque','md-default-theme'],
+  classNames: ['paper-backdrop','md-opaque','md-default-theme'],
 
-  click: function(evt){
+  willDestroyElement: function() {
+    var clone  = this.$().clone();
+    this.$().parent().append(clone);
+    clone.addClass('ng-leave');
+    Ember.run.later((function() {
+      clone.remove();
+    }), 200);
+  },
+
+  click: function(evt) {
     Ember.$(evt.target).trigger('collapseSidenav');
     return false;
   }
--- a/app/styles/paper-backdrop.scss
+++ b/app/styles/paper-backdrop.scss
@@ -1,8 +1,18 @@
+@mixin animateFadeIn() {
+  -webkit-animation: $swift-ease-out-timing-function mdBackdropFadeIn 0.5s both;
+  animation: $swift-ease-out-timing-function mdBackdropFadeIn 0.5s both;
+}
+
+@mixin animateFadeOut() {
+  -webkit-animation: $swift-ease-in-timing-function mdBackdropFadeOut 0.2s both;
+  animation: $swift-ease-in-timing-function mdBackdropFadeOut 0.2s both;
+}
+
 md-nav-container {
 
   &.sidenav-expanded {
     md-backdrop {
-      animation: $swift-ease-out-timing-function mdBackdropFadeIn 0.5s both;
+      @include animateFadeIn();
     }
   }
 }
@@ -29,11 +39,14 @@ md-backdrop {
   right: 0;
   bottom: 0;
 
+  @include animateFadeIn();
+
   &.ng-enter {
-    animation: $swift-ease-out-timing-function mdBackdropFadeIn 0.5s both;
+    @include animateFadeIn();
   }
+
   &.ng-leave {
-    animation: $swift-ease-in-timing-function mdBackdropFadeOut 0.2s both;
+    @include animateFadeOut();
   }
 }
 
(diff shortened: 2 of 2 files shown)


miguelcobain:
Thank you for the effort! 
There must be other way to do this. I really don't like cloning the element before removing the original. :/
Maybe @ef4 can guide us here and on #53.

Would love to see liquid fire here. :)

alexlafroscia:
Oh, I figured that we didn't want to introduce a new dependency.  I researched it quite a bit and the answer I found in multiple posts and articles was to clone the element and do it exactly like this.  I would be happy to take a look at Liquid Fire instead, if you want to go that route.

ef4:
I don't think there's anything wrong with the cloning -- liquid fire takes the same approach sometimes.

But I do think it would be good to just use liquid-fire, because it provides a strong convention for how to organize animations in Ember.

miguelcobain:
I've never seen liquid fire in an add-on. How would we define the transitions.js file?

ef4:
I'm happy to help make that work. I don't think it would require significant changes.

But we'll want to figure out how to keep the transition rules scoped nicely so that it's hard to accidentally collide with user provided rules.

miguelcobain:
That would definitely be a powerful feature for liquid-fire!

Meanwhile, in this PR i just feel 'hacky' cloning an element to be used 200ms, removing the original and then waiting a timeout to remove the clone.

I wish I had a promise or event that we could catch when the animation ends.
After that we can safely remove the backdrop.
Maybe transitionEnd event or perhaps switching to velocity instead of css animations.

ef4:
Yes, one of the worst things about CSS transitions is that you can have scenarios where `transitionEnd` never fires at all (like if your DOM gets destroyed before the animation finished). I have found it much easier to do animations in javascript, for that specific reason.

Hopefully as the Web Animations API matures that will also be a good alternative to get accelerated animation with reliable callbacks. But velocity is very fast, and even beats CSS3 transforms under many conditions.

hhff:
perhaps `ember-paper-liquid-fire` could be a secondary package to ember-paper?

I like the ideas behind liquid fire, but I find myself using lower level techniques for animation, because I'm super particular about how animation works in almost everything I build.

Ember Paper is awesome, but I'd probably not use it if it meant Liquid Fire was compulsory.

There's a little library out there called [Ember Animate](https://github.com/gigafied/ember-animate) that I use most often in conjunction with Velocity or Animate.js (from @bendc) - it adds some useful lifecycle hooks to views & components.  @gigafied - thoughts here?

ef4:
@hhff your objection is based on a common misunderstanding of what liquid-fire is for. It is *not* a set of prebaked animations. It happens to ship with a handful of premade animations, but those are just a starting point. I'm actively trying to encourage more people to go beyond those and design their own detailed, low-level animation implementations. And if there's any way we can improve the API to make that even easier, please let me know.

I am obviously biased as the author of liquid-fire, but it seems clear to me that liquid-fire is a strictly more powerful place to put your animations. It can do things that you fundamentally cannot do if you're stuck implementing the animation from inside a single view. The official video of my emberconf demo is not online yet, but there's a clip floating around that may give some taste: https://www.youtube.com/watch?v=WZmZYLQ96r4&feature=youtu.be

hhff:
@ef4 - I saw your talks at EmberNYC & Wicked Good Ember - and I absolutely understand the bigger picture workflow implications around it.  I think it's awesome - I'm all for quick workflow, and separation of concerns.  That's the Ember way.

I've tried to use it at least a few times and each time found that it feels like such an abstraction on what's actually happening to the DOM that I end up resorting to something that is a bit closer to the browser metal to get the results I'm after.

I've got a new project coming up soon that's super rich in interaction.  I'll definitely give Liquid Fire another crack and see how it feels.

I'd be interested to see what an `ember-paper` branch might look like with liquid fire, but I'm still not convinced this is the right place for it.

hhff:
@ef4 - just got to 11.25 in that youtube vid.  That's pretty awesome

gigafied:
ember-animate doesn't and will never aspire to do what liquid-fire does. liquid-fire is a very cool project, but it really does ask people to commit hook, line and sinker.

The only other way of doing this besides cloning the element is to override the destroy method, like I do in ember-animate. I'm sad that Ember doesn't have better lifecycle hooks natively.

miguelcobain:
This is such a simple fix: https://github.com/miguelcobain/ember-paper/commit/cab2e0ce5adafde8fc0e7c33c72f67baa49b4724

I think we'll stick with this. Toughts?

alexlafroscia:
I'm sad that this PR wasn't necessary, but @miguelcobain's approach is definitely less code.  I'll just go ahead and close this then.

miguelcobain:
Your PR allowed us to have an interesting discussion. It was useful! Thanks!

alexlafroscia:
Hey, so I noticed that the implementation that we went with, just using CSS to handle fading out, triggers a fade-out even when dismissing the sidebar by adjusting the width of the window.  See below.

http://cl.ly/a8DZ

miguelcobain:
Wow, nice work finding this rather extreme scenario. Something to do with classes.

alexlafroscia:
I know you didn't like the clone-the-element approach, but it does match the way it works on material.angularjs.org

miguelcobain:
What makes you say that?
I can only see it using angular's `$animate` [here](https://github.com/angular/material/blob/master/src%2Fcomponents%2Fsidenav%2Fsidenav.js#L212).

alexlafroscia:
Oh, I only meant that the clone-the-element approach doesn't show the darkened background when resizing the window.

miguelcobain:
This little thing fixed it.

alexlafroscia:
Awesome :)

PR #56, opened 2015-03-06
Implement switch dragging

### LINKED ISSUE #29: Add support for switch dragging
Reference: https://github.com/angular/material/blob/master/src/components/switch/switch.js

alexlafroscia:
Here's what I'm doing to try to get switch dragging working.  At the moment, the following works:

- Click anywhere on the switch to toggle the value (matches Material Angular behavior)
- Drag the switch more than halfway from one side to the other to toggle it.  From what I can tell, the Angular implementation doesn't quite do this right.

***

Closes #29

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -2,7 +2,8 @@
   "predef": [
     "document",
     "window",
-    "-Promise"
+    "-Promise",
+    "Hammer"
   ],
   "browser": true,
   "boss": true,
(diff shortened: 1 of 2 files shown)


alexlafroscia:
Since @miguelcobain just mentioned Hammer.js to me, I will take a look at that and see if I can make the implementation a little cleaner by using it.

alexlafroscia:
Alright, I moved over using Hammer for the events and that helped clean up the code quite a bit.  It seems to work the same way that the Angular code works, and the code was based pretty heavily on their implementation.

For some reason the tests failed due to an issue in Bower.  I'm not sure how to request that they get run again, but they all passed for me locally.

alexlafroscia:
There we go :)

miguelcobain:
This is so awesome, @alexlafroscia !!!

PR #63, opened 2015-03-08
[Don't Merge] Paper Select

hhff:
Hi @miguelcobain !

Paper Select working nice: http://cl.ly/1e3c1M1R0R2G

However - I'm not sure how you have the SCSS setup for theming and such.  Would you be willing to tidy up the `paper-select.scss` file?

I also don't know what you want to do for documentation - but happy to add to that here too :smile_cat:

### DIFF
--- a/addon/components/paper-switch.js
+++ b/addon/components/paper-switch.js
@@ -1,3 +1,5 @@
+/* globals Hammer */
+
 import Ember from 'ember';
 import BaseFocusable from './base-focusable';
 import RippleMixin from '../mixins/ripple-mixin';
--- a/addon/components/paper-item.js
+++ b/addon/components/paper-item.js
@@ -1,9 +1,21 @@
 import Ember from 'ember';
+import RippleMixin from '../mixins/ripple-mixin';
 
 export default Ember.Component.extend({
   tagName:'md-item-content',
   classNames:['paper-item'],
+  classNameBindings: ['active'],
   click: function() {
     this.sendAction('action', this.get('param'));
+  },
+
+  active: Ember.computed('param', 'activeParam', function() {
+    return this.get('param') === this.get('activeParam');
+  }),
+
+  willInsertElement: function() {
+    if (this.get('ripple')) {
+      RippleMixin.apply(this);
+    }
   }
 });
--- a/addon/components/paper-select.js
+++ b/addon/components/paper-select.js
@@ -3,80 +3,50 @@ import Ember from 'ember';
 export default Ember.Component.extend({
   tagName:'md-select',
   classNames: ['paper-select'],
-  classNameBindings: ['active', 'valid'],
+  classNameBindings: ['active', 'selection:valid'],
   active: false,
-  selectOptions: null,
-  $select: null,
-  $view: null,
-  activeOption: null,
-  promptOption: null,
+  prompt: null,
+  content: null,
+  optionValuePath: null,
+  optionLabelPath: null,
+  selection: null,
 
-  valid: Ember.computed('activeOption', function() {
-    if (this.get('hasPrompt')) {
-      return this.get('activeOption') !== this.get('promptOption');
+  displayString: Ember.computed('selection', function() {
+    var promptString = this.get('prompt') || "Select an option";
+    var selection    = this.get('selection');
+    if (!selection) { return promptString; }
+
+    var optionLabelPath = this.get('optionLabelPath');
+    var selectionString;
+    if (optionLabelPath) {
+      var relativeOptionLabelPath = optionLabelPath.replace(/^content\.?/, '');
+      selectionString             = this.get('selection.'+relativeOptionLabelPath);
     } else {
-      return true;
+      selectionString = selection.toString();
     }
+    return selectionString;
   }),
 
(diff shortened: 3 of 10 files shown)


addon/components/paper-switch.js
```diff
@@ -1,3 +1,5 @@
+/* globals Hammer */
```
hhff: This isn't related to `paper-select`, just here cause I noticed it in the lint!

miguelcobain:
@hhff This is awesome. Looks super!

Things I don't like / would like to improve:
- Using an inner Ember select for maintaining the selection logic and then hiding it. I think we shouldn't go this route. I think we can implement these things easily without it.
- Clicking an option should trigger the ripple effect on the option. Good news is that because of the way we encapsulated the ripple effect, it is fairly easy to add it to a component. When we iterate and print the options in the template, we probably want to create several `{{paper-option}}` components that have `RippleMixin`. This new `{{paper-option}}` component should probably only be available to the addon (under `addon/`), not the consuming app (under `app/`), since the user may not use it at all.

Did you write the scss yourself?
I usually just copy paste `select.scss` and then `select-theme.scss` on the same file, separated by comments. We don't have themeing for now. We deal with this separation later.

For the documentation, it's all under the dummy app. We just need to create another route and a simple demo there, along with the templates. Just follow the existing demos.

Thanks again for this effort. This is a very important component that was missing in ember-paper and we definitely have a VERY good starting point.

hhff:
Cool!

Re: Select
- I tend to agree - and I'll have a go at cutting it out.  That said, I've written a bunch of these for Ember, and my experience has always been that writing it from scratch introduces unforseen complications, when Ember Select is rock solid to begin with.

Re: Ripple Effect
- I agree!  However I feel like `{{paper-option}}` should be exposed to the Application developer - as it could also be used for items in the sidenav.  Thoughts?  Perhaps it could be called ```{{paper-list-item}}```?

Re: SCSS
Yup - I wrote it myself.  Where exactly are you copy / pasting the SCSS from?

Re: Docs
Too easy.  Will add to this.

miguelcobain:
In this case from https://github.com/angular/material/blob/master/src%2Fcomponents%2Fselect%2Fselect.scss and https://github.com/angular/material/blob/master/src%2Fcomponents%2Fselect%2Fselect-theme.scss. The tricky part is replacing the new color placeholders, etc. I often go back in history because I don't understand them yet.
Then I try to make the template render something similar to what angular material renders so the styles don't break up. Also it makes it easier to bring angular material's style updates and fixes to ember paper and vice versa.

I really like that idea for using {{paper-option}} in the sidenav! We already have `{{paper-item}}` which is an item of a `{{paper-list}}`. If the styles are the same. I thought paper-option would only be used inside the paper-select, as a "private" component.

hhff:
Ok @miguelcobain !  Got another Q.

Update: Culled the `Ember.Select` and got the menu part working nicely with `{{paper-list}}` & `{{paper-item}}`.

I'm now working on bringing the SCSS over.  

My Q is:  How strictly does the SCSS need to mimic the angular/material SCSS?  I think the SCSS for [Angular's Material Select](https://material.angularjs.org/#/demo/material.components.select) is poorly written.  

For example - open their select, and then resize the window.  It:
- Locks the window for scrolling
- Loses it's horizontal position

It also relies on Javascript to do the initial positioning.  My implementation is Pure CSS, and isn't effected by scrolling or window resizes.

Should I try and port it 1:1, or just do my best to make use the same selectors and SCSS vars where possible?

miguelcobain:
You're right. I don't like it either. Let's do it our own way for this.

I was writing a comment but ended up creating another issue for this: #65

Let me know what you think.

hhff:
So I just pushed up my WIP code for reference here - but let's continue the discussion at the issue before we do any more work :smile:

addon/components/paper-switch.js
alexlafroscia: I added it to the global `.jshintrc` file in c80fb32, so we don't have to add it to individual files in the future.

miguelcobain: @alexlafroscia HammerJS is AMD compliant. I think the way to go would be to import it like [this](http://www.ember-cli.com/managing-dependencies/#standard-amd-asset). Meanwhile, I think explicitly adding `/* globals Hammer */` is a reminder for us to do it. Later we just replace it with an ES6 `import`.

alexlafroscia: Oh okay, I didn't know that.  I'll update the places I've used Hammer to import it the right way.

brandonjmckay:
:+1: on being able to specify content explicitly with a `:yield`.

hhff:
Haven't been able to give any time to this unfortunately guys - I won't get a chance to look at it for at least another couple months.  If someone wants pickup where I left off, please do!!

brandonjmckay:
Right now I'm trying to familiarize myself with the codebase, and get a better understanding of what's here. But menus are good. So if I can continue on this I will. I guess I'll need to open a new PR from my branch at that point, so I'll reference this PR from there I guess.

Sound good?

miguelcobain:
Please use the existing `material-menus` branch which follows the rationale one #65.
#65 is the right issue for this discussion.

PR #64, opened 2015-03-08
Fix dismissing sidenav on iOS

### LINKED ISSUE #57: Touch on paper-backdrop doesn't dismiss sidebar on mobile
The `paper-backdrop` component should recognize a tap as well as a click to trigger hiding the sidebar again.

alexlafroscia:
For some reason the default `click` event wasn't working on iOS.  I swapped it
out with a Hammer event and it seems to work much better now.

Closes #57

### DIFF
--- a/addon/components/paper-backdrop.js
+++ b/addon/components/paper-backdrop.js
@@ -4,8 +4,14 @@ export default Ember.Component.extend({
   tagName: 'md-backdrop',
   classNames: ['paper-backdrop', 'md-opaque', 'md-default-theme'],
 
-  click: function(evt) {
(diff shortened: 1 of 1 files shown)


hhff:
thankyou!!  i was about to look into this

miguelcobain:
Cool!

Please add the destroy logic like [ember docs suggest](http://emberjs.com/api/classes/Ember.View.html#event_willDestroyElement).
I did this in paper-switch here https://github.com/miguelcobain/ember-paper/commit/6e3d914963d6b1abd82a6767e6c14e1ec227f283.

alexlafroscia:
@miguelcobain Good catch there, I added it in :+1:

PR #67, opened 2015-03-10
Update dependencies

alexlafroscia:
Updated a bunch of the dependencies, including the Ember CLI.  I checked what was needed to update for addons in version `0.2.0`, and nothing used in Ember Paper requires a changed.


hhff:
:+1: had to update glob locally - this is rad

PR #70, opened 2015-03-10
Add size property to paper-text component

andrewobrien:
Allows size to be adjusted as per normal ember input


peec:
After #170 was merged, size is now possible to to pass with "attr-size". Along with all the other attributes available, closing this

PR #74, opened 2015-04-08
adding a title attribute binding and click handler for paper-icons

SirZach:
I think it would be quite the common scenario to want to put a title attribute and click handler on icons.


SirZach:
bump. any thoughts on this?

miguelcobain:
My first thought was that if you want to click an icon, you should wrap an icon in a button (raised or not).
An icon role is to show an icon. A button role is to be clicked.

Please let me know if you have a use case where this doesn't solve.

SirZach:
@miguelcobain didn't realize i could wrap them inside paper-buttons. thanks a bunch!

PR #77, opened 2015-04-17
Expose changed action on paper-radio

ksin:
Allows you to call an action after a user interaction causes your `selected` value to update. Preferred over adding observers on the `selected` value.

Same as: https://github.com/yapplabs/ember-radio-button


miguelcobain:
I like this!
Thanks.

PR #78, opened 2015-04-21
Fix checking a disabled radio-button

topaxi:
Radio buttons should not be able to be checked while disabled.

### DIFF
--- a/tests/unit/components/paper-radio-test.js
+++ b/tests/unit/components/paper-radio-test.js
@@ -110,7 +110,7 @@ test('it is possible to uncheck and set selected property to null if toggle is t
   assert.equal(component.get('selected'), null);
 });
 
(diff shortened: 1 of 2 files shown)


tests/unit/components/paper-radio-test.js
```diff
@@ -109,3 +109,16 @@ test('it is possible to uncheck and set selected property to null if toggle is t
 
   assert.equal(component.get('selected'), null);
 });
+
+test('it isn\'t possible to check a disabled button', function(assert) {
```
miguelcobain: Rename "button" to "radio button" or "radio"?

topaxi: Done

miguelcobain:
Good catch. Thanks.

PR #79, opened 2015-04-21
Expose actions on checkbox, switch, and text field

ksin:
Similar to: https://github.com/miguelcobain/ember-paper/pull/77

Adding tests for switch and text-field are non-trivial because of the way their events are triggered. Could use a hand.


miguelcobain:
Related #103

DanChadwick:
These have been implemented in the `wip/v1.0.0` branch. Be sure to look at the `changelog.md` in that branch for implementation details. You are looking for `onChange`, `onBlur`, and `onFocus`. Since that's where development is focused, Let's close this on master.

PR #81, opened 2015-04-27
Add ember-prism for better code highlighting

brandonjmckay:
Just tossing a quick example together of using PrismJS for better syntax highlighting in the docs.
Wanted to get your thoughts on if this is ok to add before I go through all the docs and update the code usage.

Thoughts?


miguelcobain:
I've been wanting to do something like this for a long time. Thanks!

Instead of
```html
<code class='language-handlebars'>
  <!-- code -->
</code>
```

what if we create a simple `{{code-block}}` component, making it:
```handlebars
{{code-block lang="handlebars"}}
  {{!-- code --}}
{{/code-block}}
```

It may not be a big deal now, but if we ever need to change the way we render code snippets, we do it in just one place.
What do you think?

brandonjmckay:
Love it. Will do. What do you think about the inline version?

```
{{#code-block lang='handlebars' style='inline'}}woot{{/code-block}}
#or
{{#code-inline lang='handlebars'}}...
```

brandonjmckay:
This is just for the dummy app, right? Wasn't sure how to best include it so it didn't pollute the actual add-on with prism stuff.

miguelcobain:
- To create a dummy app only component just create the `tests/dummy/app/components/code-block.js` file.
- Also, please don't include `prism.js` file directly. Create a bower dependency running `bower install --save prism`. Then, in Brocfile.js, include the file like you would do in a normal app. The included files there will be concatenated with the vender styles/scripts in the dummy app.
- I think an `inline=true` would be better?

brandonjmckay:
Cool. Wasn't sure, I haven't worked on any addons yet and was worried that adding it with bower would add bloat or something to the actual addon.

Not too sure how to add it through bower, as they seem to have a few modules, and a little JS creation tool. I'll look into whether there are separate bower components for the `scss`, and `handlebars` modules which I'm sure we'll want.

brandonjmckay:
After going through it, I'm thinking I might like the separate versions of `{{#code-block}}`, `{{#code-inline}}`. Partly because I couldn't make sense of how to set a dynamic `tagName` for the component.

I'm used to using coffeescript for ember at work, not sure if I was botching the syntax. Everytime I tried to do a computed property like:

```
...
myProperty: function(){
  console.log('woot');
}.property('myOtherProperty'),
...
```

It would botch terribly. So I used the `Ember.computed('propertyName', function(){...` style instead. But I also was having difficulty finding documentation on how to use that with multi property watchers.

package.json
```diff
@@ -38,7 +37,8 @@
     "ember-disable-prototype-extensions": "^1.0.0"
   },
   "dependencies": {
-    "ember-cli-babel": "^4.0.0"
+    "ember-cli-babel": "^4.0.0",
+    "ember-cli-htmlbars": "^0.7.6"
```
brandonjmckay: FYI, this needs to be moved to regular dependencies otherwise cli would just barf at me when I tried to start the app. The debug message told me to move it here.

miguelcobain:
> After going through it, I'm thinking I might like the separate versions of {{#code-block}}, {{#code-inline}}

I'm cool with it.

> So I used the Ember.computed('propertyName', function(){... style instead. 

That's expected. We can't assume that the consuming app will have the prototype extensions enabled. So we just disable them to make sure we don't use them. `ember-disable-prototype-extensions` dependency does just that.

> But I also was having difficulty finding documentation on how to use that with multi property watchers.

Just do like this:

```javascript
myProperty: Ember.computed('prop1', 'prop3', 'prop3', function(){
  console.log('woot');
}),
```

brandonjmckay:
Perfect. Thanks. I'll make those changes when I get the chance

brandonjmckay:
I've started moving some of this out to a new cli addon, cause... why not?
https://github.com/brandonjmckay/ember-prism

I haven't made one from scratch before, so I've just moved the components over so far. Trying to sort out why the dummy app doesn't show any of the changes I tried putting in `tests/dummy/app/templates`. So, WIP.

miguelcobain:
:+1: For the initiative!

brandonjmckay:
It'd be good to use this addon instead of baked in stuff, but I'm not sure I'll be able to figure out how to add only the modules someone wants from it yet.
So not sure how to proceed for it right now.

I guess I could clean up the `ember-prism` addon at least enough to know we have working components there, and then use those here. But we'll need to choose which parts of Prism we want in our own app.

So `ember-prism` wouldn't add the prism library for us. Any thoughts?

miguelcobain:
What exactly is your problem with ember-prism?

brandonjmckay:
Not sure how to automatically include the source file. 

It has some sub-modules which are optional, I haven't figured out whether I can provide some way for people to customize their build without too much tomfoolery.

miguelcobain:
Basic strategy:
- Add `prism` bower package 
- include `prism.js` and `themes/prism.css`
- optional: allow users to customize the theme to use providing an option in Brocfile (tell me if you need an example of this)
- the rest is normal Ember

brandonjmckay:
An option in the Brocfile? Ya, if you have an example of that it would be awesome. Otherwise, I at least now where to look.

What about optional sub modules for stuff like `scss`, `ruby`, and such? I'm wondering if the end-user could just import those modules in their brocfile, and `ember-prism` would import the basics.

brandonjmckay:
Not sure if that's how it works. But that's my thought.

miguelcobain:
In the addon `included` hook you can access Brocfile options.

Example included hook: https://github.com/miguelcobain/ember-cli-selectize/blob/master/index.js
Example usage:
```javascript
//your-app/Brocfile.js

var EmberApp = require('ember-cli/lib/broccoli/ember-app');

var app = new EmberApp({
  'ember-cli-selectize': {
    //valid values are `default`, `bootstrap2`, `bootstrap3` or false
    'theme': 'bootstrap3'
  }
});

module.exports = app.toTree();
```

Just include prism.js bower dependency using `afterInstall` like this: https://github.com/miguelcobain/ember-cli-selectize/blob/master/blueprints%2Fember-cli-selectize%2Findex.js

This dependency will download prism repository. Then just import any required files, depending on the options passed.
I currently don't have the time to help you as I would want, sorry. :cry:

brandonjmckay:
You've been more than helpful already! I'm trying to help the ember-paper movement, just haven't made an addon before. Really appreciate the examples. I'll try to get something up soon.

brandonjmckay:
OK! Cool, I got it moved out to [ember-prism](https://github.com/brandonjmckay/ember-prism), so I'm going to update this PR to reflect that usage now.

The only thing I'm not sure of yet is how to only use that addon only for our test app

brandonjmckay:
I've rebased this onto the latest master now, the only question is whether the addon I've added will only be added to the dummy app.

If that's the case, this should be good to go. If not, it shouldn't be too big of a fix I'd hope.

miguelcobain:
@brandonjmckay What I've noticed is that this dark theme does not really go along with the rest of the overall appearance. I made a PR to ember-prism to allow us to load the default theme. Maybe it fits better with some tweaks and overrids.

Huge respect for taking this opportunity to make your own addon.
The open source community never ceases to amaze me. :)

bower.json
```diff
@@ -11,6 +11,7 @@
     "hammerjs": "~2.0.4",
     "jquery": "1.11.1",
     "loader.js": "ember-cli/loader.js#3.2.0",
-    "qunit": "~1.17.1"
+    "qunit": "~1.17.1",
+    "prism": "*"
```
brandonjmckay: @miguelcobain Should this be in dev dependencies, or will this be ok?

miguelcobain: Dependencies is ok.

brandonjmckay:
@miguelcobain my company usually rebases a branches changes on master, and then squashes the commit to one-line, or keeps the non-crufty descriptive commits.

What do you want to do for ember-paper? Do you have a preference of merging master in here, or rebasing on master?

Do you want me to keep as much history as possible?

miguelcobain:
Rebasing this would be a lot better. :)

brandonjmckay:
Will do gladly.

brandonjmckay:
@miguelcobain this should be good to merge. I checked in a test app, and it didn't add prism to it.

PR #88, opened 2015-05-21
Add paper-slider component

mike1o1:
I combined the paper-slider component my [ember-material-design](https://github.com/mike1o1/ember-material-design) project and added a demo page. 

There are some .scss differences, as I have a `color($color, $type)` method, which tries to mirror the theming support from the angular project. If you'd rather those be simple `map-get` calls, I can refactor.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 # compiled output
 /dist
 /tmp
+.idea
 
 # dependencies
 /node_modules
--- a/addon/components/paper-slider.js
+++ b/addon/components/paper-slider.js
@@ -10,6 +10,7 @@ var MdSlider = BaseFocusable.extend(EventsMixin, {
 
   attributeBindings: ['min', 'max', 'step', 'md-discrete', 'flex', 'tabindex'],
 
+  classNames: ['md-default-theme'],
   classNameBindings: ['isMinimum:md-min', 'active', 'dragging'],
 
   min: 0,
--- a/addon/components/paper-slider.js
+++ b/addon/components/paper-slider.js
@@ -1,0 +1,146 @@
+import Ember from 'ember';
+import EventsMixin from '../mixins/events-mixin';
+import BaseFocusable from './base-focusable';
+
+var MdSlider = BaseFocusable.extend(EventsMixin, {
+
+  constants: Ember.inject.service('constants'),
+
+  tagName: 'md-slider',
+
+  attributeBindings: ['min', 'max', 'step', 'md-discrete', 'flex', 'tabindex'],
+
+  classNameBindings: ['isMinimum:md-min', 'active', 'dragging'],
+
+  min: 0,
+  max: 100,
+  step: 1,
+  tabindex: 0,
+
+  trackContainer: Ember.computed(function() {
+    var element = this.$()[0];
+
+    return this.$(element.querySelector('.md-track-container'));
+  }),
+
+  activeTrackStyle: Ember.computed('percent', function() {
+    var percent = this.get('percent') || 0;
+    return Ember.String.htmlSafe("width: " + (percent * 100) + "%");
+  }),
+
+  thumbContainerStyle: Ember.computed('percent', function() {
+    var percent = this.get('percent') || 0;
+    return Ember.String.htmlSafe("left: " + (percent * 100) + "%");
+  }),
+
+  isMinimum: Ember.computed('percent', 'min', function() {
+    return this.get('percent') === this.get('min');
+  }),
+
+  percent: Ember.computed('value', 'min', 'max', function() {
+    var min = parseInt(this.get('min'), 10);
+    var max = parseInt(this.get('max'), 10);
+
+    return (this.get('value') - min) / (max - min);
+  }),
+
+  positionToPercent(x) {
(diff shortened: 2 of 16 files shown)


miguelcobain:
@mike1o1 Thanks a lot for your effort! This isn't forgotten.

Can you elaborate on the `color` function? I've never really understood how angular material has, for example, `background-color: '{{background-500-0.2}}';`.

I think it would be a huge win to bring this to ember-paper, possibly using that function, especially because we could take the styles from angular material more easily.

I'm just asking so I can convert the rest of the project to this convention.

mike1o1:
The theming code in Angular Material is pretty complex, and I haven't converted all of the functionality either. I don't think it's possible just using sass, but I did my best, which is very close to your implementation.

Your example basically means, find whatever color is defined as '500' for the background, and then add an opacity. That would be converted to sass as:

```sass
background-color: rgba(color($background, '500'), 0.2);
```

A better example might be

```js
background-color: '{{foreground-4-0.5}}';
```

Would be converted to:

```sass
rgba(color($foreground, '4'), 0.5);
```

This would find the variable defined as the $foreground, which would either by light or dark. Then, it would do another lookup into that light or dark table to find the entry for '4', and apply the opacity.

There are still things I'm missing, such as the hues, and better support for contrasts, and I plan on adding better support for those soon.

miguelcobain:
Cool stuff. Understood and definitely desirable.

Another perhaps more complex question is how angular material (or we in the future) tackle `{{accent-color}}`, `{{accent-contrast}}` and `{{warn-color}}`?
Does the same logic apply, like "get 'color' key from the 'accent' table", etc?

I don't see these `accent`, `warn` and `background` palettes defined anywhere in the repo, but I believe they're dynamically generated.

mike1o1:
In [ember-material-design](Ember Material Design) I have `$warn`, `$accent`, etc. defined as a variable in [variables.scss](https://github.com/mike1o1/ember-material-design/blob/bdd77be2ea3e9a0d0a73ce15bd570e79196ce5c7/app/styles/core/variables.scss#L3) which points to a color, which is the way the Angular Material themes are setup. For `$warn`, it's then just a matter of doing the same lookup as `color($warn, '500')`.

For contrast, I added a `contrast` entry to each color table. All the user has to do is do `color($warn, 'contrast')` the same way that they'd look up any other color.

For `{{warn-color}}`, I just make a call to `color($warn)` which defaults to '500' if no shade is passed in. I'll have to double check the Angular theming code to see if that is correct. I think that might actually also generate `.md-hue-1`, `.md-hue-2` and `.md-hue-3` classes. It's been awhile since I read through it.

peec:
+1 for this, need the slider for my current project as well.

benedikt-d:
I'm curious of what the differences between ember-material-design and ember paper are?

peec:
@benedikt-d 

I myself just started investigating and testing both ember-material-design and ember-paper, ember-paper seems much more clean in its implementation. 

I think it's "unfortunate" that there currently are 3 separate projects:
- [ember-paper](https://github.com/miguelcobain/ember-paper) (sources:  [angular material](https://github.com/angular/material) and [WSK material sprint branch](https://github.com/google/web-starter-kit/tree/material-sprint))
- [ember-material-design](https://github.com/mike1o1/ember-material-design) (sources: [angular material](https://github.com/angular/material))
- [ember-materialize](https://github.com/sgasser/ember-cli-materialize) (sources: [Materialize](http://materializecss.com/)


There are also some difference in what paper components that are available from the Material design: See this comparison table made by :
https://github.com/mike1o1/ember-material-design/issues/8#issuecomment-95284499

We should definitely gather everyone to one single repo as it would be more value to everyone including the devs working on these projects. And thus I think ember-paper is the best of the three as a base for further enhancements as its the project that has most traction and a longer commit history. But as of now, all of the 3 projects seems to be actively maintained.

This pull request makes the first way of porting ember-material-design to ember-paper though, so a start would be to merge this in to ember-paper :)

peec:
After more testing, https://github.com/mike1o1/ember-material-design seems more complete imo (and actively maintained). So I guess my pull requests will go for https://github.com/mike1o1/ember-material-design

miguelcobain:
@mike1o1 I finally had the chance to review your PR. Very sorry for the delay.

I just made some changes that can be seen in 7136c91d47b86dc648369cf02a1bf9d490366a54.
I think we don't need a "constants" service (for now). That is very "angular". We can just go with a regular es6 module and avoid the service (minimal) overhead.

Hope to see more features implemented. We have seen many many projects merging through the ember community. This is the right way to open source. I hope to port some components from your repo to this one myself.

mike1o1:
I agree with the constants service. I'll refactor that in my addon as well.

I have a fork of ember-paper with tabs working, but unit tests are failing with the latest Ember. Functionality everything is OK, so I assume it's an issue with the test itself. I'll work on getting that fixed and can submit a PR soon.

miguelcobain:
> I have a fork of ember-paper with tabs working, but unit tests are failing with the latest Ember.

Check my recent commits. I fixed those problems, along with deprecation warnings.
Ideally, you should merge upstream changes to your fork.

mike1o1:
Sorry, I meant that my tests specific to the tabs are failing.

PR #91, opened 2015-05-27
Ember cli 0.2.5

razor-x:
I started working on an update to ember-cli 0.2.5. Right now this is a pretty minimal approach: just compare the diff from `ember update` and adjust appropriately.

There are probably still some things that should be done, either with this PR or soon after. In particualar I can think of the following:
- Move to `ember-cli-sass` as this is the recommenced tool to use for Sass according to the docs. I have an initial attempt in this branch https://github.com/rxfork/ember-paper/tree/ember-cli-sass ~~but when I test with `ember serve`, the css files are 404. I tried following the [Addon Usage section](https://github.com/aexmachina/ember-cli-sass#addon-usage) in the docs but it didn't fix it for me. Maybe it is expecting a different folder structure? Not sure here.~~ Works now and added to PR.
- More dependency bumps? There seem to be some other PRs for those. I'm not sure how you want to handle that.
- Content Security Policy violations for things hosted on github.io (buttons I think). Maybe update the addon's `config/environment.js` to fix these and also add a note to the README alongside the one about the fonts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ $ ember install ember-paper
 Ember-paper uses sass for its styles. To import them run:
 
 ```
-$ npm install --save-dev broccoli-sass
+$ ember install ember-cli-sass
 ```
 
 and then create a file in `app/styles/app.scss` and import the styles at the beginning of your file:
--- a/bower.json
+++ b/bower.json
@@ -4,12 +4,13 @@
     "ember": "1.11.1",
     "ember-cli-shims": "ember-cli/ember-cli-shims#0.0.3",
     "ember-cli-test-loader": "ember-cli-test-loader#0.1.3",
+    "ember-data": "1.0.0-beta.16.1",
     "ember-load-initializers": "ember-cli/ember-load-initializers#0.1.4",
     "ember-qunit": "0.3.1",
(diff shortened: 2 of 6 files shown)


razor-x:
Additional note about `ember-cli-sas`: seems like it only works at `4.0.0-beta.5` as when I tried going to `4.0.0` it broke again. Not sure if this will affect a normal ember project or only the addon dummy site.

miguelcobain:
I started from your PR and made the other changes.

Regarding `ember-cli-sass`, I remind you that our use case is a bit different from the usual.
We want users to be able to **import** sass styles. The instructions at `ember-cli-sass` repo are for using sass in addons, but exposing css to the consuming app.

We need users to be able to `@import "ember-paper"` in their scss files. That way they can customize various ember-paper variables and behaviors.

Also, it looks like the workaround of installing autoprefixer is not needed anymore.

razor-x:
Thanks, that makes sense. I'm just starting with Ember and the upcoming v2.0 is bringing a lot of changes, so I figured I might as well help where I can.

miguelcobain:
Thanks a lot and I hope you enjoy Ember. I'm sure you will. :)

PR #101, opened 2015-06-21
Bind-attr and hasBlock deprecation fixes

brianmbutler:


miguelcobain:
Two problems:
1. Like Travis shows us, the tests fail in ember-beta
2. There other places we use `{{#if template}}`

brianmbutler:
Yeah I need to rebase and squab and look at the fall back to pre-glimmer.   In 1.13+ I think you can just yield

miguelcobain:
I see you've made changes like from this
```hbs
<label {{bind-attr for=inputElementId}}>{{label}}</label>
```
to
```hbs
<label>{{label}}</label>
```

This will make attribute bindings go way. I think we should use the new syntax:
```hbs
<label for="{{inputElementId}}">{{label}}</label>
```

brianmbutler:
I can fix that up today.  I've seen other components who declare the bound attrs in the component.js but then just {{yield}} but I suppose then you have to declare it in your template like  {{#foo inputElementId=inputElementId}}.

Right?

miguelcobain:
That's right.

miguelcobain:
Fixed deprecations on master. Thanks.

PR #103, opened 2015-06-22
Ember Textsupport for paper-text

benedikt-d:


benedikt-d:
I guess it failed because glimmer is not working :-(

miguelcobain:
This was more difficult than it appears to.
Bubbling arbitrary actions from a sub component isn't easy.

benedikt-d:
Yes just noticed i forgot half the code :-) I'm on it...

benedikt-d:
Wondering if we would need something like target in the paper button.

miguelcobain:
I found this: https://gist.github.com/fotinakis/dbb33b5418b9561e8d31

But we still need to handle `focus-in` and `focus-out` inside out paper-button to set `focus` property.
I was trying to avoid to write all action handlers and just bubble them manually.

benedikt-d:
Hm not quite working. When focusing:
Uncaught TypeError: Cannot read property 'get' of null
Uncaught TypeError: Cannot read property 'set' of null

benedikt-d:
I don't quite understand what base focusable does in this case? And also why `this.set('focus',true);`is called twice once in base focusable and once in paper-text. The problem seems to be that the values of pressed and focus are null when these methods are called directly. Don't know how exactly I can avoid that.

benedikt-d:
Put the sendAction into the base-focusable mixin as this can be useful for other components as well. I think it should be ready now.

benedikt-d:
If one of those new other methods are useful for other components they can be moved there too I guess.

benedikt-d:
This solution causes problems with other components which use base-focusable...
Cannot read property 'set' of null
Maybe someone has some good advice on how to fix it.

PR #104, opened 2015-06-23
Finish textsupport

benedikt-d:


benedikt-d:
I added the actions now. Probably can be implemented in a nicer way. Also I'm not sure how to include the target stuff. I'm still kind of new to ember so feel free to reject. I just said to myself better try and fail than wait for someone else to tackle the problem.

PR #106, opened 2015-06-23
Feature/clickable list items

peec:
Work in progress here (mostly working),

Implementing List Controls, as of https://material.angularjs.org/latest/#/demo/material.components.list.

Implementation ideas gotten from both:
- https://github.com/angular/material/blob/master/src/components/list/list.js
- https://github.com/mike1o1/ember-material-design


miguelcobain:
Great jobs, @peec!

Adding ripples in ember-paper is very simple, just include RippleMixin in any component.
By default, the ripples will be appended to the topmost component element, but you can change the ripple container to a child element by setting the `rippleContainerSelector ` property, like we do in [paper-checkbox.js](https://github.com/miguelcobain/ember-paper/blob/master/addon/components/paper-checkbox.js#L24).

Much more self contained, modular and cleaner than a ripple service, IMO.

A really nice use case for this would be the sidemenu list from the demo app.

peec:
Thanks! 

Using the mixin for ripple effect with rippleContainerSelector seem to work very well :).

miguelcobain:
Relationships between components should use Ember Components's `nearestOfType` and not jquery.
But I think I can refactor that once we finish this.

peec:
@miguelcobain I have rewritten to use ember  way instead of using jQuery, also added examples of all now working components. Added all of the examples that angular have. 

Documentation page added also,

peec:
@miguelcobain This PR is complete from my perspective, with much refactoring to make it more ember like. Take a look when you have time, cheers.

miguelcobain:
Impressive work here, @peec.
Looks amazing.

Shouldn't this be under "List" route?
Angular Material has this stuff under the "List" route, separated by "Basic usage" and "List controls".

I can do that, no problem. Just asking for your opinion.

miguelcobain:
`paper-item` ended up being really simpler. Ember ftw. :dancer:

https://github.com/miguelcobain/ember-paper/blob/master/addon%2Fcomponents%2Fpaper-item.js

PR #109, opened 2015-06-24
Added bounds and type validation to keyboard input handler

### LINKED ISSUE #108: Discrete sliders go to infinity on right key press
via @mansona:
> I don't know if it's appropriate to mention a bug here (please let me know) but when I select one of the Rating: /5 - demo of theming classes sliders and hit right on the keyboard the slider decides to make it's way to Infinity.

/cc @mike1o1

elliot:
Fixes #108


miguelcobain:
Thanks a lot!

PR #112, opened 2015-06-25
General accessibility enhancements

peec:
This PR adds general accessibility for checkbox and switch components.

- add correct `role="checkbox"` to checkbox and switches. 
- add `aria-label` to checkbox and switch when label attribute is present. 
- Tabbing between checkboxes and switches gives a nice visual effect like angular material does.
- **switch:** Allow to use space to turn switch on/off. 
- **list controls**: Hover effect and focus effects on corresponding buttons. ( This gives the side menu a nice touch).

Source of implementation:
https://material.angularjs.org/latest

### DIFF
--- a/addon/components/paper-button.js
+++ b/addon/components/paper-button.js
@@ -8,9 +8,10 @@ export default BaseFocusable.extend(RippleMixin, ProxiableMixin, ColorMixin, {
   attributeBindings: ['target', 'action'],
   tagName: 'button',
   classNames: ['md-button','md-default-theme'],
-  classNameBindings: ['raised:md-raised', 'icon-button:md-icon-button'],
+  classNameBindings: ['raised:md-raised', 'icon-button:md-icon-button', 'focus:md-focused'],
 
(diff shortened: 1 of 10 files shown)


miguelcobain:
Two things:
- What's the problem with tabbing through the menu items? I believe that is expected. The flaw here is that there is no visual feedback. That may be fixed by #106 
- Can we use the constant for `KEY_CODE_SPACE` in utils?

peec:
@miguelcobain 

> What's the problem with tabbing through the menu items? I believe that is expected. The flaw here is that there is no visual feedback. That may be fixed by #106

You are correct, I initially thought it was like that but after checking angular material I see that they can tab through menu items.

> Can we use the constant for KEY_CODE_SPACE in utils?

Added constants in checkbox and switch.

miguelcobain:
I'm currently a bit busy, but I promise to review all these PR very very soon.
Many thanks!

peec:
This is too outdated for now, I'll create a new PR for this soon. Closing this for now.

Issue #86, opened 2015-05-09
Glimmer support / hasBlock support


SirZach:
After upgrading to ember/canary to get the glimmer engine, paper-buttons no longer show up in the app. I'm not entirely sure why this is breaking.

SirZach:
If you could point me in the right direction I could take a crack at getting a PR going

miguelcobain:
@SirZach, ember-try tests against ember canary. Looks like the tests are passing in ember canary.

What is exactly the problem? What don't the buttons show up?
Please try with the latest `0.1.0` version.

SirZach:
I'll try with 0.1.0 tonight. The text in the buttons were not showing up.

miguelcobain:
How did it go?

SirZach:
I'm seeing that if I use the buttons as block components then the text on the button doesn't show up. If I switch it over to a non-block component and set the text to the label attribute, it'll show up.

miguelcobain:
Nice.

A failing test would be awesome.

miguelcobain:
Looks like the way to go now is using `hasBlock`: http://emberjs.com/api/classes/Ember.Component.html#property_hasBlock

SirZach:
I'll take a crack at spinning up a PR although not sure when you can merge it since I'm not sure when you want to start supporting glimmer...

miguelcobain:
Ideally we should support the same versions that ember-try tests, and that includes Glimmer.
There are things we can do to support both.

SirZach:
So I've been playing around with some tests and whatnot and have found that if I use ember-1.13 or later then using <code>hasBlock</code> is the way to go, otherwise, we need to use <code>template</code>. I can write a computed property in the base component to check for <code>hasBlock</code> first and then fall back to <code>template</code> if that makes sense to you.

SirZach:
currently dealing with a bug where the actual implementation for <code>hasBlock</code> has disappeared. tracking progress at https://github.com/emberjs/ember.js/issues/11301. until that's fixed my PR will have to be on hold

SirZach:
When https://github.com/emberjs/ember.js/pull/11313 is merged in, we'll be able to continue to check for <code>{{#if template}}</code> until you don't want to support pre-Glimmer and then you can make the change to <code>hasBlock</code>.

miguelcobain:
Bingo. :+1:

benedikt-d:
https://github.com/miguelcobain/ember-paper/pull/118

[...]

PR #113, opened 2015-06-25
Upgrade/emberjs 1.13.2

peec:
This PR will upgrade the ember version to 1.13.2 and also include changes to make paper compatible with 1.13.2.

Specifically this is done:

- Change to 1.13.2 of emberjs.
- ~~The use of nearestOfType (`paper-sidenav-toggle` component) does not seem to work in 1.13.2. Instead a global event bus is used here. The event bus can come handy for other components as well. The event bus principle is explained quite nicely [here](http://www.thesoftwaresimpleton.com/blog/2015/04/27/event-bus/). As of such one can now trigger from any component:~~

    ```javascript
        paperEventBus: Ember.inject.service('paper-eventbus'),
        someAction: function () {
            this.get("paperEventBus").publish('paper:toggle-sidenav');
            // or 
            this.get("paperEventBus").publish('paper:expand-sidenav'); 
        }
    ```
    ~~Obviously this event-bus is global, so it should only be used for global components. One could also make these events non-global by adding attribute to the event name. i.e. `.publish('paper:toggle-sidenav-' + this.get("menuhandle")')`. Though I do not think this is necessary for the sidenav component.~~
    **EDIT**: using parentView now.

- Updated `{{#if template}}` to respective `{{#if hasBlock}}`. ( With respective mixin to add support for pre 1.13.X releases).
- Updated `ember-prism` to 0.0.5.


miguelcobain:
@peec Please read https://github.com/emberjs/ember.js/issues/11170

tl;dr
- Right now, we should use `parentView`
- In the future we may be doing stuff like this:
```hbs
{{#paper-nav-container as |c|}}
  {{#c.paper-sidenav}} {{!-- sidenav here --}} {{/c.paper-sidenav}}
  {{#c.paper-content}} {{!-- content here --}} {{/c.paper-content}}
{{/paper-nav-container}}
```

Can you please refactor this PR to use `parentView` instead?

miguelcobain:
Also, does using `hasBlock` break when using Ember 1.12?
Check #86.

peec:
@miguelcobain 

I have now done the following:
- **hasBlock:** Created backwards compatible mixin so it doesn't break pre - 1.13.X releases.
- Removed service principle from commits and added support for **parentView**

miguelcobain:
@peec I think `nearestOfType` works fine now.

peec:
@miguelcobain 

Good idea, I tested with nearestOfType and it works fine now. 

Added:
- nearestOfType support again
- Check if hasBlock exist, if it does use it otherwise use this.get("template")

miguelcobain:
Please remove `app/services/paper-eventbus.js`.

miguelcobain:
Thanks @peec!

I opted to keep ember 1.12 for now. That should affect people's projects, because ember-paper is now compatible with 1.13 as well.

When we move to ember-cli 1.13.1 we use its ember version.

PR #111, opened 2015-06-24
implementing paper-dialog

mansona:
This is quite a rough draft with a lot of copy-paste from Angular-Paper but it works :+1: let me know if you have any questions

PR #116, opened 2015-06-26
implementing paper-dialog

mansona:
This is quite a rough draft with a lot of copy-paste from Angular-Paper but it works :+1: let me know if you have any questions

Just re-submitting https://github.com/miguelcobain/ember-paper/pull/111


mansona:
I've updated this in line with master and also fixed an issue I created with backdrop pointer events

miguelcobain:
@mansona I think we currently have two problems with this PR:
- No animations? `transition-in` class is there, but I don't see animation.
- Dialogs have the title hardcoded in a `paper-toolbar` component. That may not be always the case. Check alert and confirm [angular-material demos](https://material.angularjs.org/latest/#/demo/material.components.dialog). 

My suggestion: If there is a `title` property, print it in an `<h2 class="md-title">` (just like the markup in alert and confirm demos). Otherwise don't print any title and delegate all the content to the user through `{{yield}}`. In Angular material, if you want a toolbar in a dialog, you have to do it yourself, and I think that's fine, i.e semantically speaking I feel right wring a `paper-toolbar` inside a `paper-dialog`.

peec:
Since we both have #169 and #172 that implements dialog with animation, we should use one of those. So closing this for now

Issue #100, opened 2015-06-17
Versioning of ember-paper


benedikt-d:
Ember Data and Ember Liquid Fire are going to switch to a new versioning system. Should this library follow?

http://emberjs.com/blog/2015/06/16/ember-project-at-2-0.html

SirZach:
:+1:

miguelcobain:
Doesn't seem like a common practice with 3rd party libraries, and ember-paper isn't **that much** tied to each ember version.

Closing for now.

PR #118, opened 2015-06-27
Removed template and replaced with hasBlock.

benedikt-d:
Because template is deprecated in Ember 1.13


miguelcobain:
1. There are other places where we use `template`, not just on paper-button. When we change it, we must change it everywhere.
2. This would fail for people using Ember 1.12, right?
3. I think a good approach would be to have a computed property called `hasBlock` on a separate mixin that would check for both `hasBlock` (super) and `template`. When Ember 1.13 hits release channel, we can get rid of that mixin.

benedikt-d:
1. Maybe have not seen them. Gotta check.
2. Yes I think it would fail.
3. Hasn't 1.13 hit the release channel? You are right though backwards compatibility is always good!

benedikt-d:
Another possibility is to make different forks for different versions of Ember. The downside is that you have to support different versions and that older versions of ember probably won't benefit from all the new stuff that is coming unless there is pull requests for those... https://github.com/miguelcobain/ember-paper/issues/100

miguelcobain:
@benedikt-d 

Briefing:
- `hasBlock` is now supported **with** backwards compatibility via #113 
- paper-text was rewritten from scratch

So, can you merge these changes with latest master?

PR #134, opened 2015-07-15
Feature: Progress Linear/Circular Component

peec:
This PR gives 2 new components:

- Progress Circular
- Progress Linear

These works just like ember-material-design.

Converted from http://mike1234.com/ember-material-design/ ( Thanks @mike1o1  ).

Notes:
- Added [Sniffer service](https://github.com/mike1o1/ember-material-design/blob/master/app/services/sniffer.js). Also added a modified service of "constants", named "browser-compatibility" which is used by the linear progress component. A service fits very well for this purpose because the detection only needs to be done once and not per component basis. So I also added these services which the progress-linear component needs to create dynamic `transforms` attribute on the DOM.

### DIFF
--- a/addon/components/paper-progress-circular.js
+++ b/addon/components/paper-progress-circular.js
@@ -1,0 +1,30 @@
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+  browserCompatibility: Ember.inject.service('browser-compatibility'),
+
+  classNames: ['md-default-theme'],
+
+  tagName: 'md-progress-circular',
+
+  attributeBindings: ['value', 'md-mode'],
+
+  mdDiameter: 48,
+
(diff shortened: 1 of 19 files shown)


peec:
@miguelcobain This PR ( progress-linear ) is now the only thing left before autocomplete #132 is done. This PR implemented as ember-material-design. Could you take a look and possibly merge this when you have time?

peec:
@miguelcobain I see that progress-linear is now merged, thanks. I did some changes today, in this PR to also include paper-progress-circular as well. Should i create a new PR for progress circular now that linear is in master?

miguelcobain:
@peec A new PR would be awesome!

peec:
Closed, ref #136

PR #135, opened 2015-07-19
Fix deprecation: paper-progress-linear

peec:
This PR fixes deprecation notice for progress-linear, tested with (1.13.3):

```
DEPRECATION: A property of <dummy@view:-outlet::ember279> was modified inside the didInsertElement hook. You should never change properties on components, services or models during didInsertElement because it causes significant performance degradation.
```

### DIFF
--- a/addon/components/paper-progress-linear.js
+++ b/addon/components/paper-progress-linear.js
@@ -23,7 +23,10 @@ export default Ember.Component.extend(ColorMixin, {
   didInsertElement() {
     this._super(...arguments);
 
-    this.set('ready', true);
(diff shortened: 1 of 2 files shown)


miguelcobain:
I don't really understand the point of this `md-ready` class.

We're adding it on the `didInsertElement` hook. Isn't that the same as including that class directly in the template?

peec:
@miguelcobain Yeah, I just tried without the whole didInsertElement and just use `md-ready` directly in the template. Updated the PR with working changes. I think it's only needed because how angular renders the template.

PR #136, opened 2015-07-19
Feature: paper-progress-circular component

peec:
- Converted from ember-material-design
- Added tests
- Uses the new color mixin just as paper-progress-linear.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Ember Paper Changelog
 
-### 0.2.5 (Jul 18, 2015)
+### 0.2.5 (Jul 20, 2015)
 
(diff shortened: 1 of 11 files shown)


miguelcobain:
Perfect!
Just two things:
 1. I'm trying to avoid any `md-` in the public API, just like I did with paper-linear-progress. Can we do the same here?
 2. Can you update the `CHANGELOG.md`

peec:
@miguelcobain Ah, didn't notice the md attributes, updated PR with removed md- syntax and changelog.

miguelcobain:
Tests are failing. #137 

I merged the changes anyway, because I believe nothing was changed on ember-paper to cause those fails.

PR #138, opened 2015-07-23
Sidenav + layout fixes to ember-paper

peec:
Demo of this PR: http://jensb.home.pkj.no/ember-layout/index.html

This PR is about:

- using latest angular-material scss sources for the sidenav and backdrop. Backdrop is also used in other components (menu) so we need to make it modular and removable from DOM.
- paper-menu-toggle is now flexible and accepts any component (changed all to: paper-button (which gives e.g.  ripple effect ).
- paper-sidenav supports argument `locked-open` which supports all given constants (in constants service like in agular-material): `sm, gt-sm, md, gt-md, lg, gt-lg` etc. This means you can have many components with flexible lock-open on the menu.
    - Polyfill for `window.matchMedia` is included as a bower dependency. ( Which is needed to check for locked-open attributes ). 
- Cleaned up, moved demo related styles to "demo.scss". Styles that are needed is in app.scss still such as 100% width's / height's .
- paper-backdrop: Now reusable ( many components use md-backdrop with a custom action ).
- Now uses https://github.com/peec/ember-css-transitions so we can use angular's stylesheets for all animations.

Some of this are from ember-material-design, such as making menu animating ( see `app/styles/paper-sidenav.scss` )

**Edit: Updated to latest 0.2.7**


peec:
Animations should now work ( and are generic implementation against angular's structure of defining animations in the css (that is compatible with `$animate` angular service.

addon/components/paper-sidenav.js
```diff
@@ -1,7 +1,78 @@
...
+
+  'locked-open': 'gt-sm',
+  closed: true,
+
+  navContainer: Ember.computed(function () {
+    return this.nearestOfType(PaperNavContainer);
+  }),
+
+  lockedOpenDefinition: Ember.computed.alias('locked-oppen'),
+
```
peec: locked-oppen , should be open.

peec:
@miguelcobain Refactored some code style to correct and polished this one, it should be good and contains latest scss for backdorp and also animation now works. This contains the important animation-mixin that  is used to animate components and conforms to the ng-enter ng-enter-active style approach to animations (so we can use the CSS directly from angular ) Could you take a look at this one when you have fime ?

peec:
@miguelcobain This PR now uses ember-css-transitions. Demo updated also to see that backdrop works:  http://jensb.home.pkj.no/ember-layout/index.html

miguelcobain:
@peec great work, as usual! I believe the introduction of ember-css-transitions is a step forward.

We can focus on possible animation improvements separately on that project.

PR #139, opened 2015-07-24
Fix focusables, allow focus effect when being focused.

peec:
Right now, tabbing through items that are focusable does not trigger md-focused class. This fixes it for radio, checkbox and button.


miguelcobain:
@peec Wouldn't this be an ideal candidate for `BaseFocusable`?

Affected components would be `paper-button`, `paper-checkbox`, `paper-input`, `paper-radio`, `paper-slider` and `paper-switch`.

PR #146, opened 2015-08-10
Upgrade to ember 1.13.7

peec:
Upgrade to ember 1.13.7 including cli:
- *classNames* to **class** in component invocation (Which is indeed best practice.).  See https://github.com/emberjs/ember.js/issues/12008
- `@each` to `[]` in CP deps.
- Upgrade deps according to official notes: https://github.com/kellyselden/ember-addon-output/commit/f6f61d55c31d631203bc5491432b435e2cc807c2


miguelcobain:
Awesome! :shipit: 

Can we get an entry on Changelog in version `0.2.7`?

peec:
@miguelcobain Done :+1:

PR #148, opened 2015-08-13
Upgrade to ember 2.0.0.

peec:
- Upgrades to ember 2.0.0
- Autocomplete: Fix deprecation, Ember.String.fmt is deprecated in 2.0.0, and no alternative for that exist (only compile time ES6 (does not work with the use-case we have)). So, only used `.replace` for the dynamic part `$searchText`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Ember Paper Changelog
 
+### 0.2.8 (Aug 13, 2015)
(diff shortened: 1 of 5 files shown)


peec:
This is too early.

balupton:
Why is this too early?

peec:
@balupton I think we should [wait](https://github.com/ember-cli/ember-cli/pull/4677) for ember-cli to align with 2.0.0 - just for simplicity, ember-paper will still work with 2.0 on our own projects.

PR #151, opened 2015-08-17
Only show validation errors if user has interacted with the element. …

### LINKED ISSUE #150: Only show validation errors after user has interacted with element
Currently, `{{paper-input}}` is in violation of [Material Design specification](http://www.google.com/design/spec/patterns/errors.html) which states that an application should 'show error text only after user interaction with a field'. Right now, the paper input component shows error messages before the user has interacted with it. This can be fixed by adding an `isTouched` property which gets set on focusOut, and then only displaying error messages if `isTouched` is true.

mike1o1:
…Fixes #150

### DIFF
--- a/addon/components/paper-input.js
+++ b/addon/components/paper-input.js
@@ -14,7 +14,7 @@ export default BaseFocusable.extend(ColorMixin, FlexMixin, {
   inputElementId: Ember.computed('elementId', function() {
     return 'input-' + this.get('elementId');
   }),
-  isInvalid: Ember.computed('value', function() {
+  isInvalid: Ember.computed('isInvalid', 'value', function() {
     return this.validate();
   }),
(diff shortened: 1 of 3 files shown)


addon/components/paper-input.js
```diff
@@ -14,7 +14,7 @@ export default BaseFocusable.extend(ColorMixin, FlexMixin, {
   inputElementId: Ember.computed('elementId', function() {
     return 'input-' + this.get('elementId');
   }),
-  isInvalid: Ember.computed('value', function() {
+  isInvalid: Ember.computed('isInvalid', 'value', function() {
```
miguelcobain: Why does `isInvalid` depends on `isInvalid`?

peec:
Nice one! :+1:

PR #161, opened 2015-08-23
Allow paper-input to be a block component, so user can add helper text

mike1o1:
User can add anything else they want inside the `md-input-container` block.

Also add an example of how to use this new block format to create a helper text.

### DIFF
--- a/app/templates/components/paper-input.hbs
+++ b/app/templates/components/paper-input.hbs
@@ -21,3 +21,5 @@
 {{#if maxlength}}
     <div class="md-char-counter">{{renderCharCount}}</div>
 {{/if}}
+
+{{yield this}}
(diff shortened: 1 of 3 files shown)


mike1o1:
Any thoughts or feedback on this, or anything preventing this from being merged in?

peec:
@mike1o1 This should be merged it means more flexibility. :+1: 

@miguelcobain  @mike1o1  As of things being merged, I think we could need some more core devs on ember-paper, since the activity is growing on this project. There is only so much one person can do on such a large project as its growing in the code-base. Just a thought.

DanChadwick:
Old issue, but it seems like a good idea. Can this be done with just angular material 1.0.6 styles?  If se, a rebate on wip/v1.0.0 is needed.

jamesdixon:
Just ran into a situation where this would be great to have. Specifically, I'm trying to add an icon that's not a material icon and alignment is a pain. Would be much easier if I could place it within the block.

PR #162, opened 2015-08-26
Add matchMedia to bower.json

mnutt:
Fixes a missing dependency added by https://github.com/miguelcobain/ember-paper/pull/138.


miguelcobain:
Good catch. Thank you.

PR #163, opened 2015-08-28
Enable passing additionalParams to paper-item proxy action

mhretab:
this PR just adds additional params to paper-item action proxy. This is so that more than one parameters are possible to pass to paper-item. One use-case is to use it with transitionToRoute('route', id).


miguelcobain:
The correct way to handle this now is to use closure actions.
Thanks anyway for the PR!

PR #165, opened 2015-09-01
Feature: fab-buttons & fab-speed-dial / tooltips

chrisitananese:
Demo: http://chrisitananese.github.io/ember-paper/#/fab-speed-dial

* Full features work
* Angular scss
* Animations work

```handlebars
{{#paper-fab-speed-dial direction="down" animation="scale" warn=true}}

      {{#paper-button fab=true}}{{paper-icon icon="android"}}{{/paper-button}}
      {{#paper-button fab=true}}{{paper-icon icon="person"}}{{/paper-button}}
      {{#paper-button fab=true}}{{paper-icon icon="message"}}{{/paper-button}}

{{/paper-fab-speed-dial}}
```

PR #169, opened 2015-09-04
Feature/paper dialog

chrisitananese:
Demo: http://chrisitananese.github.io/ember-paper/#/dialog

* Added alert, confirm and custom dialogs.
* Animations don't work. Same problem as in https://github.com/miguelcobain/ember-paper/pull/165 .
The transissions uses different classes than 'enter' and 'leave'.

Had an issue with chrome with custom dialogs, they should work properly now.


miguelcobain:
Same here, @chrisitananese.

DanChadwick:
Dialogs are now implemented in the `wip/v1.0.0` branch. Since all work is focused on finishing that branch and releasing 1.0, let's close this since it duplicates work that has already been merged. Thanks!

PR #172, opened 2015-09-10
Dialogs

peec:
Supports:
- Angular SCSS
- Animations - translate3d mixin was needed as it's not the same concepts as regular css animations and rather computed styles with JS.
- All dialog specific feature works.

Can be enhanced with: keyevents + some button focusing on e.g. confirm dialog, but I currently have no time to do that.


jamesdixon:
any plans on merging this one in anytime in the near future? thanks!

miguelcobain:
This PR needs a rebase. :cry:

afomi:
bump (and nice feature!)

miguelcobain:
@peec I suspect you have less time now, but I wonder if you could work to make this on par with #249. It will be the big thing in ember-paper.

peec:
@miguelcobain Yup you are correct, awesome work you did there, I have had no time for this lately. I will see if I can find time for it - I guess I soon will have,

DanChadwick:
Thanks for the effort here. I would love to see this merged.

PR #188, opened 2015-10-04
Allow 'type' attribute on {{paper-button}} component

mike1o1:
Allow `type` attribute to be passed to the `{{paper-button}}` component


miguelcobain:
We should probably add this to changelog as well.

mike1o1:
No problem. I was a bit confused on the current changelog, as it references a version 0.2.9 which doesn't seem to be released yet.

miguelcobain:
That's right. We're "stacking" changes there. :)

PR #197, opened 2015-10-22
check if grid list is destroyed before invalidating it fixes #196

### LINKED ISSUE #196: "Assertion Failed: calling set on destroyed object" when leaving route containing paper-grid-list
`Assertion Failed: calling set on destroyed object` is thrown when calling `this.setProperties` in the `finally` block of paper-grid-list `layout` function.

It looks like the root of this is invalidating the layout in the `invalidateLayout` function when component is being destroyed.

EricSchank:
fixes #196 

Fix assertion thrown when invalidating destroyed `paper-grid-list` component.

### DIFF
--- a/addon/components/paper-grid-list.js
+++ b/addon/components/paper-grid-list.js
@@ -326,7 +326,7 @@ export default Ember.Component.extend({
     },
 
     invalidateLayout() {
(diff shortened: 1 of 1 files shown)


mike1o1:
Great catch and thanks for the fix! Looks good to me.

PR #198, opened 2015-10-22
Add install and build details

qaid:
This would admittedly only serve people relatively unfamiliar with running an Ember app.

If someone finds this project without much Ember experience, I think it would be useful to be able to spell out all the steps to get them up and running.

What do you think?


miguelcobain:
Not needed, IMO. The docs already point at ember-cli and ember. Since this is an ember-cli addon, one must know ember-cli first.

PR #199, opened 2015-10-24
Autocomplete/feature Search for a parts of string

canufeel:
Take a set of words like full address which can consist from `City` `Street` `Country`. If your backend implements spliting the search string on to separate parts and making search across multiple tables the search result returned to autocomplete might not comply `search.indexOf(searchText) === 0;` The other case when it might not comply if you are doing letter-wise search on the backend. etc.

Added feature for paper-autocomplete to support providing promises to model property.

Added feature to set `flex` property on paper-autocomplete. The provided `flex` should be numeric value  `1-100` like `flex="60"`. Setting flex would scale the element to the provided percentage of parent container

### DIFF
--- a/addon/components/paper-autocomplete.js
+++ b/addon/components/paper-autocomplete.js
@@ -26,7 +26,7 @@ export default Ember.Component.extend(HasBlockMixin, {
 
   tagName: 'md-autocomplete',
   classNameBindings: ['notFloating:md-default-theme'],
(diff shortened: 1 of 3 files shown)


DanChadwick:
@canufeel -- Shouldn't 951b4e3 be in #237 since it's unrelated to the autocomplete functionality?

canufeel:
@DanChadwick You are actually right. I think i would split it and add another PR

PR #209, opened 2015-11-10
Fix issue with `border-bottom-color` in paper-select not correctly ge…

mike1o1:
Paper-select is not using the $primary color directly. The stylesheet was referencing the $primary variable directly, instead of accessing it through the `color` function.


xomaczar:
@miguelcobain can we publish a new version to npm with this fix, currently with ember-cli@1.13.12/13 building fails due to ^

miguelcobain:
@xomaczar It should be published

xomaczar:
@miguelcobain thx

PR #219, opened 2015-11-13
Inject wormhole outlet through addon hook

### LINKED ISSUE #214: Add note about Wormhole
Using Ember Paper for the first time in a while, it seems like it's improved a ton! One thing I noticed though; you need to add the Wormhole outlet to your `application.hbs` file, as such:

```hbs
<div id="paper-wormhole"></div>
```

However, this doesn't seem to be mentioned anywhere.  If it's not, could this be added to the Readme?

alexlafroscia:
Instead of needing to include the Wormhole outlet in your application manually, the addon will now add the `div` automatically.

Note: I'm not sure if this will be backward compatible with having it also defined in the template, for people that already are using it the previous way.

Closes #214

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Ember Paper Changelog
 
+### 0.2.10
+- [#219](https://github.com/miguelcobain/ember-paper/pull/219) Inject wormhole outlet through addon hook
+
 ### 0.2.9 (Sep 20, 2015)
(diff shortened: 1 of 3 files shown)


miguelcobain:
@alexlafroscia Looks great!

Could you please add a changelog entry for the next patch version?

alexlafroscia:
@miguelcobain done :+1:

PR #232, opened 2015-12-03
Enable paper-input to display server generated errors

### LINKED ISSUE #216: Ember Validations
I see inputs have error handling which is great, but is it possible to integrate with ember validations? https://github.com/dockyard/ember-validations
At the moment I don't quite see a way to integrate server side validation error messages into the paper inputs.

chbonser:
Server side or other client generated errors can be displayed via paper-input using the `errors` attribute. The `errors` attribute expects an array of JavaScript objects which each respond to a `message` attribute.  `errors` can be set to the DS.Errors object directly off of your model to easily display server side errors.

The `errors` attribute can be used in conjunction with paper-input native errors.  In the case where both types of errors occur simultaneously, only the `errors` attribute messages will be displayed.

This would resolve #216

### DIFF
--- a/addon/components/paper-input.js
+++ b/addon/components/paper-input.js
@@ -15,14 +15,25 @@ export default BaseFocusable.extend(ColorMixin, FlexMixin, {
   inputElementId: Ember.computed('elementId', function() {
     return 'input-' + this.get('elementId');
   }),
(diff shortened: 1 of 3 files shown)


DanChadwick:
Closing in favor of discussion in #216. Please reopen if the new paper-input doesn't help you.

PR #238, opened 2015-12-06
paper-select-core: apply the disabled attribute

### LINKED ISSUE #236: paper-select: not disabled properly
If a user uses a `{{paper-select disabled=true}}` component with a disabled property, the disabled property does not get applied correctly to the `paper-select-core` component. The component is unable to be activated, which is the correct behavior, but since the `md-select` element does not have the `[disabled]` attribute, the CSS styling is incorrect.

mike1o1:
Closes #236

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Ember Paper Changelog
 
+### 0.2.11 (TBD)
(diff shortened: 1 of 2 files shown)


miguelcobain:
This should work correctly in the new paper-select.

PR #242, opened 2015-12-11
paper-autocomplete: cancel debounce on component destroy

steveklewis:
If the delay is long enough, sometimes the component will be destroyed by ember before the setDebouncedSearchText function is actually called. This pull request just cancels the debounce initiated prior to willDestroy.


xomaczar:
This has been addressed in the latest release of autocomplete

PR #253, opened 2015-12-30
Add `closeOnClick` to paper-sidenav

### LINKED ISSUE #204: Sidenav immediately closes upon any click - user unable to interact
If a user opens the `{{paper-sidenav` component and performs any kind of click then it immediately gets closed, even if that click is on a non-link, such a toggle for a second level menu item. It appears that there is a `click` action handler that toggles the menu closed.

I don't think this is ideal, but I can see how it's needed to close the menu on route change. One option is to require the user to handle the `didTransition` event in their application route, to toggle the `sidebarVisible` property of the sidebar element to false.

This can be made a non-breaking, opt-in change by adding a new `closeOnClick` property with a default value of true. If this isn't set, then don't close on click. This way users can opt-in by explicitly setting this property to false.

knownasilya:
In response to #204 (fixes #204)

* `closeOnClick` defaults to `true`, making this option backwards compatible.
* Updated changelog
* Updated sidenav docs with attribute and example code fix (formatting was horrible)
* Fixed doc site not scrolling

screen shot 2015-12-31 at 12 53 09 pm

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Ember Paper Changelog
 
+### 0.2.11
+- [#253](https://github.com/miguelcobain/ember-paper/pull/253) Add `closeOnClick` to paper-sidenav
+
 ### 0.2.10 (Nov 23, 2015)
 - [#178](https://github.com/miguelcobain/ember-paper/pull/178) Listen for model changes
 - [#219](https://github.com/miguelcobain/ember-paper/pull/219) Inject wormhole outlet through addon hook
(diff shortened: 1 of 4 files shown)


mike1o1:
Thanks for taking the time to implement! Would you mind adding a changelog entry?

knownasilya:
All set @mike1o1

mike1o1:
Docs are in the dummy app. Thanks!

knownasilya:
@mike1o1 all fixed, see the screenshot above.

miguelcobain:
@knownasilya Perfect PR. Thanks.

knownasilya:
No, thank you! :smile:

PR #260, opened 2016-01-10
jquery version fixed

canufeel:


mike1o1:
What's the reason for this change?

miguelcobain:
I heard this broke tests?

canufeel:
@mike101 @miguelcobain yes indeed it does. The test wouldn't work with `"jquery": "^1.11.3"` as it now resolves to `1.12.0`

PR #264, opened 2016-01-18
Update README.md

SirZach:
Adding a note explaining why someone might get a compile error around app.scss not existing if they were upgrading from previous versions of ember-paper.


miguelcobain:
Glad it worked. :)

PR #275, opened 2016-02-03
ripples and icons

peec:
Fixes (1.0 wip):

- CSP for fonts include.
- Ripples: The ripple mixin now works like intended.
- paper-icon now works. Fix icon-fonts and upgrade fonts to v11.


miguelcobain:
@peec I appreciate the effort.

Isn't a ripple service an overkill? A service should be used whenever there is a shared state. What shared state is there to hold in ripple effects?

peec:
@miguelcobain True, there is no shared state. I figured that as a service the ripple effect can be used on DOM elements separately without any need of a component and multiple times. E.g:

```javascript
didInsertElement() {
    this.get('ripple').attach(this.$('.some-item-one'));
    this.get('ripple').attach(this.$('.some-item-two'));
}
```
So basically it's completely separated from the component, rather bound to a DOM node. 

But then again, I understand that it might be overkill.

miguelcobain:
@peec 

I don't see why that wouldn't be possible without a service.
Example: 
```js
didInsertElement() {
    this.attachRipple(this.$('.some-item-one'));
    this.attachRipple(this.$('.some-item-two'));
}
```

`attachRipple` would be a method provided by including `RippleMixin`.

miguelcobain:
I still like the `rippleContainerSelector` approach for the simple scenarios:

1. This would atttach ripple to `this.$()` (or `this.element`)
```js
export default Ember.Component.extend(RippleMixin, {
});
```
2. This would atttach ripple to `this.$('.inner-selector')`
```js
export default Ember.Component.extend(RippleMixin, {
  rippleContainerSelector: '.inner-selector'
});
```
3. This would atttach ripple to `this.$('.inner-selector')` and `this.$('.another-selector')`
```js
export default Ember.Component.extend(RippleMixin, {
  rippleContainerSelector: ['.inner-selector', '.another-selector']
});
```
4. This would not attach any ripple, basically turning it into the "manual" mode. This means the component including the mixin will have to manually call `attachRipple` with some selector/element at some point.
```js
export default Ember.Component.extend(RippleMixin, {
  rippleContainerSelector: null
});
```

I think these guidelines would make it a really flexible mixin!

peec:
@miguelcobain Good points. Moved from service to ripple-mixin.

addon/mixins/ripple-mixin.js
```diff
@@ -1,287 +1,285 @@
...
-  getRippleCss(size, left, top) {
-    var rect,
-    css = {
-      backgroundColor: rgbaToRGB(this.color),
-      borderColor: rgbaToRGB(this.color),
-      width: size + 'px',
-      height: size + 'px'
-    };
+  createContainer() {
+    var container = Ember.$('<div class="md-ripple-container"></div>');
```
miguelcobain: We're now using `let` for variables and `const` for constants.

addon/mixins/ripple-mixin.js
```diff
@@ -1,287 +1,285 @@
...
-/* global propagating */
+/* global window */
+
+const DURATION = 400;
 
 export default Ember.Mixin.create({
-  mousedown: true,
-  hover: true,
-  focus: true,
+  util: Ember.inject.service(),
```
miguelcobain: We're now using `const { inject } = Ember;` and then `util: inject.service()`.

addon/mixins/button-ripple-mixin.js
```diff
@@ -0,0 +1,9 @@
+import Ember from 'ember';
+import Ripple from './ripple-mixin';
+
+export default Ember.Mixin.create(Ripple, {
+    rippleContainerSelector: null,
+    fitRipple: Ember.computed.alias('isIconButton'),
```
miguelcobain: Same here `const { computed } = Ember;`.

miguelcobain: It is probably better to use `computed.readOnly` instead of `alias`.

miguelcobain:
@peec I'd prefer to eliminate the intermediate RippleMixins you created and have the properties on the component itself.

Otherwise we have things like importing `checkbox-ripple-mixin` on paper-switch.
I see it as more explicit.

You have any objections?

peec:
@miguelcobain Agreed. I updated and now it's just ripple-mixin along with optimizations for code structure.

addon/mixins/ripple-mixin.js
```diff
@@ -1,287 +1,286 @@
...
+          return false;
+        }
+      }
+
+    } while (element = element.parentNode);
+
+
+    return true;
+  },
+  inkRipple() {
```
miguelcobain: Since `inkRipple()` is basically a wrapper around `rippleInk` computed, we can replace all `inkRipple()` entries with `this.get('rippleInk')`.

addon/mixins/ripple-mixin.js
```diff
@@ -1,287 +1,286 @@
...
+    if (this.get('rippleInkColor')) {
+      return this.get('rippleInkColor');
+    }
+    return "";
+  }),
 
-  ripples: [],
-  rippleStates: [],
+  didInsertElement() {
+    this.rippleElement = this.$(this.get('rippleContainerSelector'));
```
miguelcobain: let's call `this._super(...arguments);` to make sure we don't break inheritance.

addon/mixins/ripple-mixin.js
```diff
@@ -1,287 +1,286 @@
 import Ember from 'ember';
-/* global Hammer */
-/* global propagating */
+const { inject, computed } = Ember;
+/* global window */
+
+const DURATION = 400;
 
 export default Ember.Mixin.create({
```
miguelcobain: `const { Mixin } = Ember`
`Mixin.create`

miguelcobain:
Why did we remove those tests in `paper-icon-test.js` and `paper-input-test.js`?

tests/dummy/app/templates/list-controls.hbs
```diff
@@ -28,7 +28,7 @@
         {{#paper-subheader class="md-no-sticky"}}Clickable Items with Secondary Controls{{/paper-subheader}}
 
         {{#paper-item action="transitionToWifiMenu"}}
-          {{paper-icon "network-wifi"}}
```
miguelcobain: Why did these happen?
Did the stylesheets change icon names?

app/templates/components/paper-icon.hbs
```diff
@@ -0,0 +1 @@
+{{iconClass}}
```
miguelcobain: Is this some kind of leftover? Not sure why we need to render the iconClass.

tests/dummy/app/templates/list-controls.hbs
peec: See https://github.com/google/material-design-icons/blob/master/iconfont/codepoints , this is valid names. I could not get it working with dashes.

app/templates/components/paper-icon.hbs
peec: See: http://google.github.io/material-design-icons/ . Usage: 
<i class="material-icons">face</i>

miguelcobain: ah, ligatures. Cool!

tests/dummy/app/templates/list-controls.hbs
miguelcobain: I see. The icon names have underscores.

I wonder why we had dashes before. Did material icons change?

peec: I think something changed but not sure where, you can also try here:
- http://jsbin.com/rifotulaja/edit?html,css,output
Change the underscores to dashes and it will break the icon,
ref. https://github.com/google/material-design-icons/issues/177

peec:
@miguelcobain tests now passes, i think its ready.

miguelcobain:
@peec only thing missing is to update the icon name list in the docs at `ember-paper/tests/dummy/app/controllers/icons.js`.

Alternatively we could just point users to the material icons page for that list. Thoughts?

peec:
@miguelcobain I actually updated that list, but its too large for github to see here. https://github.com/peec/ember-paper/blob/wip/v1.0.0/tests/dummy/app/controllers/icons.js

List is directly from https://github.com/google/material-design-icons/blob/master/iconfont/codepoints

Ill merge this then

miguelcobain:
Ah, sorry. I didn't see that.
Good job!

PR #277, opened 2016-02-04
1.0: paper-progress-linear

peec:
Update code-style, docs and unit -> integration test.


miguelcobain:
Could you please add more test cases like the ones here:
https://github.com/angular/material/blob/v1.0.4/src%2Fcomponents%2FprogressLinear%2Fprogress-linear.spec.js

Let's ignore accessibility for now. We'll deal with that in a separate issue.

addon/components/paper-progress-linear.js
```diff
@@ -1,5 +1,7 @@
 import Ember from 'ember';
 import ColorMixin from 'ember-paper/mixins/color-mixin';
+const {inject, computed, Component, isPresent} = Ember;
```
miguelcobain: to conform with other stlyes, please add spaces around declarations.

`const { inject, computed, Component, isPresent } = Ember;`

Sorry for this. We need to add decent jscs support in the project.

addon/components/paper-progress-linear.js
```diff
@@ -12,26 +14,25 @@ const MODE_DETERMINATE = "determinate",
   MODE_BUFFER = "buffer",
   MODE_QUERY = "query";
```
miguelcobain: Use single quotes in js files.

miguelcobain:
Also, we currently have a problem with component attributes. We have both `dash-attr` and `camelAttr` currently. That is inconsistent.

I'll discuss this with ember folks at slack.
I'm inclined to `camelAttr`. If that ends to become the decided format, we should change `buffer-value` to `bufferValue` and document this in the CHANGELOG.

Thank you so much for this, @peec.

addon/components/paper-progress-linear.js
```diff
@@ -12,26 +14,25 @@ const MODE_DETERMINATE = "determinate",
...
 
   init() {
     this._super(...arguments);
     this.setupTransforms();
   },
 
-  mode: Ember.computed('value', function() {
+  mode: computed('value', function() {
     var value = this.get('value');
     var bufferValue = this.get('buffer-value');
```
miguelcobain: We're now using `let` instead of `var`, expect for constants.

tests/integration/components/paper-progress-linear-test.js
```diff
@@ -0,0 +1,57 @@
...
+  this.render(hbs`{{paper-progress-linear value=12}}`);
+  assert.equal(this.$('md-progress-linear').attr('md-mode').trim(), 'determinate');
+});
+
+test('should auto-set the md-mode to "buffer" if a value and bufferValue is specified', function(assert) {
+  this.render(hbs`{{paper-progress-linear  value=50  buffer-value=100}}`);
+  assert.equal(this.$('md-progress-linear').attr('md-mode'), 'buffer');
+});
+
+test('should auto-set the md-mode to "buffer" if a value and bufferValue is specified', function(assert) {
```
miguelcobain: Duplicated test?

addon/components/paper-progress-linear.js
```diff
@@ -60,27 +61,26 @@ export default Ember.Component.extend(ColorMixin, {
...
-  bar1Style: Ember.computed('clampedBufferValue', function() {
-    return new Ember.Handlebars.SafeString(this.get('constants.CSS.TRANSFORM') + ': ' + this.transforms[this.get('clampedBufferValue')]);
+  bar1Style: computed('clampedBufferValue', function() {
+    return Ember.String.htmlSafe(this.get('constants.CSS.TRANSFORM') + ': ' + this.transforms[this.get('clampedBufferValue')]);
   }),
 
-  bar2Style: Ember.computed('clampedValue', function() {
+  bar2Style: computed('clampedValue', function() {
 
     if (this.get('mode') === 'query') {
```
miguelcobain: test against `MODE_QUERY` constant.

addon/components/paper-progress-linear.js
```diff
@@ -1,37 +1,38 @@
...
+    let value = this.get('value');
+    let bufferValue = this.get('buffer-value');
 
-    if (Ember.isPresent(value)) {
-      if (Ember.isPresent(bufferValue)) {
+    if (isPresent(value)) {
+      if (isPresent(bufferValue)) {
         return 'buffer';
       } else {
         return 'determinate';
```
miguelcobain: return query mode constants in this cp.

addon/components/paper-progress-linear.js
```diff
@@ -1,37 +1,38 @@
...
 
-const MODE_DETERMINATE = "determinate",
-  MODE_INDETERMINATE = "indeterminate",
-  MODE_BUFFER = "buffer",
-  MODE_QUERY = "query";
-
+const MODE_DETERMINATE = 'determinate',
+  MODE_INDETERMINATE = 'indeterminate',
+  MODE_BUFFER = 'buffer',
+  MODE_QUERY = 'query';
```
miguelcobain: make each const declaration separate. Similar to https://github.com/miguelcobain/ember-paper/blob/wip/v1.0.0/addon%2Fcomponents%2Fpaper-progress-circular.js#L8-L9

Again, sorry for this, but this will be in the jscs rules in a near future. So we're saving time upfront.

addon/components/paper-progress-linear.js
```diff
@@ -41,7 +42,7 @@ export default Ember.Component.extend(ColorMixin, {
     }
   }),
 
-  queryMode: Ember.computed('mode', function() {
+  queryMode: computed('mode', function() {
```
miguelcobain: Please change `queryMode` to `queryModeClass` as it makes this more evident that its purpose is to be used in the template.

addon/components/paper-progress-linear.js
```diff
@@ -60,27 +61,26 @@ export default Ember.Component.extend(ColorMixin, {
...
     }
 
-    return new Ember.Handlebars.SafeString(this.get('constants.CSS.TRANSFORM') + ': ' + this.transforms[this.get('clampedValue')]);
+    return Ember.String.htmlSafe(this.get('constants.CSS.TRANSFORM') + ': ' + this.transforms[this.get('clampedValue')]);
   }),
 
-  clampedValue: Ember.computed('value', function() {
-
-    var value = this.get('value');
+  clampedValue: computed('value', function() {
```
miguelcobain: Here is a much cooler way to clamp values: https://github.com/miguelcobain/ember-paper/blob/wip/v1.0.0/addon%2Fcomponents%2Fpaper-progress-circular.js#L29-L32

I'll make this a computed macro in the future.
Can you change both cps here to clamp using that method?

miguelcobain:
Perfect. Let's just wait for some input on `dash-attr` vs `camelAttr`.
What are your thoughts on that?

peec:
@miguelcobain Yes I agree, best to follow the ember-way on attribute naming, if they have suggestions

peec:
@miguelcobain I will merge this as it holds up my wip/v1.0.0 branch (should have branched it to wip/v1.0.0/paper-progress-linear instead). And we can discuss further in issue of wip1.0.0 i guess ?

PR #279, opened 2016-02-04
Upgrade to ember 2.3.0 and use "transition-to" helper

peec:
- Upgraded ember demo site to 2.3.0
- Use [transition-to](https://github.com/peec/ember-route-action) addon to route from components. This is not included as a dependency but rather dev-dependency.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,16 +1,16 @@
 {
   "name": "ember-paper",
   "dependencies": {
-    "ember": "1.13.11",
-    "ember-cli-shims": "0.0.6",
-    "ember-cli-test-loader": "0.2.1",
+    "ember": "2.3.0",
+    "ember-cli-shims": "0.1.0",
+    "ember-cli-test-loader": "0.2.2",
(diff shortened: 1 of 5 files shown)


miguelcobain:
@peec `transition-to` is available since which ember version?

peec:
@miguelcobain Sorry, I didn't explain - the `transition-to` helper is a [addon](https://github.com/peec/ember-route-action) in the dev-dependencies. I think this helper fits very well where we don't use link-to component

miguelcobain:
ah, got it.

But it would be useful in our project only, right?
We're not "re-exporting" that helper in any way.

I'm ok with that. Super useful helper.

peec:
@miguelcobain Yes you're right, I only included as a dev-dependency and used it in the demo pages. So It's not re-exported in any way, and should only be used on the demo dummy pages.

I think generally people would want the helper, but they can then include it themselves if they want the `transition-to` helper.

PR #281, opened 2016-02-05
V1.0.0/autocomplete

peec:
This is attempt to fix autocomplete.

Angular Material 1.0 introduced a new component (virtual-repeat) which is quite neat, but hard to implement. See more of virtual repeat here: https://material.angularjs.org/latest/demo/virtualRepeat 

So two new components is created: paper-virtual-repeat-container and paper-virtual-repeat.

This PR fixes the positioning of the autocomplete component which is basically broken in wip/v1.0.0. 

- [x] fix positioning of autocomplete by using virtual-repeat DOM structure so CSS styles get applied. Also fix JS for positioning the wrapper element.
- [ ] implement virtual-repeat ( this is hard. help is welcomed ) . 
- [ ] fix autocomplete overflow issue ( basically fixed when we have working virtual-repeat ).


miguelcobain:
For the virtual repeat we have good alternatives in ember. some and mirrors, probably.
@runspired, can you give an insight here?

runspired:
@miguelcobain I'd love to help out here.

DanChadwick:
I see both ember-smoke-and-mirrors and ember-collection. Any idea which would be better? There seem to be other options, too. I am also using ember-table, which plans to replace is homebrewed lazy rendering with (probably) ember-collection.

miguelcobain:
I believe smoke and mirrors is a bit more optimized.
@runspired, you have anything to say in your defense?

runspired:
There's a few major things coming in smoke-and-mirrors that will end any debates, and will land before ember-paper needs to make any decisions.   That, and ember-collection is always going to be a step or two behind the curve regardless of whether it fixes it's layout issues.

miguelcobain:
@runspired we don't get horizontal scrolling in neither of the libraries, right?

runspired:
The support is there under the hood in smoke-and-mirrors, I've avoided developing it because with the changes in https://github.com/runspired/smoke-and-mirrors/pull/98 enabling horizontal or vertical columns or grids will be trivial.

SaladFork:
Unfortunately GitHub doesn't seem to allow you to change the base branch you are PRing against and automatically closes any PRs when the base branch is deleted.

Moving forward [we are using the `master` branch for `1.0.0`](https://github.com/miguelcobain/ember-paper#version-roadmap) work so please feel free to open a new PR against `master` :)

PR #283, opened 2016-02-10
Autocomplete refactor

canufeel:
This is refactored version of #199 which adds:
 - `fullTextSearch` support for autocomplete
 - assigning of `Ember.ObjectProxy` objects to autocomplete model attribute on init or during usage.
 - Has dummy site docs updated to explain how these features work as well as some basic tests for `fullTextSearch`.
 - Fixes `Ember.String.fmt` deprecation by substituting it with `Ember.String.loc`
 - Adds its 5 cents to the autocomplete code readability.


miguelcobain:
The original autocomplete (from angular material) doesn't support full-text search?

brendanoh:
My thoughts:

@canufeel this PR looks good to me. @miguelcobain I am not sure I understand why this project is always limited by what ``` angular material ``` has or doesn't have. 

Personally, I have used Ember Paper twice on projects and both times have had to add in full text search to "fix" the autocomplete control. Fix in quotes cause it is not a bug but definitely feels missing to me.

I mean it boils down to these lines of text:

```
+      if (get(this,'fullTextSearch')){
 +        return search.indexOf(searchText) !== -1;  
 +      } else {
 +        return search.indexOf(searchText) === 0;
 +      }
```

I get trying to emulate the work done there to be consistent but this feels like a mistake on their part that we are replicating.

miguelcobain:
> I am not sure I understand why this project is always limited by what angular material has or doesn't have.

@brendanoh this is not true. The only thing we are somewhat limited is to angular material's styles. ember-paper can have the api it wants, and the functionality it wants, as long as it doesn't clash with angular-material's styles. This isn't certainly the case.

I merely asked because it felt impossible that angular material didn't have this. I wasn't suggesting *at all* that we shouldn't have this. On the contrary, I think this is something very fundamental.

Just trying to understand why we don't have this in the first place.

brendanoh:
Ahhhhhh @miguelcobain that makes so much more sense now...

In Angular Material you write your own comparator which would allow you do it any way you please.

CodePen of rewritten ```createFilterFor```: http://codepen.io/anon/pen/adgPgq

```
    function createFilterFor(query) {
      var lowercaseQuery = angular.lowercase(query);
      return function filterFn(state) {
        //return (state.value.indexOf(lowercaseQuery) === 0);
        return (state.value.indexOf(lowercaseQuery) !== -1);
      };
    }
```

miguelcobain:
@canufeel can you add this to the changelog so I can merge this?

canufeel:
@miguelcobain yeah, sure.

brendanoh:
Any idea when this will be merged? I see 2 other autocomplete related PR's as well.

canufeel:
@miguelcobain Please suggest if any other changes should be made.

PR #284, opened 2016-02-11
Uncomment paper-menu positioning for top/bottom and left/right

DanChadwick:
This allows the current paper-menu to be used for drop-down menus (of a sort).

### DIFF
--- a/addon/components/paper-menu.js
+++ b/addon/components/paper-menu.js
@@ -96,13 +96,13 @@ export default PaperMenuAbstract.extend({
       case 'target':
         position.top = existingOffsets.top + originNodeRect.top - alignTargetRect.top;
         break;
-      // Future support for mdMenuBar
-      // case 'top':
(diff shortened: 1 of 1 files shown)


miguelcobain:
@peec Do you have any idea of why that was commented? o.O

DanChadwick:
It appears to have been written as part of a port, in anticipation of a toolbar component, which hasn't been finished yet. Using this positioning is very useful for creating something close to an toolbar, but with icons. The biggest difference is that going from one menu to another requires two clicks -- the first to close one menu and the second to open another. While I'm waiting for the toolbar component, I'm satisfied with that behavior, and would love to see this positioning incorporated into the master (or WIP) branch.

DanChadwick:
Thank you *very* much for your dedication and contribution. I am very grateful. :+1:

miguelcobain:
You're welcome. :)

PR #288, opened 2016-02-17
paper-icon: add deprecation warning for old usage format

SaladFork:
We can support only the `{{paper-icon "someIcon"}}` format in `1.0.0` as it's a major version bump.

This adds a deprecation warning for those using version `<1.0.0`. It uses `Ember.deprecate` with a deprecation ID so it can be managed using [ember-cli-deprecation-workflow](https://github.com/mixonic/ember-cli-deprecation-workflow) for those who might want to remain on `<1.0.0` but not see the deprecations.

### DIFF
--- a/addon/components/paper-icon.js
+++ b/addon/components/paper-icon.js
@@ -6,12 +6,24 @@ var PaperIconComponent = Ember.Component.extend(ColorMixin, {
   classNames: ['paper-icon', 'md-font', 'material-icons', 'md-default-theme'],
   classNameBindings: ['iconClass', 'sizeClass', 'spinClass'],
 
-  icon: '',
   spin: false,
   reverseSpin: false,
 
   iconClass: Ember.computed('icon', 'positionalIcon', function() {
-    var icon = this.getWithDefault('positionalIcon', this.get('icon'));
+    let oldIcon = this.get('icon');
+
+    Ember.deprecate(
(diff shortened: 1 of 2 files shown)


miguelcobain:
since on `1.0.0` `{{paper-icon icon="check"}}` will also work, does this deprecation still makes sense? :confused:

joukevandermaas:
IMO since Glimmer components won't support positional params, Ember Paper should not deprecate the 'normal' invocation. That would make the eventually possible upgrade path nicer and at the very least maintain some consistency.

SaladFork:
If the format continues to work we absolutely should not add this warning, agreed. If positionalParams are going away completely with glimmer components we should rethink the recommended `paper-icon` usage. Probably `<paper-icon>icon-name</paper-icon>` or `<paper-icon icon="icon-name" />`.

miguelcobain:
`<paper-icon>icon-name</paper-icon>` and `{{paper-icon}}icon-name{{/paper-icon}}` only work because we're using ligatures.

For that reason, I think we should encourage the use of `<paper-icon icon="icon-name" />`, `{{paper-icon icon="icon-name"}}` or `{{paper-icon "icon-name"}}`. This way, if the stylesheets stop using ligatures for some reason, we don't need to change any API.

DanChadwick:
If I'm reading the above discussion, the consensus is that this PR should not be merged. Ok to close, @SaladFork?

SaladFork:
Thanks for the ping @DanChadwick; looks like we won't be needing this.

PR #289, opened 2016-02-17
[1.0.0] Remove old non-positional-param usage of paper-icon

SaladFork:
Also removes `iconClass` and just uses `icon` as it is no longer being used as a class.

If we're going to remove support for an existing syntax we should probably add a deprecation warning for older versions to help developers who are upgrading. See #288 for a PR addressing this.

### DIFF
--- a/addon/components/paper-icon.js
+++ b/addon/components/paper-icon.js
@@ -6,14 +6,8 @@ let PaperIconComponent = Ember.Component.extend(ColorMixin, {
   classNames: ['paper-icon', 'md-font', 'material-icons', 'md-default-theme'],
   classNameBindings: ['sizeClass', 'spinClass'],
 
-  icon: '',
   spin: false,
   reverseSpin: false,
-
-  iconClass: Ember.computed('icon', 'positionalIcon', function() {
-    let icon = this.getWithDefault('positionalIcon', this.get('icon'));
-    return icon;
-  }),
 
   spinClass: Ember.computed('spin', 'reverseSpin', function() {
     if (this.get('spin')) {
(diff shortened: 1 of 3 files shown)


miguelcobain:
AFAIK, this will still allow usage of `{{paper-icon icon="check"}}`, right?

When you do `positionalParams: ['icon']`, it makes `{{paper-icon "check"}}` identical as `{{paper-icon icon="check"}}` (which I don't mind).

So, I like this PR because it removes `iconClass` indirection.

Also, this doesn't change any public API, so no changelog entry is needed.

I'm going to ask you if you can `const { Component, computed } = Ember;` like we've been doing on the other components.
I believe that would change the paper-icon chapter for good.

Thanks!

miguelcobain:
This twiddle confirms that `{{paper-icon "check"}}` is identical as `{{paper-icon icon="check"}}`: https://ember-twiddle.com/f284a7729b08681f761e?openFiles=paper-icon.component.js%2Capplication.template.hbs

SaladFork:
I'm concerned about compatibility across Ember versions as the behavior differs. Using the code in your twiddle but changing Ember versions, here's what I'm seeing for the output of each invocation:

Invocation | Ember <= 2.0.2 | Ember 2.1.0 | Ember >= 2.1.1
---------------|---------------------|----------|----
`{{paper-icon "bla1"}}` | `bla1` | ERROR¹ |  `bla1`
`{{paper-icon icon="bla2"}}` | ` ` | ERROR¹ | `bla2`

¹ Uncaught Error: Assertion Failed: You cannot specify both a positional param (at position 0) and the hash argument `icon`.

DanChadwick:
We are using contextual components so ember1 compatibility isn't an issue. I think we are in agreement that both forms will be supported. Please reopen if I'm missing something.

PR #290, opened 2016-02-17
[1.0.0] Add aria-label support to `paper-icon`

### LINKED ISSUE #276: Accessibility:  Consider making paper-icon tag an `img` tag vs.  `md-icon`
We are considering using Paper in an educational app where accessibility is of high importance for our users. For the most part, this library cooperates with screen readers we've tested with.

We'd like to make use of many of the paper icons in our app. We've used several directly in buttons for image-like buttons. But since Paper icons aren't actually images, a screen reader has nothing to read (mainly the `alt` tag) to gather any context on the button's importance. 

It would be great if:
* paper icon's base tag was `img`
* The component bound on an `alt` property to allow clients to define a nice description for the button's utility.

SaladFork:
Closes #276.

Do we want to back-port this to `<1.0.0`?

I've also set it up to be a clean merge from current `wip/v1.0.0`. If #289 is merged, this will need to be adjusted (to use `icon` rather than `iconClass`).

### DIFF
--- a/addon/components/paper-icon.js
+++ b/addon/components/paper-icon.js
@@ -5,6 +5,7 @@ let PaperIconComponent = Ember.Component.extend(ColorMixin, {
   tagName: 'md-icon',
   classNames: ['paper-icon', 'md-font', 'material-icons', 'md-default-theme'],
   classNameBindings: ['sizeClass', 'spinClass'],
+  attributeBindings: ['aria-label'],
 
   icon: '',
   spin: false,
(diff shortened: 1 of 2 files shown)


miguelcobain:
> Do we want to back-port this to <1.0.0?

It would be great because of #276

miguelcobain:
Actually, I think that #289 could also be backported to <`1.0.0` along with this.

mellatone:
+1 on merging this

PR #293, opened 2016-02-22
Updated paper-switch component

jorgelainfiesta:
Hi,

I'm opening this so you can start reviewing my changes. This is my first PR so I'd like to know if it's fine. I need help to test the Hammer.js behaviors on this component. I made some research on how to mock the gestures but still don't know what's the correct approach for the project.

Thanks,
Jorge L


addon/components/paper-switch.js
```diff
@@ -84,12 +92,13 @@ export default BaseFocusable.extend(RippleMixin, ProxiableMixin, ColorMixin, {
   },
```
miguelcobain: Instead of using jquery's `css()` and `removeAttr()` functions on `.md-thumb-container`, let's use a computed property that returns the style and bind it to the style attribute on the element like 

```hbs
<div class="md-thumb-container" style={{thumbContainerStyle}}>
  <!-- inner elements -->
</div>
```

addon/components/paper-switch.js
```diff
@@ -25,12 +25,35 @@ export default BaseFocusable.extend(RippleMixin, ProxiableMixin, ColorMixin, {
...
+      this.switchHammer.destroy();
+    }
+    if (this.thumbElementHammer) {
+      this.switchHammer.destroy();
+    }
+  },
 
+  disabledDidChange: observer('disabled', function() {
+    this._setupSwitch();
+  }),
```
miguelcobain: Let's use the `didReceiveAttrs` instead of this observer.
More info: https://guides.emberjs.com/v2.2.0/components/the-component-lifecycle/#toc_formatting-component-attributes-with-code-didreceiveattrs-code

> Since the didReceiveAttrs hook is called every time a component's attributes are updated whether on render or re-render, you can use the hook to effectively **act as an observer**, ensuring code is executed every time an attribute changes.

addon/components/paper-switch.js
```diff
@@ -47,21 +70,6 @@ export default BaseFocusable.extend(RippleMixin, ProxiableMixin, ColorMixin, {
     let switchHammer = new Hammer(element);
     this.switchHammer = switchHammer;
```
miguelcobain: let's prefix hammer variables with `_` e.g `this._switchHammer` and `this._thumbElementHammer`.

miguelcobain: Also, let's destructure `run` at the top, so that we use `run.bind` instead of `Ember.run.bind`.

addon/components/paper-switch.js
```diff
@@ -69,11 +77,11 @@ export default BaseFocusable.extend(RippleMixin, ProxiableMixin, ColorMixin, {
   },
 
   _drag(event) {
-    if (this.get('disabled')) { return; }
+    if (this.getAttr('disabled')) { return; }
```
miguelcobain: Let's use normal `get` for now, to conform with the rest of the project and in the light of this article: http://locks.svbtle.com/to-attrs-or-not-to-attrs

Please point out your reasons on the main issue, if you find appropriate.

addon/components/paper-switch.js
joukevandermaas: if you do that, you have to return a `SafeString` from the computed and use triple curlies.

miguelcobain: Safestring yes, but you don't need triple curlies.

Example:
  - https://github.com/miguelcobain/ember-paper/blob/wip/v1.0.0/addon%2Fcomponents%2Fpaper-progress-circular.js#L43-L49
  - https://github.com/miguelcobain/ember-paper/blob/wip/v1.0.0/app%2Ftemplates%2Fcomponents%2Fpaper-progress-circular.hbs#L1

jorgelainfiesta:
Thanks for the comments Miguel! I've done most of them, but I'm now dealing with issues that the changes created.

I was honestly lost in your conversation with Chris (runinspired) so I don't know what to do with Hamme now.

jorgelainfiesta:
I made the changes but now the sliding is really unstable. Not sure of what's causing it. The tests pass because they don't test Hammer gestures.

miguelcobain:
I'm stocked that jscs tests are passing here. You should be able to write `const` in functions.

jorgelainfiesta:
I didn't know that. I actually rarely use `let` because most of the time I won't ever change the variable value.

miguelcobain:
JSCS tests are failing because of: https://github.com/DockYard/ember-suave/issues/106#issuecomment-187720996

If JSCS tests were working you could not use `const` inside functions. Reasoning: http://madhatted.com/2016/1/25/let-it-be

jorgelainfiesta:
Thanks for the info. I'll definitely change it today.

miguelcobain:
Please run `jscs -c .jscsrc addon` to see your jscs errors (after `npm install -g jscs`).
You might see other errors related with other files. Ignore them please for now.

jorgelainfiesta:
I tried pulling wip/v1.0.0 and rebasing my branch to resolve conflicts but git says I'm up to date.

jorgelainfiesta:
Hi @miguelcobain, have you guys decided how we'd handle the hammer stuff? or should I try another component?

miguelcobain:
@jorgelainfiesta 

1. you need to rebase your branch to reapply upstream commits to make this mergeable again
2. Let's use vanilla HammerJS for now. We can switch to an ember addon later when things are more clear.

HammerJS recognizes gestures. We use it to recognize the pan action to move the switch.
Notice that the original angular material code binds the gestures in different components than this branch.

miguelcobain:
I've finished the above stuff. Thanks.

jorgelainfiesta:
Sorry, thanks for fixing it. I totally messed up my repo trying to do it, so I'd get some help from a friend by tomorrow. I'm glad it's merged now.

PR #300, opened 2016-02-26
Fix #299: Regression cause by array spread with jQuery objects.

### LINKED ISSUE #299: Making ember paper stylish introduces regression(s). E.g. paper-menu is broken.
The commit 2b483d2 "ember-paper is now stylish" introduces one and probably more regressions.

```javascript
     positionMenu(el) {
       // containerNode = wrapper
 -     var containerNode = el[0],
 +     let [containerNode] = el;
...
```
`el` is a jquery object representing the component. To spread `el` into `containerNode`, Babel generates a shim which accepts either an actual array or something with a Symbol.iterator. The jQuery object is neither, so an exception is thrown.

If you look at the hideous shim generated, I would suggest reverting this change. I realize the "ember-suave" standard prefers spreading, but it is both slower and in this case broken.

I have not searched for other similar changes, but anywhere a jQuery object is being referenced, it should not be treated as an array. `el[0]` is really `el.0`, except that `0` is not a valid identifier.

DanChadwick:
#299 reports a regression where `let [x] = someJqueryObject;` raises an exception because Babel expects the right-side expression to be an actual Array or iterable object, which jQuery objects aren't

### DIFF
--- a/addon/components/paper-autocomplete-list.js
+++ b/addon/components/paper-autocomplete-list.js
@@ -79,7 +79,7 @@ export default Ember.Component.extend({
     }
 
     let ul = this.$();
-    let li  = ul.find(`li:eq(${this.get('selectedIndex')})`)[0];
+    let li  = ul.find(`li:eq(${this.get('selectedIndex')})`).get(0);
     let top = li.offsetTop;
     let bot = top + li.offsetHeight;
     let hgt = ul[0].clientHeight;
--- a/addon/components/paper-autocomplete-list.js
+++ b/addon/components/paper-autocomplete-list.js
@@ -79,7 +79,7 @@ export default Ember.Component.extend({
     }
 
     let ul = this.$();
-    let [li]  = ul.find(`li:eq(${this.get('selectedIndex')})`);
+    let li  = ul.find(`li:eq(${this.get('selectedIndex')})`)[0];
     let top = li.offsetTop;
     let bot = top + li.offsetHeight;
     let hgt = ul[0].clientHeight;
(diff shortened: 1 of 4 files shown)


DanChadwick:
Testing failure is unrelated to this PR.

miguelcobain:
@DanChadwick if you take a close look, you're getting `requireArrayDestructuring` jscs errors at the end.
That's because jscs rules think you must destructure. However, like you've pointed out, jquery objects can't be destructured, so jscs is telling us to do something that isn't possible. :D

I've come across this before, and the solution was this: https://github.com/miguelcobain/ember-paper/blob/wip/v1.0.0/tests%2Fintegration%2Fcomponents%2Fpaper-progress-linear-test.js#L26

Basically use `get(0)`.

Good catch on this! I didn't worry too much because someone would notice bugs when updating that components (the component wasn't even tested).

Please fix that on this PR and we can merge. Thanks!

DanChadwick:
Ah right. I can do as you request, and considered it, but it is more brittle than using array dereferencing because the code now assumed the passed element is a jQuery object, rather than something that is array-like. Perhaps an unlikely scenario...

The alternative would be to override the warnings with:
```
// thingie may be a jQuery object (which does not support spread), rather than an actual array
// jscs:disable requireArrayDestructuring
let [subthingie] = thingie;
```

miguelcobain:
> it is more brittle than using array dereferencing because the code now assumed the passed element is a jQuery object

In all your changes it is a jquery object (let me know if I'm wrong). I'd say to keep it simple for now, while conforming to the code style rules.

miguelcobain:
Thanks!

DanChadwick:
Miguel, thank you very much for your hard work on this module, as well as your nearly instantaneous response to my recent PRs. This sort of support is both highly appreciated and very encouraging to contributors who might take the extra step from reporting an issue to fixing it. Really excellent!

miguelcobain:
@DanChadwick I wish I had more time for ember-paper. Thanks for your kind works. :)

Meet us at #ember-paper channel on ember's slack. :)
:beers:

PR #301, opened 2016-02-26
Fix #193 by zkwentz. Allow scss override via !default.

### LINKED ISSUE #193: Customize/Override default scss theme?
I see in the sources that the project is designed to work with theme, but I don't see any way to see a `THEME` variable somewhere.

What is the proper way to override the default theme?
Is there a way to activate my own theme after I've set the `theme-name` in my style file like following:

```scss
//app/styles/custom.scss

$theme-name: "my-theme";

$primary: "red";
$accent: "yellow";
$warn: "brown";
$background: "red";
$link: "blue";
...
```

DanChadwick:
This is zkwentz's PR to allow style color variables to be overridden by declaring them with !default. This PR applies to the WIP/1.0.0 branch, rather than master. I believe that active new features are being added in this branch.

Credit to zkwentz, not me. :+1:

### DIFF
--- a/app/styles/color-palette.scss
+++ b/app/styles/color-palette.scss
@@ -1,8 +1,8 @@
 // Utility Color Classes
 
-$light-contrast-color: rgba(255, 255, 255, 0.87);
-$dark-contrast-color: rgba(0, 0, 0, 0.87);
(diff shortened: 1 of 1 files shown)


miguelcobain:
:+1:

PR #303, opened 2016-02-26
Allow paper-menu's to be displayed with md-dense.

DanChadwick:
Angular defines a md-dense style, but because it needs to be applied in the paper-menu-content-pane, which is in the paper-menu-content, which is in turn in the paper-menu, there is no easy way for md-dense to be specified. This PR adds a "dense=true" option to paper-menu, which "bubbles down" (sinks??) down to the paper-menu-content-pane, where it adds the required class.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - [#286](https://github.com/miguelcobain/ember-paper/pull/286) paper-radio no longer has the class `paper-radio`
 - [#286](https://github.com/miguelcobain/ember-paper/pull/286) paper-radio now uses `groupValue` instead of `selected`
 - [#286](https://github.com/miguelcobain/ember-paper/pull/286) paper-radio now sends the action `onchange` instead of `changed` and it is mandatory (see ddau)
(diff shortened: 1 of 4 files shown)


DanChadwick:
Again, test failure is unrelated to this PR.

miguelcobain:
The `md-dense` class is only applied in the menu bar directive: https://material.angularjs.org/latest/demo/menuBar

I'm divided. On one side, we should encourage following the material design spec. And it looks like the material design spec doesn't have "dense" menus outside menu bars? (need confirmation)
On the other hand, this would make the component slightly more flexible, at little cost (the styles are there). Also, I'm afraid that AM suddenly ceases to support that outside menu bars and we're left with a burden because we've provided it to people in the first place.

I think this could be added, as it will help us when we develop menu-bars.

miguelcobain:
Ah, looks like material design mentions nothing about dense lists only on menu bars: https://www.google.com/design/spec/components/lists.html#lists-specs

@DanChadwick can you please add an entry to CHANGELOG.md pointing to this issue?
I'm trying not to lose track of what's added in v1.0.0. People will need that later.

DanChadwick:
Changelog updated, and again the failure is unrelated to this PR.

DanChadwick:
Thanks again. And on a Sunday, no less.

PR #304, opened 2016-02-29
auto-complete: added guards to fix errors

baileymiller182:
I had a lot of trouble with `{{paper-autocomplete}}` when following the documentation here: http://miguelcobain.github.io/ember-paper/#/autocomplete. I was not doing anything extremely novel or divergent from given examples. I had a very simple component definition, returning a promise as the source. 

Issues:
- Undefined `searchText` in `isMinLengthMet`, triggered when focusing on the component.
- `modelDidChange` will attempt to look up the label of a nulled out model. The odd symptom of this is that the first character typed into the input does not appear.
- `observeIndex` may be triggered before suggested elements appear in the DOM, triggering errors on either `ul.find` or `li.offsetTop`.

I found that a few guards were missing, and adding just a couple of additional checks makes the component more robust (at least for my needs).

### DIFF
--- a/addon/components/paper-autocomplete-list.js
+++ b/addon/components/paper-autocomplete-list.js
@@ -78,11 +78,21 @@ export default Ember.Component.extend({
       return;
     }
 
-    var ul = this.$(),
-      li  = ul.find('li:eq('+this.get('selectedIndex')+')')[0],
-      top = li.offsetTop,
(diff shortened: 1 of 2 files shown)


xomaczar:
Paper-menu branch addresses this. Let's close it

PR #308, opened 2016-03-12
Added tests and further implementation of paper-dialogs

mellatone:
Did

- [x] Refactored `paper-wormhole` to key off the rootElement 
- [x] Add basic test around dialogs
- [x] Dialog support for scoping backdrop and container to parent selector
- [x] Modified docs to highlight dialog scoping
- [x] Removed alert and confirm examples
- [x] Key event handling: escape to close

Still outstanding
- [ ] Implement tests and functionality for transitions
- [ ] Button focusing
- [ ] Key event handling: return on focused button should click the button.


mellatone:
I feel like "enter on focused button to trigger action" is best as a `{{paper-button}}` concern because modals may contain forms that need complex tab indexing.

miguelcobain:
Can you explain me the reasoning behind removing from `contentFor` and adding an initializer?

mellatone:
@miguelcobain There's a few reasons I have for suggesting this change. The changes are specifically around refactoring how `<div id="ember-wormhole"></div>` is set up. If `{{content-for}}` has other use that I didn't consider, please let me know.

- **It's testable**. While hard-coding `<div id="ember-wormhole"></div>` into the template through the build pipeline works for now, there's an assumption that the wormhole is there. The initializer is testable, and because a lot of other components depend on it, it feels better to make sure that it works.
- **It respects the app's environment**. Placing the wormhole underneath the body, assumes that this is where the application lives and where components that use wormhole should be rendered. This may not always be the case. The `rootElement` is a more accurate representation of that.
- **It better supports embedded applications**. One practical example of this is documentation, where we want to demo an embedded application. Wormhole'd components should be rendered relative to this. Another example of this is integration and acceptance testing. Wormhole'd components are being rendered into QUnit's body element, outside of the testing container.
- **This technique has precedent**. See: https://github.com/yapplabs/ember-modal-dialog/blob/master/addon/initializers/add-modals-container.js

miguelcobain:
@mellatone point given!

@cibernox maybe you're interested?

PR #309, opened 2016-03-12
paper-select-menu: prevent scrolling whole page when focusing items in menu by keyboard

jankopriva:


miguelcobain:
Doesn't appear to be happening on new menus. Please reopen if issue persists.

PR #311, opened 2016-03-15
Move `paper-wormhole` div out of head tag.

blimmer:
The `paper-wormhole` div is currently insterted into the `<head>` of the index.html page,
which causes DOM parsing problems for browsers. The `div` element cannot appear in the
`<head>`, and causes Chrome to move the `<body>` tag up to enclose the first `<div>`, breaking
all subsequent `<head>` tags (favicons meta, etc.)

This change places the tag as the first element in the `<body>` tag, which should
be equivalent for Paper.


blimmer:
This can definitely cause some really strange bugs (which is how I found this bug in the first place), so I think this warrants a bugfix release.

miguelcobain:
I don't know how this ended up like that this far.
Do you mind adding a change log entry under master?

blimmer:
Updated changelog and amended commit.

miguelcobain:
Thanks!

miguelcobain:
Published 0.2.12

blimmer:
Thanks for handling this so quickly!

PR #313, opened 2016-03-18
Push paper-dialog forward. Still WIP.

DanChadwick:
* Fixes escape to close.
* Fixes click outside dialog to close. Required reinstating `paper-dialog-container`.
* Convert named-by-string actions to action closures.
* Remove vestigial `paper-dialog-parent`.
* Improves documentation and dummy app.
* Update other components to conform to prior API change to paper-backdrop tap action.
* **Temporarily disabled transitions. Transitions do not work and prevent functioning.** When transitions work again. the commit for this should be reversed. Changes are marked with TODO.
* Added "prompt dialog" example.
* Applies fixed positioning to backdrop if no parent is defined. This satisfies the automated test and AM behavior, but at least on desktop, I can detect no different in actual browser behavior.
* Fixed fullscreen (test passes now).
* Updated dummy app to show primary-colored buttons and demonstrate fullscreen.


DanChadwick:
Test failures are unrelated to the PR. If transitions are disabled, only two tests fail, but unrelated:
* not ok 65 PhantomJS 2.0 - Integration | Component | paper dialog: should prevent scrolling entirely behind fixed modal
* not ok 69 PhantomJS 2.0 - Integration | Component | paper dialog: has opt-in support for fullscreen at responsive breakpoint

DanChadwick:
Updated top comment to reflect the last 4 commits. This branch should now be functional again, but with transitions disabled.

addon/components/paper-dialog-inner.js
```diff
@@ -4,10 +4,10 @@ import PaperDialog from './paper-dialog';
 
 const { Component, computed } = Ember;
 
-export default Component.extend(Translate3dMixin, {
+export default Component.extend(/*Translate3dMixin, TODO */ {
   tagName: 'md-dialog',
-  classNames: ['md-default-theme'],
-  classNameBindings: ['contentOverflow:md-content-overflow', 'fullscreen:md-dialog-fullscreen'],
+  classNames: ['md-default-theme'     , 'md-transition-in'     /* TODO */ ],
+  classNameBindings: ['contentOverflow:md-content-overflow', 'dialogComponent.fullscreen:md-dialog-fullscreen'],
```
miguelcobain: Can't we get rid of `dialogComponent` properties by invoking `paper-dialog-inner` as
```hbs
{{#paper-dialog-inner fullscreen=fullscreen clickOutsideToClose=clickOutsideToClose ...
```
?

Since `paper-dialog-inner` is a private component, we can use it like we want to. No need for a `nearestOfType` dependency, probably.

DanChadwick: We use dialogComponent in `paper-inner-dialog` for 4 different things. We could pass it by binding `this`, we could bind all four things we use, or we can keep what we have. I thought the discussion from `tabs` was that it was better to connect the two components than to bind a bunch of properties between two components.

miguelcobain: That discussion does not apply for private components.
We had to resort to such mechanisms because we can't expect the user to pass them by us.

PR #322, opened 2016-03-25
Update inverse template for autocomplete as it doesnt accept parameters

kat3kasper:
Inverse component templates can not recieve parameters. This means
there is no access to `searchText` in the inverse block.

- Try to render notFoundComponent then inverse block then default notFoundMsg
- Update documentation app to reflect that searchText is not available in inverse bloc.

See: https://github.com/emberjs/ember.js/pull/11084

### DIFF
--- a/app/templates/components/paper-autocomplete.hbs
+++ b/app/templates/components/paper-autocomplete.hbs
@@ -58,11 +58,13 @@
 
         {{else}}
           {{#if showLoadingBar}}
-          {{!-- Render named component then default --}}
+          {{!-- Render named component, then inverse block then default --}}
             {{#if notFoundComponent}}
-                <li>{{component notFoundComponent searchText=searchText}}</li>
+              <li>{{component notFoundComponent searchText=searchText}}</li>
+            {{else if (has-block "inverse")}}
+              <li>{{yield to="inverse"}}</li>
             {{else}}
-                <li>{{notFoundMsg}}</li>
+              <li>{{notFoundMsg}}</li>
             {{/if}}
           {{/if}}
         {{/each}}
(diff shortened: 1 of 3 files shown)


app/templates/components/paper-autocomplete.hbs
```diff
@@ -58,15 +58,11 @@
...
+            {{#if notFoundComponent}}
+                <li>{{component notFoundComponent searchText=searchText}}</li>
             {{else}}
-              {{#if notFoundComponent}}
-                  <li>{{component notFoundComponent searchText=searchText}}</li>
-              {{else}}
-                  <li>{{notFoundMsg}}</li>
-              {{/if}}
+                <li>{{notFoundMsg}}</li>
             {{/if}}
```
miguelcobain: I think we can be a bit more friendlier and backwards compatible at the same time.
What do you think about:
```hbs
{{#if notFoundComponent}}
  <li>{{component notFoundComponent searchText=searchText}}</li>
{{else if (has-block "inverse")}}
  <li>{{yield to="inverse"}}</li>
{{else}}
  <li>{{notFoundMsg}}</li>
{{/if}}
```

miguelcobain:
Can you please add a new CHANGELOG entry under a new version named `master`?

kat3kasper:
@miguelcobain took the suggestion for not breaking backwards compatibility. I like this idea and implemented it. My only concern is that it doesn't match the ordering of what happens when an item is found. There we follow hasBlock, itemComponent, default. I think it will still work as people expect though either implementing an inverse block or a notFoundComponent.

miguelcobain:
I think it's ok since people will generally use just one of them. Thanks again.

PR #323, opened 2016-03-25
Pass item through as expected rather than label

kat3kasper:


miguelcobain:
Thank you for the PR!
Can you please add a new CHANGELOG entry under a new version named master?

kat3kasper:
@miguelcobain I believe this is what you are asking for

miguelcobain:
Perfect. Thanks.

miguelcobain:
@kat3kasper if you're using ember-paper, we may need your help :)

Join us at #e-paper in ember's slack. :+1:

PR #324, opened 2016-03-27
Updated paper-button

jorgelainfiesta:
Hi,

Here's the PR for the update to `paper-button`, jscs and tests are passing. I'll thank you all for the comments. 

Jorge L


addon/components/paper-button.js
```diff
@@ -4,45 +4,53 @@ import RippleMixin from '../mixins/ripple-mixin';
...
+  fitRipple: computed.readOnly('isIconButton'),
+  center: computed.readOnly('isIconButton'),
+  dimBackground: computed.not('isIconButton'),
 
-  noSpan: Ember.computed('no-span', function() {
-    return this.get('no-span');
-  }),
+  noSpan: computed.readOnly('no-span'),
 
   // RippleMixin overrides
```
miguelcobain: Remove this comment.

addon/components/paper-button.js
```diff
@@ -4,45 +4,53 @@ import RippleMixin from '../mixins/ripple-mixin';
...
-  isIconButton: Ember.computed(function() {
-    return this.classNames.any(function(className) {
-      return className.indexOf('md-icon-button') !== -1;
-    });
+  isIconButton: computed({
+    get() {
```
miguelcobain: let's use normal get-only computeds to conform with the rest of the project.

addon/components/paper-button.js
```diff
@@ -4,45 +4,53 @@ import RippleMixin from '../mixins/ripple-mixin';
...
 import ColorMixin from 'ember-paper/mixins/color-mixin';
 
+const {
+  assert,
+  computed,
+  isPresent
+} = Ember;
+
 export default BaseFocusable.extend(RippleMixin, ProxiableMixin, ColorMixin, {
   attributeBindings: ['target', 'action', 'type'],
```
miguelcobain: Why do we need to bind an `action` attribute? also, `target`?

miguelcobain:
create the test:
>uses `md-icon-button` class when `iconButton=true`

addon/components/paper-button.js
```diff
@@ -4,45 +4,53 @@ import RippleMixin from '../mixins/ripple-mixin';
...
   tagName: 'button',
   themed: true,
-  classNameBindings: ['raised:md-raised', 'icon-button:md-icon-button', 'focus:md-focused', 'themed:md-default-theme', 'themed:md-button'],
+  classNameBindings: [
+    'raised:md-raised',
+    'icon-button:md-icon-button',
+    'focus:md-focused',
+    'themed:md-default-theme',
+    'themed:md-button'
```
miguelcobain: We're adding `md-default-theme` by default in other components. e.g https://github.com/miguelcobain/ember-paper/blob/wip/v1.0.0/addon%2Fcomponents%2Fpaper-switch.js#L18

miguelcobain:
create the test:
> uses `md-raised` class when `raised=true`

addon/components/paper-button.js
```diff
@@ -4,45 +4,53 @@ import RippleMixin from '../mixins/ripple-mixin';
...
+
 export default BaseFocusable.extend(RippleMixin, ProxiableMixin, ColorMixin, {
   attributeBindings: ['target', 'action', 'type'],
   tagName: 'button',
   themed: true,
-  classNameBindings: ['raised:md-raised', 'icon-button:md-icon-button', 'focus:md-focused', 'themed:md-default-theme', 'themed:md-button'],
+  classNameBindings: [
+    'raised:md-raised',
+    'icon-button:md-icon-button',
+    'focus:md-focused',
```
miguelcobain: this is pointless. BaseFocusable already has such a binding: https://github.com/miguelcobain/ember-paper/blob/wip/v1.0.0/addon%2Fcomponents%2Fbase-focusable.js#L14

Also, `focus` was renamed to `focused`. Anyway, this logic is no longer handled in paper-button.

addon/components/paper-button.js
miguelcobain: I'm wondering why do we need this computed since we have `iconButton` property.

miguelcobain:
Thanks for the PR! I've added some notes. Thanks.

jorgelainfiesta:
Thanks for the comments! Sorry I'm so slow, have had very busy days. I'll push the changes later today.

addon/components/paper-button.js
jorgelainfiesta: Nothing really, I'll remove them.

addon/components/paper-button.js
jorgelainfiesta: I honestly just left this property there in case it was used by some parent class, but seems pointless from what I've seen. Should I remove it?

addon/components/paper-button.js
jorgelainfiesta: Ok

addon/components/paper-button.js
jorgelainfiesta: With normal get-only you mean `computed(function() {})`, right?

addon/components/paper-button.js
jorgelainfiesta: Ok done. I also added `paper-button` class.

addon/components/paper-button.js
miguelcobain: We should remove this. And I'm also not sure about `isMenuItem`.
I know it has something to do with list controls, but when we update them, we should solve that problem there.

`isMenuItem` doesn't look like it belongs here. Nothing references it inside button.

jorgelainfiesta: Sure thing, I'll remove them.

miguelcobain: > With normal get-only you mean computed(function() {}), right?

Yes, I did. But now they will be removed. :stuck_out_tongue:

jorgelainfiesta:
Changes are up :)

addon/components/paper-button.js
```diff
@@ -4,45 +4,36 @@ import RippleMixin from '../mixins/ripple-mixin';
...
+  attributeBindings: ['type'],
   tagName: 'button',
+  classNames: ['paper-button', 'md-default-theme'],
   themed: true,
-  classNameBindings: ['raised:md-raised', 'icon-button:md-icon-button', 'focus:md-focused', 'themed:md-default-theme', 'themed:md-button'],
+  classNameBindings: [
+    'raised:md-raised',
+    'iconButton:md-icon-button',
+    'themed:md-default-theme',
+    'themed:md-button'
```
miguelcobain: these classes should be hardcoded into `classNames` directly.
Remove `themed` property.

addon/components/paper-button.js
```diff
@@ -4,45 +4,36 @@ import RippleMixin from '../mixins/ripple-mixin';
...
-  fitRipple: Ember.computed.readOnly('isIconButton'),
-  center: Ember.computed.readOnly('isIconButton'),
-  dimBackground: Ember.computed.not('isIconButton'),
-
-  noSpan: Ember.computed('no-span', function() {
-    return this.get('no-span');
-  }),
+  fitRipple: computed.readOnly('isIconButton'),
+  center: computed.readOnly('isIconButton'),
+  dimBackground: computed.not('isIconButton'),
```
miguelcobain: `isIconButton` doesn't exist anymore, right?
We use `iconButton` directly now.

addon/components/paper-button.js
```diff
@@ -4,45 +4,36 @@ import RippleMixin from '../mixins/ripple-mixin';
...
 
+const {
+  assert,
+  computed,
+  isPresent
+} = Ember;
+
 export default BaseFocusable.extend(RippleMixin, ProxiableMixin, ColorMixin, {
-  attributeBindings: ['target', 'action', 'type'],
+  attributeBindings: ['type'],
```
miguelcobain: let's make the default value of `type` to `'button'`.
This way we won't trigger forms by default. That's also how angular material does.

addon/components/paper-button.js
```diff
@@ -4,45 +4,36 @@ import RippleMixin from '../mixins/ripple-mixin';
...
-    return this.classNames.any(function(className) {
-      return className.indexOf('md-icon-button') !== -1;
-    });
-  }),
-  isMenuItem: Ember.computed(function() {
-    return this.classNames.any(function(className) {
-      return className.indexOf('md-menu-item') !== -1;
-    });
-  }),
+  noSpan: computed.readOnly('no-span'),
```
miguelcobain: I don't know were this came from, but let's never use a `<span`. Angular Material uses it because it has to do with the way it does bindings, I think.
Also remove from paper-button template.

tests/integration/components/paper-button-test.js
```diff
@@ -0,0 +1,65 @@
...
+});
+
+test('uses md-icon-button class when iconButton=true', function(assert) {
+  this.render(hbs`
+    {{#paper-button iconButton=true}}
+      A label
+    {{/paper-button}}
+  `);
+  assert.ok(this.$('.md-button').hasClass('md-icon-button'));
+});
```
miguelcobain: Let's add a little test to ensure that the default button `type` attr is `'button'`.

miguelcobain:
@jorgelainfiesta I've reviewed again. I think that is all. It will be an awesome button! :smile:

miguelcobain:
@jorgelainfiesta ping

jorgelainfiesta:
Oh I forgot to push changes (facepalm)

jorgelainfiesta:
Oh apparently I only made the changes in my mind, sorry. Doing changes now, should take only a few mins.

addon/components/paper-button.js
```diff
@@ -4,45 +4,32 @@ import RippleMixin from '../mixins/ripple-mixin';
...
   click(event) {
-    let target = this.get('target');
-
-    if (target) {
-      this.get('target').send(this.get('action'), this.get('param'), event);
-    } else {
-      this.sendAction('action', this.get('param'), event);
+    let onClick = this.get('onClick');
+    if (isPresent(onClick)) {
+      this.get('onClick')(event);
```
miguelcobain: Please use `this.sendAction('onClick')`:
1. it is compatible with closure actions
2. you don't need to check if an action exists that way

addon/components/paper-button.js
```diff
@@ -4,45 +4,32 @@ import RippleMixin from '../mixins/ripple-mixin';
...
-  attributeBindings: ['target', 'action', 'type'],
+  attributeBindings: ['type'],
+  type: 'button',
   tagName: 'button',
-  themed: true,
-  classNameBindings: ['raised:md-raised', 'icon-button:md-icon-button', 'focus:md-focused', 'themed:md-default-theme', 'themed:md-button'],
+  classNames: ['paper-button', 'md-default-theme', 'md-button'],
+  classNameBindings: [
+    'raised:md-raised',
+    'iconButton:md-icon-button',
```
miguelcobain: please remove trailing comma

jorgelainfiesta:
All changes are up :)

addon/components/paper-button.js
```diff
@@ -4,45 +4,30 @@ import RippleMixin from '../mixins/ripple-mixin';
...
-    let target = this.get('target');
-
-    if (target) {
-      this.get('target').send(this.get('action'), this.get('param'), event);
-    } else {
-      this.sendAction('action', this.get('param'), event);
-    }
-
-    return this.get('bubbles');
+    let onClick = this.get('onClick');
```
miguelcobain: `onClick` is unused.

jorgelainfiesta: Sorry, missed it.

PR #325, opened 2016-03-28
Dialog API buff-up.

DanChadwick:
`parent`, `origin`, `openFrom`, `closeTo` are all optional and all take a jQuery selector or jQuery object.

I found the original API inconsistent and inconvenient in actual use. This PR normalized the API and cleans up some old code remnants. 

Note that `parent` is *not* an id, but a jQuery selector or object. An id is created, if needed, for `{{ember-wormhole}}`.

### DIFF
--- a/addon/components/paper-dialog.js
+++ b/addon/components/paper-dialog.js
@@ -1,5 +1,13 @@
 import Ember from 'ember';
-const { Component, computed, $, inject: { service } } = Ember;
+import { toJQuery } from '../mixins/translate3d-mixin';
+
+const {
+  Component,
+  computed,
+  inject: { service },
+  isEmpty,
+  guidFor
+} = Ember;
 
 export default Component.extend({
   tagName: '',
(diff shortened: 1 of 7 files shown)


DanChadwick:
Test failure seems to be a travis error. It can't install phantomjs for ember-release, but default, beta, and canary all pass fine.

addon/components/paper-dialog.js
```diff
@@ -7,25 +15,40 @@ export default Component.extend({
...
+    }
+    return id;
+  }),
 
   constants: service(),
 
   didInsertElement() {
     if (this.get('escapeToClose')) {
-      let parent = this.get('parentElementSelector');
-      $(parent).on(`keydown.${this.elementId}`, (e) => {
```
miguelcobain: Why this change?

```diff
+this.get('defaultedParent')
-this.get('parentElementSelector')
```

`defaultedParent` is the new `parentElementSelector`?

addon/mixins/translate3d-mixin.js
```diff
@@ -59,8 +57,10 @@ export default Mixin.create({
       // Wait while CSS takes affect
       // Set the `main` styles and run the transition-in styles
       run.next(() => {
-        this.waitTransitionEnd(this.element).then(() => {
-          this.onTranslateFromEnd();
+        this.waitTransitionEnd($(this.element)).then(() => {
+          if (!this.get('isDestroying') && !this.get('isDestroyed')) {
```
miguelcobain: What is the reason for this guard?
For situations where you create and destroy very quickly?
What was the problem you were having?

addon/components/paper-dialog.js
mellatone: +1, `parentElement` || `parentElementSelector` is a very clear descriptor.

miguelcobain: I'm not complaining about the names. Just trying to understand.

PR #328, opened 2016-03-31
paper-menu: update paper-menu-abstract to ES6 and add onOpenMenu/onCloseMenu

DanChadwick:
This is just a minor update to this one file to allow my development to proceed. It adds two actions: `onMenuOpen` and `onMenuClose`. I would have named these `onOpen` and `onClose` but for the existing attribute `onMenu`, which accepts a promise that yields menu items.

@miguelcobain: My hope is to merge this into `wip/v1.0.0` as a minor update, rather than taking on all of the menu-related components right now.

To-do:
Consider renaming `onMenu` to `getItems` so that the actions can be `onOpen` and `onClose`.

### DIFF
--- a/addon/components/paper-menu-abstract.js
+++ b/addon/components/paper-menu-abstract.js
@@ -1,15 +1,17 @@
 import Ember from 'ember';
 
-export default Ember.Component.extend({
-  constants: Ember.inject.service(),
+const { Component, computed: { alias }, inject: { service }, assert } = Ember;
+
+export default Component.extend({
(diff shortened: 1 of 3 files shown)


DanChadwick:
Found a probably-long-standing bug: If you click a menu item that sends a action which causes that menu item to become disabled, the menu never closes because it fails to process the click on the disabled menu item. This is caused by duplicated effort in `paper-menu-content`. Instead, `paper-item` will `send` and action to toggle the menu closed.

Issue #224, opened 2015-11-14
Failed to find a valid digest in the 'integrity' attribute for resource


robinboening:
I get this error on heroku (not locally)

```
Failed to find a valid digest in the 'integrity' attribute for resource 'http://rubyissues.herokuapp.com/assets/rubyissues-dc683ef8d41f45e5fdafceda1a26598f.css' with computed SHA-256 integrity 'vGH6uxiZBXnDr46nPJPx2/V82if1Sij2FDhPgXSk6xI='. The resource has been blocked.
```

It seems to be similar to this: https://github.com/ember-cli/ember-cli/issues/5040, but I can not fix it by adding an empty `integrity` attribute to the `link` tag. (I can see another integrity attribute is added automatically)

I am posting here because I broke the issue down to one of the `@import` statements in my `app.scss`:
The `@import 'ember-paper';` - When I remove this line its working.

Seems to be an issue in chrome only. No problem when using Firefox.

mike1o1:
I had a similar issue running in production, and just ended up adding the empty integrity tag to every script and link tag. It resolved it for me. Don't think there's much we can do on our end here.

Here is what my index works, and this is working fine for me. Did you add it to all tags?

*Note I haven't tested with the latest ember-cli, but this is working for me*

```html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8">
    <meta http-equiv="X-UA-Compatible" content="IE=edge">
    <title>Client</title>
    <meta name="description" content="">
    <meta name="viewport" content="width=device-width, initial-scale=1">

    {{content-for 'head'}}

    <link rel="stylesheet" href="assets/vendor.css" integrity="">
    <link rel="stylesheet" href="assets/client.css" integrity="">

    {{content-for 'head-footer'}}
  </head>
  <body layout="row">
    {{content-for 'body'}}

    <script src="assets/vendor.js" integrity=""></script>
    <script src="assets/client.js" integrity=""></script>

    {{content-for 'body-footer'}}
  </body>
</html>
```

mansona:
This issue is probably caused by some encoding problem somewhere in the included files, take a look here to follow the progress of the bug https://github.com/jonathanKingston/ember-cli-sri/issues/5 . 

The simplest way to "solve" this while we're waiting for it to get fixed is to just remove the `"ember-cli-sri": "^1.0.1",` line from your package.json and it will no longer add the integrity attribute :+1:

jonathanKingston:
[...]

PR #337, opened 2016-04-14
Fix SRI issue #224

chbonser:
ember-paper runs autoprefixer in a postprocessTree hook, it therefore must run before ember-cli-sri or sri errors will ensue

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,6 +70,7 @@
   ],
   "ember-addon": {
     "configPath": "tests/dummy/config",
-    "demoURL": "https://miguelcobain.github.io/ember-paper"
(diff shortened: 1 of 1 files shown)


miguelcobain:
The build errors were related with jscs. They don't happen anymore.
Thanks a lot for this fix. We were having a hard time detecting the source of it.

blimmer:
Should we consider backporting this to 0.2.x? I just ran into this problem with 0.2.14 and a base ember-cli app.

v3ss0n:
I am still having this when i build ember-cordova with crosswalk.

blimmer:
@v3ss0n it's probably another addon that's a dependency of ember-cordova. this is fixed in paper.

PR #338, opened 2016-04-14
Update paper-card for 1.0

joukevandermaas:
- [x] Add support for headers & avatars
- [x] Create components for most common card 'parts' (actions, header, title, content)
- [x] Use contextual components (but not mandatory)
- [x] Add support for cards with various image sizes
- [ ] ~~Add support for 'dark' (colored) cards~~ (not in Angular Material 1.06)
- [x] Update changelog

PR #341, opened 2016-04-19
paper-card: implementation for v1.0

joukevandermaas:
Re-submitting #338.

- [x] Add support for headers & avatars
- [x] Create components for most common card 'parts' (actions, header, title, content)
- [x] Use contextual components (but not mandatory)
- [x] Add support for in-card media
- [x] Bring examples on par with [Angular Material](https://material.angularjs.org/1.0.6/demo/card)
- [x] Update changelog


miguelcobain:
I just added a couple of tests and updated some styling.
I also updated some namings. Should help if we ever need to support non contextual components.

PR #343, opened 2016-04-22
fix dep errors + version number

eriktrom:
3 minor changes:

- add ember-route-action to dependencies instead of devDependencies
- point to the projects nodeModulesPath - this allows consumers to see angular-material-source in their node_modules directory(and keeps the demo working as well)
- changed the version back to 0.2.12 as it was overwritten in 8b7dd871496e312afad5ef0d9e928a36307ebdfd

### DIFF
--- a/index.js
+++ b/index.js
@@ -103,8 +103,8 @@ module.exports = {
       'components/dialog/dialog-theme.scss'
     ];
 
-    var pathBase = this.project.addonPackages['ember-paper'].path;
-    var angularMaterialPath =  'node_modules/angular-material-source/src';
+    var pathBase = this.project.nodeModulesPath;
+    var angularMaterialPath =  'angular-material-source/src';
 
     var angularScssFiles = new Funnel(path.join(pathBase,angularMaterialPath), {
       files: scssFiles,


miguelcobain:
Why would we add ember-route-action to dependencies instead of devDependencies?

eriktrom:
I'm not completely sure TBH but without that the helper `transition-to` was not available

eriktrom:
That said - in order for npm to add a module to a consumers node_modules directory (which is not nested in npm 3) the dep has to be in dependencies not devDependencies (IIRC)

miguelcobain:
That is something we use for our demo app. It isn't something we should provide as part of ember-paper.
Users can use it as well, but not required.

DanChadwick:
Since we don't have tree-shaking yet, I think we want to avoid adding dependencies that not all users will need. Or am I missing something?

eriktrom:
Right on - its on the first page of the docs code that I copy pasted so perhaps we should just document it, no need to add it as a dep for everyone, I agree

I'll remove it. Is the rest fine?

DanChadwick:
I'm not sure about the version number. Version 0.2.12 would result from the branch for release 0.2.x. We could bump the version to something like 1.0.0-dev-0 maybe? Or even 0.9.0?

eriktrom:
Sure we can do whatever version you guys want. I just noticed there was a merge overwrite so I figured I'd better say something :)

DanChadwick:
After discussion on slack, please remove the change to the version number and we'll merge this.

eriktrom:
kool, ready

DanChadwick:
Needs a revert to the change to `ember-route-action`, right?

eriktrom:
yep - sorry misread your comment - should be good now

index.js
```diff
@@ -103,8 +103,8 @@ module.exports = {
       'components/dialog/dialog-theme.scss'
     ];
 
-    var pathBase = this.project.addonPackages['ember-paper'].path;
-    var angularMaterialPath =  'node_modules/angular-material-source/src';
+    var pathBase = this.project.nodeModulesPath;
```
miguelcobain: Do you mind explaining the reason for this change?
Does `this.project.nodeModulesPath` equal `this.project.addonPackages['ember-paper'].path + 'node_modules'`?

eriktrom: sure - prior to this change it was looking for angular-material-source, from my ember-cli terminal(for my project) at `node_modules/ember-paper/node_modules/angular-material-source/src`. This works fine for the dummy app b/c the lookup path then becomes `node_modules/angular-material-source/src`. To make this work when installing it in another app however, I changed it to be the `node_modules/angular-material-source` by using the current projects node_modules folder as the start of the lookup path instead of ember-paper's root dir only.

as an analogy - the net affect is like using process.cwd() instead of __dirname when making a node module - the lookup path is now relative the consumers root path and the source modules root too.

eriktrom: in npm 2.x this would not have been an issue (worth noting)

miguelcobain: In npm 3 there is a common `node_modules` folder. In npm 2.x each dep has its own `node_modules`.
With this change, we make sure it always points to the correct node_modules. Is this correct?

eriktrom: yep - sorry if i made that not so clear, my bad - that is the issue - npm 2.x was a recursive dep tree, in 3.x its flat

PR #346, opened 2016-04-23
Switch component must behave exactly like md angular example

xomaczar:
Fixed a couple of issues I've noticed:
1. When a pressed switch looses focus, and then re-gains it - the thumb must glow.
  1.a. Further changes to the pressed/unpressed state must keep the thumb glowing.
  1.b. When tapping the thumb ripple should be generate over the glow.
2. Clicking on the actual switch should not generate ripple effect on the thumb.
3. Space/Enter key on focused switch should behave like 1.a.
4. Fixed the drag --> was using the wrong elemen's innerWidth()

Compare this <-> ember-paper/master <-> angular md

### DIFF
--- a/addon/components/paper-switch.js
+++ b/addon/components/paper-switch.js
@@ -20,6 +20,8 @@ export default BaseFocusable.extend(RippleMixin, ProxiableMixin, ColorMixin, {
   classNameBindings: ['value:md-checked', 'dragging:md-dragging'],
   toggle: true,
 
+  constants: inject.service(),
+
   /* Ripple Overrides */
   rippleContainerSelector: '.md-thumb',
   center: true,
@@ -29,8 +31,6 @@ export default BaseFocusable.extend(RippleMixin, ProxiableMixin, ColorMixin, {
   value: false,
   disabled: false,
   dragging: false,
-
-  constants: inject.service(),
 
   thumbContainerStyle: computed('dragging', 'dragAmount', function() {
     if (!this.get('dragging')) {
@@ -74,20 +74,21 @@ export default BaseFocusable.extend(RippleMixin, ProxiableMixin, ColorMixin, {
   },
 
   _setupSwitch() {
-    this.set('switchWidth', this.$('.md-bar').width());
+    this.set('switchWidth', this.$('.md-thumb-container').innerWidth());
 
-    // Enable dragging the switch
     let switchContainer = this.$('.md-container').get(0);
-    let switchContainerHammer = new Hammer(switchContainer);
-    this._switchContainerHammer = switchContainerHammer;
-    switchContainerHammer.get('pan').set({ threshold: 1 });
-    switchContainerHammer.on('panstart', run.bind(this, this._dragStart));
-    switchContainerHammer.on('panmove', run.bind(this, this._drag));
-    switchContainerHammer.on('panend', run.bind(this, this._dragEnd));
+    let switchHammer = new Hammer(switchContainer);
+    this._switchContainerHammer = switchHammer;
 
-    let switchHammer = new Hammer(this.element);
-    this._switchHammer = switchHammer;
-    switchHammer.on('tap', run.bind(this, this._dragEnd));
+    // Enable dragging the switch container
+    switchHammer.get('pan').set({ threshold: 1 });
+    switchHammer.on('panstart', run.bind(this, this._dragStart))
+      .on('panmove', run.bind(this, this._drag))
+      .on('panend', run.bind(this, this._dragEnd));
+
+    // Enable tapping gesture on the switch
+    this._switchHammer = new Hammer(this.element);
+    this._switchHammer.on('tap', run.bind(this, this._dragEnd));
   },
 
   _teardownSwitch() {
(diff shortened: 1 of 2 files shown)


xomaczar:
The fix can also be done w/o modifying the base-focusable by overwriting focusIn hook for switch itself.

addon/components/base-focusable.js
```diff
@@ -10,6 +10,7 @@ export default Component.extend(EventsMixin, HasBlockMixin, {
   active: false,
   focused: false,
   hover: false,
+  focusable: true,
```
DanChadwick: It seems like maybe you could just set `focusOnlyOnKey` to `true` in `paper-switcth` to achieve the same effect. Or maybe I'm missing something? If you do need to keep `focusable`, maybe change the wordwrapping of the `if` in `focusIn()` to correspond to the semantics of the operator precedence as it is a touch misleading as wordwrapped.

addon/components/paper-switch.js
```diff
@@ -74,20 +77,19 @@ export default BaseFocusable.extend(RippleMixin, ProxiableMixin, ColorMixin, {
...
-    this._switchContainerHammer = switchContainerHammer;
-    switchContainerHammer.get('pan').set({ threshold: 1 });
-    switchContainerHammer.on('panstart', run.bind(this, this._dragStart));
-    switchContainerHammer.on('panmove', run.bind(this, this._drag));
-    switchContainerHammer.on('panend', run.bind(this, this._dragEnd));
-
-    let switchHammer = new Hammer(this.element);
-    this._switchHammer = switchHammer;
-    switchHammer.on('tap', run.bind(this, this._dragEnd));
+    this._switchContainerHammer = new Hammer(switchContainer);
```
DanChadwick: Are you sure that all js runtimes successfully optimize the local variable factoring? If not, it would seem better to leave the first `let`. Also, since `on` returns `this`, these may be chained in a fluent style a la jQuery.

 The second is a left-over from prior code cleanup and is of course good.

addon/components/base-focusable.js
xomaczar: Yes same thing can be archived with focusOnlyOnKey - however the intent is completely hidden away.

addon/components/paper-switch.js
xomaczar: I am not sure what you mean by successfully optimize - If you are implying the lookup speed of local variable  vs something on this -- I think it's negligible

DanChadwick: I totally agree its a tiny difference. Still, the original is smaller (after minificiation and probably after jit) and faster, it doesn't seem worth removing the optimization. But the best way out would probably be to just chain the `on`'s.

addon/components/base-focusable.js
DanChadwick: I haven't looked but I thought both checkboxes and switches could be focused and toggled with the keyboard. If focusOnlyOnKey needs a new name to cover all its uses, I think that would be a bit better

xomaczar:
Made changes we talked about, added more tests.

DanChadwick:
@xomaczar: Thanks!

addon/components/paper-switch.js
```diff
@@ -74,20 +74,21 @@ export default BaseFocusable.extend(RippleMixin, ProxiableMixin, ColorMixin, {
...
-
-    let switchHammer = new Hammer(this.element);
-    this._switchHammer = switchHammer;
-    switchHammer.on('tap', run.bind(this, this._dragEnd));
+    let switchHammer = new Hammer(switchContainer);
+    this._switchContainerHammer = switchHammer;
+
+    // Enable dragging the switch container
+    switchHammer.get('pan').set({ threshold: 1 });
+    switchHammer.on('panstart', run.bind(this, this._dragStart))
```
canufeel: What is the rationale for implementing Hammer Event handlers with Embers `run.bind`?

DanChadwick: @canufeel: Good question. That was the initial non-hammer implementation from [bb48eba]

canufeel: In my effort to rework paper-list-item i have a lot of troubles with debugging this events.
It seems that Hammer creates its own event from `click` and converts it to `tap`. But the initial `click` event is also emitted, is not handled anywhere and would bubble up.

xomaczar: Run.bind wraps handlers so that the handler has the correct context(i.e. this). If you look at the body of each handlers we're changing the state of the component through this.set invocation

canufeel: @xomaczar That is reasonable. I was just wondering if it might be run-loop related as the context binding might be also done with `.apply(this,args)` and for some reason is done via `run.bind` instead.

xomaczar: Any 3rd party javascript library with async callbacks need to be wrapped into an Ember.run. Doing apply(this, args) will most likely work in production as ember tries to automatically run stuff in a runloop for you - a feature known as autoruns. However, during testing autoruns are disabled - test async helpers (e.g andThen, etc) wait for runloop to finish its work after which you should assert. If you try to run callback handlers outside of the runloop, your asserts might occur before those handlers had a chance to run and your tests will fail. TLDR use runloops when integrating 3rd party line with Ember

xomaczar: @canufeel I checked the issue you are seeing with paper-list-item, specifically, the events bubbling cancellation issue. As far as I can tell, hammerjs v.2.x does not provide a clean way to cancel event propagation causing paper-switch to return to its original state when interacted with.
FYI, I also noticed multiple deprecation warnings in paper-list section of the dummy app - they're all seem to be related to the proxiable-mixin proxyRegistration

PR #347, opened 2016-04-26
don't import hammerjs and match media if being called from node

BenMorganIO:
While adding Ember Fastboot support to an application, I was receiving:

```javascript
$ ember fastboot --serve-assets
Built project successfully. Stored in "fastboot-dist".
Installing FastBoot npm dependencies
document is not defined
ReferenceError: document is not defined
```

This was because HammerJS and MatchMedia use the `document`, but NodeJS, which Fastboot uses, doesn't use the `document`, which leads the `ReferenceError`. Therefore, if Ember Fastboot is requiring `ember-paper`, lets skip HammerJS and MatchMedia.

### DIFF
--- a/index.js
+++ b/index.js
@@ -13,8 +13,11 @@ module.exports = {
   included: function(app) {
     this._super.included(app);
 
-    app.import(app.bowerDirectory + '/hammer.js/hammer.js');
-    app.import(app.bowerDirectory + '/matchMedia/matchMedia.js');
+    if (!process.env.EMBER_CLI_FASTBOOT) {
+      app.import(app.bowerDirectory + '/hammer.js/hammer.js')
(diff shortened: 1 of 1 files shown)


miguelcobain:
Thanks a lot. :+1:

BenMorganIO:
@miguelcobain if you could release a new bug fix version, that would super awesome :)

SaladFork:
For releasing a new version (pre 1.0.0) we will need to cherry-pick this to the `release-0-2` branch.

BenMorganIO:
@SaladFork thanks :)

SaladFork:
@BenMorganIO Just released `0.2.14` with your fix. Thanks for the contribution!

Issue #339, opened 2016-04-17
paper-button: documentation for links missing


borisrorsvort:
After inspection of the source I found it could be useful to add an example with buttons as links in the button section using the transition-to helper from ember route action.

ea:  `{{#paper-button action=(transition-to "index")}}Introduction{{/paper-button}}`

and give an example without the helper too

PR #354, opened 2016-04-29
add install instructions for ember-route-action

eriktrom:


eriktrom:
closes #339

tests/dummy/app/templates/index.hbs
```diff
@@ -45,6 +45,11 @@
   <p>You can find out more information on the CSP addon page <a href="https://github.com/rwjblue/ember-cli-content-security-policy#ember-cli-content-security-policy">here</a>.</p>
   <p>We also need to allow `data:` in `img-src` because of a <a href="https://github.com/miguelcobain/ember-paper/blob/master/app%2Fstyles%2Fpaper-button.scss">current hack</a> in paper-button.</p>
 
+  <p>
+    Finally, the examples contain a `transition-to` helper that is not installed by default. To use this helper run the following:
```
DanChadwick: I'm thinking this would be better added to the side-nav, even though it relates to installation. Only someone who both uses the side-nav and is copying the template would need this info. That way we don't overwhelm someone who is just getting started. Also, is there a reason why `ember install ember-route-action` wouldn't work? It also runs blueprints and I think just wraps `npm install --save-dev`, no? Last, if there is is a reason, I'd prefer the longer form of the option for clarity. Thanks for contributing!!

eriktrom: Howdy @DanChadwick - `ember install ember-route-action` is just fine. I'll move it to the sidenav when I'm back down there (today sometime)

DanChadwick: Thanks! I appreciate your help with this and the other issues.

eriktrom:
@DanChadwick - updated :)

Screenshot:
image

tests/dummy/app/templates/sidenav.hbs
```diff
@@ -25,6 +25,11 @@
       Defaults to <code>true</code>.
     </p>
 
+    <p>
+      <strong>Note:</strong> The following Sidenav example contain a <code>transition-to</code> helper that is not installed by default. To use this helper install the following:
+      {{#code-block language='bash'}}$ ember install ember-route-action{{/code-block}}
```
DanChadwick: Let's delete the `$ `. Sorry to nit-pick.

eriktrom: no worries, i also should add an 's' to 'contain' so all good :)

eriktrom:
done ;)

PR #359, opened 2016-05-02
Menu stuff

andybeesee:
Added some additional class bindings to paper-toolbar and paper-button

 * Paper-toolbar now has *tall* prop to make the menu...tall. 
 * Paper-button now has *fab* and *mini* props


DanChadwick:
Is there some reason we can't have separate PRs for each component? While they might be used together, I don't see any dependency between them.

After a brief chat on slack, I think we want to ensure that each boolean that controls a `classNameBinding` or similar is defined (presumably as `false`), rather than relying on undefined to cause the class to be omitted.

And last generally the project seems to like tests to ensure that the booleans do indeed render the classes.

andybeesee:
No problem, I'll close this, write tests and submit separate pull requests.

PR #362, opened 2016-05-04
Toolbar class bindings

andybeesee:
New *tall* class name binding for paper-toolbar, with test and example in documentation. Let me know what I can do to improve the test, I am still new to Ember and testing with it.

### DIFF
--- a/tests/integration/components/paper-toolbar-test.js
+++ b/tests/integration/components/paper-toolbar-test.js
@@ -1,0 +1,20 @@
+import { moduleForComponent, test } from 'ember-qunit';
+import hbs from 'htmlbars-inline-precompile';
+
+moduleForComponent('paper-toolbar', 'Integration | Component | paper toolbar', {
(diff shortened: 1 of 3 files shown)


tests/integration/components/paper-toolbar-test.js
```diff
@@ -0,0 +1,20 @@
+import { moduleForComponent, test } from 'ember-qunit';
+import hbs from 'htmlbars-inline-precompile';
+
+moduleForComponent('paper-toolbar', 'Integration | Component | paper toolbar', {
+  integration: true
+});
+
+test('should bind class names correctly', function(assert) {
```
miguelcobain: let's test for md-tall class only.
We'll probably add md-default-theme to another place in the code base soon.

miguelcobain:
There are still many things to do on toolbars, but there's no harm in accepting this.
Thanks.

PR #365, opened 2016-05-05
md-button class bindings

andybeesee:
Added fab and mini class name bindings, with tests and examples in dummy app. I appreciate any feedback I am still new to this.


miguelcobain:
Thanks! There is a slight improvement possible.
When a button is `mini` it is necessarily a `fab`. So, when we specify `mini=true` the button should assume that `fab=true` implicitly.

Suggestion: Make the default value of `fab` a computed property that has the value of `mini`, like
```js
fab: computed.reads('mini')
```

Also, add or edit a test that makes sure that `mini=true` adds both `md-fab` **and** `md-mini`.

andybeesee:
Good idea, I have made the updates.

miguelcobain:
Thanks again.

PR #366, opened 2016-05-05
Refactor of paper-item component. Bugs fixed, deprecations cleared.

### LINKED ISSUE #153: paper-item: activates disabled secondary item on click
```
          {{#paper-item class="md-with-secondary" disabled=true}}
              <p>{{model.name}}</p>
            {{paper-checkbox class="md-secondary" checked=model.isActive disabled=true}}

          {{/paper-item}}
```

If I click on the row, it still toggles the `model.isActive` state, even though the control is disabled. I can work around this by also setting `skipProxy=true` in my template markup above for the checkbox, but it would be nice if it would do this automatically based on disabled state.

### LINKED ISSUE #220: paper-item + paper-buttom + paper-icon: Tab key
Pressing the Tab key on a page that lists paper items with paper button and have paper icon or paper menu icon, will change the position of that icon.

canufeel:
- Moved paper-item `didInsertElement` hook code to separate function which runs right after proxied component is registred.
- Refactored original `didInsertElement` hook code for paper-item to avoid direct DOM manipulation.
- The proxied component is now registered in the loop next after it is initialized so the Ember double modification deprecations are avoided.
- Fixed a bug when secondary button in paper-list moved all the way left on click event or lost its `md-secondary` status.
- Fixed a bug that was causing the paper-switch click event to bubble when switch was used as a secondary item in paper-list.
- Added tests for paper-item
- Fixed paper-item secondary action classname.
- Fixes #220 
- Fixes #153


addon/components/paper-item.js
```diff
@@ -2,6 +2,19 @@ import Ember from 'ember';
 import RippleMixin from '../mixins/ripple-mixin';
 import ProxyMixin from 'ember-paper/mixins/proxy-mixin';
 
+const {
+  get,
+  set,
+  isEmpty,
+  computed,
+  run: { later }
```
miguelcobain: We're not destructuring more than one level. It helps to still have some context.
So, we should `run.later` instead of `later`.

addon/components/paper-item.js
```diff
@@ -2,6 +2,19 @@ import Ember from 'ember';
...
+} = Ember;
+
+const {
+  bool,
+  notEmpty
+} = computed;
```
miguelcobain: Same here. We use `computed.bool`.

addon/components/paper-item.js
```diff
@@ -10,100 +23,105 @@ export default Ember.Component.extend(RippleMixin, ProxyMixin, {
...
-  }),
+  hasPrimaryAction: notEmpty('onClick'),
 
-  secondaryItem: Ember.computed('proxiedComponents.[]', function() {
-    let proxiedComponents = this.get('proxiedComponents');
-    return proxiedComponents.find(function(component) {
-      return component.classNames.indexOf('md-secondary') !== -1;
-    });
+  hasSecondaryAction: computed('secondaryItem', 'onClick', {
+    get() {
```
miguelcobain: We're not using the `get`/`set` syntax. Please use the normal `function() {` for getter computeds.

tests/integration/components/paper-item-test.js
```diff
@@ -0,0 +1,182 @@
...
+  set(this, 'checkboxEnabled', false);
+  this.render(hbs`
+    {{#paper-list}}
+      {{#paper-item}}
+        <p>Checkbox 1</p>
+        {{paper-checkbox class="md-secondary" value=checkboxEnabled onChange=(action (mut checkboxEnabled))}}
+      {{/paper-item}}
+    {{/paper-list}}
+  `);
+  return wait().then(() => {
```
miguelcobain: Why do we need `wait()` here?

addon/components/paper-item.js
```diff
@@ -2,6 +2,19 @@ import Ember from 'ember';
...
+  computed,
+  run: { later }
+} = Ember;
+
+const {
+  bool,
+  notEmpty
+} = computed;
+
 export default Ember.Component.extend(RippleMixin, ProxyMixin, {
```
miguelcobain: Destructure `Component` as well.

tests/integration/components/paper-item-test.js
canufeel: As the proxies are initialized in the `Ember.run.next` without next they wont be initialized just yet.

miguelcobain: @canufeel I see.

Why can't we initialize the proxies on `init`?

miguelcobain: In ember-leaflet I'm doing it on `didInsertElement`: https://github.com/miguelcobain/ember-leaflet/blob/master/addon%2Fmixins%2Fchild.js

canufeel: @miguelcobain previous usage of `didInsertElement` triggered an Ember double modify deprecation. And it is distinct approach as it would only run once event if we would add a new component under paper-item at some point all the triggers wouldn't run simply. The `init` of paper-item is also not the best place as it would require registering an observer to run `setupProxiedComponent`. Then `setupProxiedComponent` would have to have `scheduleOnce` to avoid double invocation. In this situation the tests wouldn't run as `wait` is not so clever to wait for 2 serial async functions.

miguelcobain: What I'm saying is that we can register the proxy in `didInsertElement` without triggering the deprecation. Or maybe I'm missing something?
Why would the deprecation trigger in the first place?

canufeel: screen shot 2016-05-05 at 18 16 03

canufeel: @miguelcobain When paper-item initializes it runs the proxiable-mixin `init` hook. it had the `run.scheduleOnce('afterRender')` call to `registerProxy` method which would take care of proxies when they are already in DOM. the problem was that after the properties have setup in this run loop we would modify them again and the deprecation would trigger. And it also left some place for ambiguity as `didInsertElement` was running only once and did the DOM manipulation code for each proxy even though they might be not in the DOM just yet. So we had to provide the initial context for paper-item in a template, otherwise the buttons would not initialized properly ever as `didInsertElement` would run just once.

miguelcobain: I think it would be worth to get rid of the `run.scheduleOnce('afterRender')`. I'm pretty sure we can get around without it.
Can you point the source of the deprecation you posted?
Maybe this: https://github.com/miguelcobain/ember-paper/blob/master/addon%2Fmixins%2Fproxy-mixin.js#L9
plus this: https://github.com/miguelcobain/ember-paper/blob/master/addon%2Fcomponents%2Fpaper-item.js#L34 (and respective classNameBinding)

canufeel: @miguelcobain https://github.com/miguelcobain/ember-paper/blob/master/addon%2Fmixins%2Fproxy-mixin.js#L9 This is the source. As soon as we start adding objects to `proxiedComponents` array quite a few computed properties are being modified on the component.

miguelcobain: I see the problem now.
The parent only knows the children when rendering. So, there's no way to apply the correct classes before rendering.

Thanks.

canufeel:
Should i add change details to the changelog? Which tag should i add the changes under?

addon/mixins/proxiable-mixin.js
```diff
@@ -16,5 +16,18 @@ export default Ember.Mixin.create({
     }
   },
 
-  processProxy: null
+  processProxy: null,
+
+  // Paper item secondary container class
+  isSecondary: computed('class', {
+    get() {
```
miguelcobain: We're not using get/set computeds.

app/templates/components/paper-item.hbs
```diff
@@ -1,11 +1,3 @@
...
-    <div class="md-list-item-inner">
-      {{yield}}
-    </div>
-  {{/paper-button}}
-{{else}}
-  <div class="md-no-style md-list-item-inner" {{action "buttonAction" on="click"}}>
-    {{yield}}
-  </div>
-{{/if}}
+<div class="md-no-style md-list-item-inner" {{action "buttonAction" on="click"}}>
```
miguelcobain: Let's use the `onclick={{action "buttonAction"}}` notation instead.

tests/integration/components/paper-item-test.js
```diff
@@ -0,0 +1,182 @@
...
+
+const { set, get } = Ember;
+
+moduleForComponent('paper-item', 'Integration | Component | paper item', {
+  integration: true
+});
+
+test('single action checkboxes should react to checkbox clicks', function(assert) {
+  assert.expect(1);
+  set(this, 'checkboxEnabled', false);
```
miguelcobain: Can we use `this.set` and `this.get` to conform with the other tests?

app/templates/components/paper-item.hbs
canufeel: @miguelcobain seems that it is something different from just a notation. Changing this line breaks 2 tests. For some reason if we select the nested `paper-button` element with jquery and do a `click()` on it this `onClick` handler would run prior to the one of `paper-button` `click(event)` handler which should actually run first as event is directly targeted on it thus ignoring `paper-button` `bubbles` prop. Any ideas why this might happen?

canufeel: `<div class="md-no-style md-list-item-inner" {{action "buttonAction"}}>` also works fine. Should we stick to it?

miguelcobain: I didn't understand the problem. But, just to make sure, you tried `onclick=` and not `onClick=`, correct?

miguelcobain: If it doesn't work, leave it like that and we'll iterate later. This PR is more of a bugfix and not really an update to 1.0. 👍

canufeel: @miguelcobain Tried them both. Neither of them work. Do we need to update changelog? if yes, then under which version tag should i put the changes?

miguelcobain: The latest unreleased version. If not present, feel free to create an entry.

/CC @danchadwick

DanChadwick: I just pushed a changelog with a place to enter changes for 1.0.0-alpha.2.

Issue #207, opened 2015-11-06
paper-item: Integrate link-to similar behaviour into paper-item and similar components


brisc:
Now, although we can use a transition action to simulate the link-to helper on some components (eg: paper-item) it's not the same as a link-to. It's important that the component should generate a <a> tag with href, so that link can be opened in a new page. It's also important for accessibility and SEO.
A sidenav that uses a paper-list with paper-items should have real href links.
Maybe add a link-to attribute on paper-item and a special handling like how it was done for action?

avdmatamoros:
Agreed, it would be best to have something similar to a link and not need to rely on a custom action call for a simple redirect.

https://github.com/miguelcobain/ember-paper/issues/203#issuecomment-156232272

Subtletree:
Made a first run at a `paper-link` addon. Extends `link-to` for `paper-button` and 'paper-item`

https://github.com/Subtletree/ember-paper-link

miguelcobain:
Please see my comment at https://github.com/miguelcobain/ember-paper/issues/203#issuecomment-317150800

PR #372, opened 2016-05-12
paper-button: Support `<a>` tags with href and optional target attributes.

DanChadwick:
Also Improve dummy app page with API documentation and button. Use a-link style button to link from 1.0 dummy app to v0.2.

### DIFF
--- a/addon/components/paper-button.js
+++ b/addon/components/paper-button.js
@@ -7,14 +7,22 @@ import ColorMixin from 'ember-paper/mixins/color-mixin';
 const { computed } = Ember;
 
 export default BaseFocusable.extend(RippleMixin, ProxiableMixin, ColorMixin, {
-  attributeBindings: ['type'],
   type: 'button',
-  tagName: 'button',
+  tagName: computed('href', function() {
+    return this.get('href') ? 'a' : 'button';
+  }),
   classNames: ['paper-button', 'md-default-theme', 'md-button'],
   raised: false,
   iconButton: false,
   fab: computed.reads('mini'),  // circular button
   mini: false,
+  href: null,
+  target: null,
+  attributeBindings: [
+    'type',
+    'href',
+    'target'
+  ],
   classNameBindings: [
     'raised:md-raised',
     'iconButton:md-icon-button',
--- a/addon/components/paper-button.js
+++ b/addon/components/paper-button.js
@@ -7,15 +7,13 @@ import ColorMixin from 'ember-paper/mixins/color-mixin';
 const { computed } = Ember;
 
 export default BaseFocusable.extend(RippleMixin, ProxiableMixin, ColorMixin, {
-  type: 'button',
-  tagName: computed('href', function() {
-    return this.get('href') ? 'a' : 'button';
-  }),
+  tagName: 'button',
   classNames: ['paper-button', 'md-default-theme', 'md-button'],
   raised: false,
   iconButton: false,
   fab: computed.reads('mini'),  // circular button
   mini: false,
+  type: 'button',
   href: null,
   target: null,
   attributeBindings: [
@@ -35,6 +33,16 @@ export default BaseFocusable.extend(RippleMixin, ProxiableMixin, ColorMixin, {
   fitRipple: computed.readOnly('iconButton'),
   center: computed.readOnly('iconButton'),
   dimBackground: computed.not('iconButton'),
+
+  init() {
+    this._super(...arguments);
+    if (this.get('href')) {
+      this.setProperties({
+        tagName: 'a',
+        type: null
+      });
+    }
+  },
 
   click(event) {
     this.sendAction('onClick', event);
--- a/tests/dummy/app/templates/button.hbs
+++ b/tests/dummy/app/templates/button.hbs
@@ -8,7 +8,7 @@
 {{/paper-toolbar}}
 {{#paper-content class="md-padding demo-buttons"}}
 <div class="doc-content">
-  <div layout="row" class="flex">
+  <div class="layout-row">
     {{#paper-button onClick=(action "flatButton")}}Button with action{{/paper-button}}
     {{#paper-button noink=true primary=true}}Primary (noink){{/paper-button}}
     {{#paper-button disabled=true}}disabled{{/paper-button}}
--- a/tests/dummy/app/templates/button.hbs
+++ b/tests/dummy/app/templates/button.hbs
@@ -8,7 +8,7 @@
 {{/paper-toolbar}}
 {{#paper-content class="md-padding demo-buttons"}}
 <div class="doc-content">
-  <div layout="row" flex>
+  <div layout="row" class="flex">
     {{#paper-button onClick=(action "flatButton")}}Button with action{{/paper-button}}
     {{#paper-button noink=true primary=true}}Primary (noink){{/paper-button}}
     {{#paper-button disabled=true}}disabled{{/paper-button}}
--- a/tests/dummy/app/templates/button.hbs
+++ b/tests/dummy/app/templates/button.hbs
@@ -8,12 +8,13 @@
 {{/paper-toolbar}}
 {{#paper-content class="md-padding demo-buttons"}}
 <div class="doc-content">
-  <p>
+  <div layout="row" flex>
     {{#paper-button onClick=(action "flatButton")}}Button with action{{/paper-button}}
     {{#paper-button noink=true primary=true}}Primary (noink){{/paper-button}}
     {{#paper-button disabled=true}}disabled{{/paper-button}}
     {{#paper-button warn=true}}warn{{/paper-button}}
-  </p>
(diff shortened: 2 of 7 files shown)


addon/components/paper-button.js
```diff
@@ -7,14 +7,22 @@ import ColorMixin from 'ember-paper/mixins/color-mixin';
 const { computed } = Ember;
 
 export default BaseFocusable.extend(RippleMixin, ProxiableMixin, ColorMixin, {
-  attributeBindings: ['type'],
   type: 'button',
-  tagName: 'button',
+  tagName: computed('href', function() {
```
miguelcobain: Last time I checked using a computed `tagName` raises a deprecation. Hence the twiddle I mentioned in the issue.

tests/dummy/app/templates/index.hbs
```diff
@@ -11,11 +11,13 @@
   {{#paper-toolbar warn=true}}
     <div class="md-toolbar-tools">
       <h2>Pre-production Version</h2>
+      <span flex></span>
```
miguelcobain: Let's use `class="flex"` since we plan to leave flex attributes.

tests/dummy/app/templates/button.hbs
```diff
@@ -8,12 +8,13 @@
 {{/paper-toolbar}}
 {{#paper-content class="md-padding demo-buttons"}}
 <div class="doc-content">
-  <p>
+  <div layout="row" flex>
```
miguelcobain: Let's use the class equivalents here as well.

addon/components/paper-button.js
DanChadwick: Right you are. In wonder if there is a less ugly alternative.

addon/components/paper-button.js
```diff
@@ -7,14 +7,22 @@ import ColorMixin from 'ember-paper/mixins/color-mixin';
 const { computed } = Ember;
 
 export default BaseFocusable.extend(RippleMixin, ProxiableMixin, ColorMixin, {
-  attributeBindings: ['type'],
   type: 'button',
```
miguelcobain: I believe we need to make this a computed since we don't want to render `<a type="button"`.

tests/dummy/app/templates/button.hbs
DanChadwick: OK, but let's change the whole addon at once. I'm pretty sure I copied this from another component. Horizontal versus vertical updating, one might say.

addon/components/paper-button.js
miguelcobain: Check the twiddle. I think it is the right way to do it.

miguelcobain: Also, I think we should encapsulate the logic of testing href in a `isLinkButton` computed. That way, if we ever want to change the test we do it in just onde place.

addon/components/paper-button.js
DanChadwick: Or something. Not setting type would be cleaner, I think.

tests/dummy/app/templates/button.hbs
miguelcobain: I'm OK with that as well. Just thought that doing that for the new stuff we write will save a bit of work later.

addon/components/paper-button.js
miguelcobain: Like, not having the default type to button?

addon/components/paper-button.js
DanChadwick: Yeah, it's ugly, but maybe no alternative. When you reference a CP from init, are you guaranteed that all the parameters are already properly bound?

addon/components/paper-button.js
DanChadwick: Yes. Seems better to create the correct type attribute rather than create the wrong one and circumvent it.

addon/components/paper-button.js
miguelcobain: I don't think it is *that* ugly. Init is the place to initialize variables after all.

tagName will nota be bound. You can't change tagName after render. That's the reason it can't be a computed. Did I get your question right?

DanChadwick: We're on the same page. Technically you should be able to use a CP without a dependent key. I'll try that, but I bet it still raises a deprecation. I just hate that init function, but it's an aesthetic thing, not a functional or technical objection.

addon/components/paper-button.js
miguelcobain: You would always create the correct type.

```js
type: computed('isLinkButton', function() {
  return this.get('isLinkButton') ? null : 'button';
})
```

addon/components/paper-button.js
miguelcobain: Yes, it also raises a deprecation with a computed with no dependent keys.

miguelcobain: But you're right. A computed without a dependent key should not raise a deprecation, I think. We should PR ember core. :)

Yak shaving.

DanChadwick: And bike shedding. ;)

addon/components/paper-button.js
DanChadwick: No need for the dependent key. :)

DanChadwick:
Note: I did not create `isLinkButton` because this state is static, since the `tagName` can't change. Therefore I just set to `type` to `null` in `init`, along with the `tagName` to `a`. This is cleaner, simpler, smaller, and faster. If in the future we need a isLinkButton for some dynamic behavior, it too can be statically initialized in `init` to `true`.

tests/dummy/app/templates/button.hbs
```diff
@@ -8,12 +8,13 @@
 {{/paper-toolbar}}
 {{#paper-content class="md-padding demo-buttons"}}
 <div class="doc-content">
-  <p>
+  <div layout="row" class="flex">
```
miguelcobain: Well, this uses the layout attribute and a flex class. There is a `layout-row` *I think*. I've used it in the toolbar page.

tests/dummy/app/templates/button.hbs
```diff
@@ -63,4 +65,94 @@
...
+    </tr>
+    <tr>
+      <td><strong>target</strong></td>
+      <td>string</td>
+      <td>Sets the {{#code-inline}}&lt;a&gt;{{/code-inline}} link target attribute, such as {{#code-inline}}"_blank"{{/code-inline}}</td>
+    </tr>
+    <tr>
+      <td><strong>type</strong></td>
+      <td>string</td>
+      <td>Sets the html5 {{#code-inline}}type{{/code-inline}} attribute.</td>
```
miguelcobain: Maybe worth mentioning that it defaults to button.

tests/dummy/app/templates/button.hbs
```diff
@@ -63,4 +65,94 @@
...
+    </tr>
+    <tr>
+      <td><strong>warn</strong></td>
+      <td>boolean</td>
+      <td>Display in the theme's warn color</td>
```
miguelcobain: accent, primary and warn are part of a mixin and are present in many other components. We may need to find a better way to show common attributes to avoid repeating ourselves over and over, but for now listing them here seems the best thing to do.

miguelcobain:
Looks good to me. You can squash and merge (I'm on mobile). Thanks for this. :+1:

openhouse:
So in 1.0.0-alpha.1 I can use 
```{{#paper-button href="/settings"}}Settings{{/paper-button}}```
to get an HTML-friendly anchor tag
```<a href="/settings">Settings</a>```

But the raw HTML causes a hard reload which in the long run isn't ideal.  It's a little slower, wouldn't allow for animations and resets my route history if I'm keeping it.

I can keep the anchor tag for SEO and still use Ember's router to transition with the transition-to addon and `bubbles=false`.  But this doesn't let me command click to open the link in a new tab.
```{{#paper-button onClick=(transition-to "settings") href="/settings" bubbles=false}}Settings{{/paper-button}}```

This is starting to get a little boggling compared to the `{{link-to}}` helper. Am I going about it wrong?

I look forward to a standard API for this that is consistent with with `paper-item` as well.  It seems like mapping to `{{link-to}}` arguments would be the most Ember-intuitive way.

So grateful for this ambitious project and all your work.  Thank you!

miguelcobain:
@openhouse the idea was to have a link button that could link to external websites. I've never thought about different routes, but it should work. Hint: this addon renders url strings for a specific route: https://github.com/intercom/ember-href-to. Ideal to use with paper-button's new `href=` attribute.

openhouse:
Cool!

With a minor change to https://github.com/intercom/ember-href-to it makes seamless transitions with cmd+click opening in a new tab.

To enable this you have to have the `.href-to` class

`{{#paper-button href=(href-to "settings") class="href-to"}}Settings{{/paper-button}}`

and this pull request.
https://github.com/intercom/ember-href-to/pull/43

What this will be really great for is paper-items in the sidenav. #207

Thanks @miguelcobain for pointing me in the right direction!

PR #373, opened 2016-05-14
paper-api: A component for the dummy app to display API option tables.

DanChadwick:
The goal is to make it easier to document options and to allow commonly-used options, such as those relating to mixins, to be reused in a DRY manner. As a demonstration, see the dummy app's paper-button page.


miguelcobain:
One thing that I don't like is that suddenly we need to write markdown and text in controllers.

miguelcobain:
Do you really think this leads to a much better maintainability?
I think we're not encouraging ember best practices with this.

At first I thought that components like `api-table`, `api-line`, etc, should be enough, just to encapsulate the styles.
Then we could also have a `api/color-mixin` component to DRY the color mixin API. Same with the Ripple mixin, etc.

And to throw another thing in the discussion: what if we generate API docs from source comments. Example output: http://offirgolan.github.io/ember-light-table/docs/modules/Components.html (using https://github.com/cibernox/ember-cli-yuidoc)

DanChadwick:
I'm not sure how this isn't best practices. We could put it in the route and pass it in the model, but then this mucks with those components that use the model for generating data for the examples.

Yes, clearly I think this is a substantial step forward in maintainability. It was motivated purely by the unpleasant task of documenting `paper-button`. I thought it was worth investing a day of my time so that I or others would have to go through this for the other components, most of which don't have an API section, or have an incomplete one.

I'm not much in favor of yuidoc for this because I think having the examples, templates, and API together is very helpful. I think our current dummy app is much more helpful and engaging than `ember-light-table`, for example. (Nothing against the module itself, of course!). I find it very hard to navigate codebases which user auto-generated documentation because the code gets lost in the documentation annotations.

But, this isn't my module. Feel free to close if you don't think this is the way forward for `ember-paper`.

miguelcobain:
I see. You're considering this the routes data as well. Makes more sense from that perspective.

PR #374, opened 2016-05-15
Improve documentation site

joukevandermaas:
The goal of this PR is to

1. make it easier to create good looking/usable docs;
2. make it easier to use the docs site as an ember-paper user;
3. make the site look nicer.

Todo:
- [x] Give demos a parent route
- [x] Make the toolbar consistent across all pages
- [x] Improve the sidebar
- [x] Apply a consistent style to all demo pages
  - [ ] ~~Use foldable code snippets & cards to contain the samples~~ (will do in follow-up PR when tabs are implemented)
  - [x] Update logo & colors
  - [x] Make sure container elements, headings, etc. are the same
- [x] Extract common template/styles into components
  - [x] Use code snippets addon instead of manual copy/paste & escaping
  - [ ] ~~Component for API docs~~ (done in #373?)
- [ ] ~~Make sure the demo content is useful/complete~~ (will do in follow-up PR if needed)

Current status: http://awesan.com/ember-paper


borisrorsvort:
I'd use "components" instead of "demos" though in the sidebar

DanChadwick:
- There is something disconcerting about the tall toolbar on the left and the regular one one the right. I think it might be better without "tall". 
- I think it would be better to have have the demo/components section be expandable, or at least not by default collapsed. You will almost always want to look inside this section. I'm not sure that introducing the level of hierarchy makes it easier to understand or navigate, although if we in the future add other section on theming, colors, etc, maybe it would be warranted then?

PR #379, opened 2016-05-20
Fixed lastIsInvalid -- now being set properly

shoxter:
onInvalid was firing way more than it should have been.


DanChadwick:
Thanks!

PR #384, opened 2016-05-24
sidenav update

miguelcobain:
to let other people review this


addon/components/paper-sidenav-toggle.js
```diff
@@ -1,27 +1,26 @@
...
-    }
-  },
+  name: 'default',
 
-  didInsertElement() {
+  init() {
     this._super(...arguments);
     if (this.get('navContainer')) {
-      let lockedOpen = this.get('navContainer').get('sideBar').get('locked-open');
+      let lockedOpen = this.get('navContainer').get('sideBar').get('lockedOpen');
```
DanChadwick: Would there be a use case for lockedOpen being dynamic (i.e. changing after component creation)? I don't have one, but it would be easy to move this to didReceiveAttrs().

miguelcobain: this is on the toggle component. The toggle component is reaching the sidenav container for the `lockedOpen`. lockedOpen is never passed in to sidenav-toggle, so I don't see how we can use didReceiveAttrs?

addon/components/paper-sidenav.js
```diff
@@ -1,54 +1,66 @@
...
-      this.get('navContainer').set('sideBar', this);
-    }
+    this._super(...arguments);
 
-    this.matchMedia();
-    this.set('__resizeWindow', function() {
-      _self.matchMedia();
-    });
-  }),
+    /*if (this.get('navContainer')) {
```
DanChadwick: Leftovers?

addon/services/paper-sidenav.js
```diff
@@ -0,0 +1,58 @@
...
+
+  willDestroyElement() {
+    this._super(...arguments);
+    $(window).off('resize.sidenav-service', run.bind(this, 'updateLockedOpen'));
+  },
+
+  updateLockedOpen() {
+    let sidenavs = this.get('sidenavs');
+    Object.keys(sidenavs)
+    .map((k) => sidenavs.get(k))
```
DanChadwick: I've usually seen fluent code indented. Just an option.

addon/components/paper-sidenav.js
miguelcobain: true. I still need to make up my mind on the sidenav service. :/

addon/components/paper-sidenav-toggle.js
DanChadwick: Ah, right. Difficult to bind attributes from components that are parent/child. No matter as the use case is sketchy anyhow.

miguelcobain: still need a better way to do this with the new service. shouldn't be difficult.

tests/dummy/app/templates/components/page-toolbar.hbs
```diff
@@ -1,7 +1,7 @@
 {{#paper-toolbar class="md-whiteframe-1dp page-main-toolbar" as |toolbar|}}
   {{#toolbar.tools}}
-    {{#paper-sidenav-toggle as |toggle|}}
-      {{#paper-button onClick=(action "toggleMenu" target=toggle) iconButton=true}}
+    {{#paper-sidenav-toggle class="hide-gt-sm"}}
+      {{#paper-button onClick=null iconButton=true}}
```
joukevandermaas: Is this idiomatic? Maybe it should still be:

```hbs
{{#paper-sidenav-toggle as |toggle|}}
  {{#paper-button onClick=toggle iconButton=true}}
    {{paper-icon "menu"}}
  {{/paper-button}}
{{/paper-sidenav-toggle}}
```

That way you can also use a link or something instead and it's explicit what the purpose of the toggle+button combo is?

miguelcobain: Yeah, maybe that is more flexible as well.

PR #386, opened 2016-05-27
Fixed paper-select issue

ibarrick:
Problem was that the md-active class (which mandated that the display be block) was being removed before the transition started, causing the 'display:none' from paper-select-container to interrupt the transition. The reason it worked in dummy app was that dummy app forces all ember-views to have display flex.

Looked at AM code and it appears the correct solution was to remove md-active after the transition ended.

### DIFF
--- a/addon/components/paper-menu-container-abstract.js
+++ b/addon/components/paper-menu-container-abstract.js
@@ -55,8 +55,11 @@ export default Ember.Component.extend({
   hideWrapper() {
     let _self = this;
     return new Ember.RSVP.Promise(function(resolve/*, reject*/) {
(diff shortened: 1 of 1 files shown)


DanChadwick:
What is the symptom and/or steps to reproduce symptom? I'm not sure what's wrong so I'm not sure if this fixes it. :)

ibarrick:
Create a blank Ember app and add a select component. Open the drop down, then click off to dismiss it. The select will no longer be click able.

DanChadwick:
Hmmmm. Still can't reproduce the problem. Maybe @miguelcobain can help here.

miguelcobain:
this is an immediate fix. when this component is updated, this may not be needed at all (not sure, need to dig in).

Issue #378, opened 2016-05-19
Menu Bug


shoxter:
There seems to be an issue with menu buttons where the first menu item always maintains its focused CSS state.

This is different from the angular.js behavior where it changes the moment the cursor moves to another item or off of the menu altogether, so I'm assuming it's a bug.

PR #390, opened 2016-05-31
fix for issue #378

ibarrick:
problem with #378 was that mouseEnter events on menu items should trigger a focus if there is a non-disabled md-button inside the menu item. Corresponding AM code can be seen here:
https://github.com/angular/material/blob/master/src/components/menu/js/menuController.js
in ```this.handleMenuItemHover()```.

### DIFF
--- a/addon/components/paper-menu-item.js
+++ b/addon/components/paper-menu-item.js
@@ -12,7 +12,7 @@ export default Component.extend({
       this.sendAction('onClick', event);
     }
   },
-  mouseEnter: function() {
+  mouseEnter() {
   	this.$('.md-button:not([disabled])').focus();
   }
 
--- a/addon/components/paper-menu-item.js
+++ b/addon/components/paper-menu-item.js
@@ -11,6 +11,9 @@ export default Component.extend({
       this.nearestOfType(PaperMenuAbstract).send('toggleMenu');
       this.sendAction('onClick', event);
     }
+  },
+  mouseEnter: function() {
+  	this.$('.md-button:not([disabled])').focus();
   }
 
 });
--- a/addon/components/paper-menu-item.js
+++ b/addon/components/paper-menu-item.js
@@ -13,7 +13,9 @@ export default Component.extend({
     }
   },
   mouseEnter() {
-  	this.$('.md-button:not([disabled])').focus();
+    if (!this.get('disabled')) {
+      this.$('button').focus();
+    }
   }
 
 });


addon/components/paper-menu-item.js
```diff
@@ -11,6 +11,9 @@ export default Component.extend({
       this.nearestOfType(PaperMenuAbstract).send('toggleMenu');
       this.sendAction('onClick', event);
     }
+  },
+  mouseEnter: function() {
```
miguelcobain: let's use es6 new object syntax `mouseEnter() {`

miguelcobain: otherwise tests shouldn't pass

ibarrick: Thanks! should be good now.

miguelcobain:
I see that AM does some stuff on mouseLeave as well: https://github.com/angular/material/blob/master/src/components/menu/js/menuController.js#L103-L108

Don't we need to do that as well? If not, why?

ibarrick:
To be honest, I wasn't sure how it related to our implementation. It looks like AM supports nested submenus that open a short while after mouseEnter. It isn't shown on the demo and I've seen no indication that we support it.

ibarrick:
Correction: it is in the AM demo under "Menu Bar".

miguelcobain:
@ibarrick i see the problem now

I think we should change this to:

```js
mouseEnter() {
  if (!this.get('disabled')) {
    this.$('button').focus();
  }
}
```

Your code is correct, but I think this is (1) more clear (2) uses less "jquery css selector magic" (the less jquery, the better) (3) and it makes future `disabled` implementation changes more easy to deal with (the `disabled` property in paper-item is the thing that matters here).

Thoughts?

ibarrick:
@miguelcobain I'll change it to that now, and try to fix that JSHint issue.

addon/components/paper-menu-item.js
```diff
@@ -11,6 +11,9 @@ export default Component.extend({
       this.nearestOfType(PaperMenuAbstract).send('toggleMenu');
       this.sendAction('onClick', event);
     }
+  },
+  mouseEnter() {
+  	this.$('.md-button:not([disabled])').focus();
```
miguelcobain: A test is failing because the indentation is incorrect here.
You can run `ember test` to make sure the tests pass before commits.

miguelcobain:
It is a jscs issue, not a jshint issue. :)

Great. Thanks.

ibarrick:
Good to go, I'll make sure I test first next time!

PR #393, opened 2016-06-11
paper-content: arrow keys scroll main content, add tabindex attribute binding

openhouse:
Presently it's not possible to scroll http://miguelcobain.github.io/ember-paper/release-1/#/typography from the keyboard.

This pull request adds tabindex binding to the paper-content component, so that a tabindex may be set on the main paper-content for a page, allowing keyboard scrolling.

```
{{#paper-content class="md-padding" tabindex="0"}}
    long content which you can now click, tab into or otherwise focus to scroll
    from the keyboard
{{/paper-content}}
```

### DIFF
--- a/tests/dummy/app/templates/typography.hbs
+++ b/tests/dummy/app/templates/typography.hbs
@@ -1,6 +1,6 @@
 {{page-toolbar pageTitle="Typography" isDemo=false}}
 
-{{#paper-content class="md-padding"}}
+{{#paper-content class="md-padding" tabindex="0"}}
 <div class="doc-content">
   <h3>Headings</h3>
   <div class="preview-block">
--- a/addon/components/paper-content.js
+++ b/addon/components/paper-content.js
@@ -4,6 +4,6 @@ import FlexMixin from '../mixins/flex-mixin';
 export default Ember.Component.extend(FlexMixin, {
   tagName: 'md-content',
   classNames: ['md-default-theme'],
-  attributeBindings: ['layout-padding', 'scroll-y:md-scroll-y'],
+  attributeBindings: ['layout-padding', 'scroll-y:md-scroll-y', 'tabindex'],
   classNameBindings: ['padding:md-padding']
 });
--- a/app/services/constants.js
+++ b/app/services/constants.js
@@ -55,7 +55,7 @@ export default Ember.Service.extend({
     'xl': '(min-width: 1920px)',
     'print': 'print'
   },
-  
+
   MEDIA_PRIORITY: [
     'xl',
     'gt-lg',
--- a/tests/dummy/app/templates/components/doc-content.hbs
+++ b/tests/dummy/app/templates/components/doc-content.hbs
@@ -1,4 +1,4 @@
-{{#paper-content class="md-padding"}}
+{{#paper-content class="md-padding" tabindex="0"}}
   <div class="doc-content {{class}}">
     {{yield}}
   </div>
--- a/tests/dummy/app/templates/index.hbs
+++ b/tests/dummy/app/templates/index.hbs
@@ -1,6 +1,6 @@
 {{page-toolbar pageTitle="Introduction" isDemo=false}}
 
-{{#paper-content flex=true scroll-y=true layout-padding=true}}
+{{#paper-content flex=true scroll-y=true layout-padding=true tabindex="1"}}
 <div class="doc-content">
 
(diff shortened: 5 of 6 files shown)


tests/dummy/app/templates/typography.hbs
```diff
@@ -1,6 +1,6 @@
 {{page-toolbar pageTitle="Typography" isDemo=false}}
 
-{{#paper-content class="md-padding"}}
+{{#paper-content class="md-padding" tabindex="0"}}
```
DanChadwick: This looks good to me. Maybe @miguelcobain can take a peek. I just added `theme.hbs`, so that would need a `tabindex` too.

tests/dummy/app/templates/theme.hbs
```diff
@@ -1,6 +1,6 @@
 {{page-toolbar pageTitle="Theme & Color" isDemo=false}}
 
-{{#paper-content class="md-padding"}}
+{{#paper-content class="md-padding" tabindex="0"}}
```
openhouse: added tabindex to theme.hbs

mike1o1:
The issue also exists on the Angular Material project, and they have some discussion here: https://github.com/angular/material/issues/2961 

They seem to use `tabindex=-1` for their paper content, and they explain some accessibility issues for doing so. How does this fix work with regards to accessibility?

DanChadwick:
Does this help?
https://developer.mozilla.org/en-US/docs/Web/Accessibility/Keyboard-navigable_JavaScript_widgets

openhouse:
Quite exciting, my first pull request to Ember-Paper!  I'm digging it and appreciate all the great work you have done.

The component change is minimal, it doesn't prescribe a specific `tabindex` or set it automatically for `paper-content`.  It just exposes the component's `tabindex` attribute to be set in templates by the user.

That said, the dummy app itself is a great example.  So the dummy app is a question of best practices.

Thanks @mike1o1!  Angular Material docs also have this problem.  https://material.angularjs.org/1.0.6/  They require the mouse to scroll which is an accessibility bug and a minor headache to keyboard centric folks.  

If we want to leave the dummy app as is I could remove the dummy app changes, while leaving the `tabindex` attribute available on the `paper-content` component.  This would allow users to enable scrolling in their apps without overriding components locally.

Otherwise if we want to enable keyboard scrolling on the main `paper-content` for each page in the dummy app, should we set `tabindex` to `-1` or `0`? Both values allow the element to be focused by clicking or javascript. The `0` value allows `tab` to focus the element, while `-1` does not. Thanks @DanChadwick!

In my project I might experiment with automatically focusing the main `paper-content` after transition to make it easiest for people to start scrolling.  I may also try restructuring so primary content may be scrolled with body.

DanChadwick:
Hmmm. Does paper-content needs its open page in the dummy app? If not, how does one discover the available component attributes and other API? That doesn't necessarily have to be part of the PR tho.

openhouse:
Hi @mike1o1, 
Looking more at angular/material#2961.  Angular Material's docs have a "skip to content" accessibility feature.  This allows people who can't use the mouse to tab directly to a focusable element in the main content of the page without having to tab arduously through every navigation link first.  Part of their discussion of `tabindex="-1"` is designing to maintain this feature.

am-skip-to-content

In Angular Material docs you can `tab` to focus on an element in the main content.  This enables keyboard scrolling on the main content.  They have a bug where clicking the main content doesn't enable keyboard scrolling.

In Ember Paper a page with no focusable elements like http://miguelcobain.github.io/ember-paper/release-1/#/typography is inaccessable below the fold from the keyboard.  Two ways to fix this:
- ensure that every page has a focusable element in the main content such as a link.  AM docs have a footer with license links on every page.   This allows `tab` to unlock keyboard scrolling.
- make `paper-content` focusable with `tabindex` as in this pull request.  This makes the content keyboard scrollable by `tab` and by `click`.
  - `tabindex="-1"` enables `click` on `paper-content` containing focusable elements
  - `tabindex="0"` enables `tab` and `click` on `paper-content` containing no focusable elements

Both are probably good steps.

Some applications may wish the main content to be keyboard scrollable with no prior action, as with most web pages.  They could:
- focus the main `paper-content` on `didInsertElement`
- use a different layout that scrolls with `body`

miguelcobain:
@openhouse Why can't we `tabindex=0` in all page main contents?

PR #394, opened 2016-06-14
replace startWith with substr

mellatone:
This method used in the dummy app is not natively supported by IE11 and breaks Ember.

### DIFF
--- a/tests/dummy/app/controllers/application.js
+++ b/tests/dummy/app/controllers/application.js
@@ -14,7 +14,7 @@ export default Controller.extend({
   },
 
   expandedItem: computed('currentRouteName', function() {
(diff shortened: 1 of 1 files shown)


tests/dummy/app/controllers/application.js
```diff
@@ -14,7 +14,7 @@ export default Controller.extend({
   },
 
   expandedItem: computed('currentRouteName', function() {
-    if (this.get('currentRouteName').startsWith('layout')) {
+    if (this.get('currentRouteName').substr(0, 6) === "layout") {
```
DanChadwick: Right you are, but could you please use single quotes in .js files. Double quotes are for html-ish code.

PR #344, opened 2016-04-22
Feature/new new chips

mansona:
Second update of the Chips functionality. Continuation of https://github.com/miguelcobain/ember-paper/pull/267

PR #395, opened 2016-06-16
paper menu, paper select and paper autocomplete

miguelcobain:


ibarrick:
@miguelcobain I think it was a mistake to remove the `didOpen` from paper-menu-content. It's passed into `didInsert` of the wormhole. May be a similar story with the select component.

miguelcobain:
I was going nearly crazy with transitionend bugs across browsers, but then I found this PR. Looks like we're not alone. I'll put this here for reference:
https://github.com/facebook/react/pull/4561

We'll need to make our transitionEnd "waits" more robust.

brendanoh:
The changes to Paper-Menu look fantastic. Perfect animation. Please take any thoughts below as trying to be helpful.

I see some weird issues with timing on paper-select in the dummy app. I am working through all three (PS, PM and PA) in my app and will let you know if anything has any issues from my POV.

Amazing PR @miguelcobain this looks better than I expected.

brendanoh:
There are some strange issues with paper-select. Sometimes it closes as soon as you release the left click. Sometimes it stays open forever even after clicking on a selection. Sometimes you need to hold the left click down while you move the pointer to be able to make a selection. Happens in my app and in the demo app. 

paper-select

Similar issue with paper-menu staying open after the onClick fires. Happens in my app and in the demo app. 

paper-menu

I haven't dug into the why this is happening yet but I thought I would tell you what I am seeing. @ibarrick @miguelcobain if this isn't helpful just let me know.

ibarrick:
I'm getting the same behavior after commit 825d6b88. I think Some of the changes that were made were breaking changes.

ibarrick:
@brendanoh I was able to get it behaving normally again by changing the version of ember-basic-dropdown to 0.12.0-beta.10. Maybe we should depend on a specific version of e-b-d until it settles down.

brendanoh:
@ibarrick I tried changing ember-basic-dropdown to 0.12.0-beta.10 in my app but got the same results I did before. Is that all you changed?

ibarrick:
@brendanoh Also switched ember-wormhole to 0.3.6

brendanoh:
But you didnt change ember-power-select?
├─┬ ember-power-select@1.0.0-beta.5
│ └─┬ ember-basic-dropdown@0.12.0
│   └── ember-wormhole@0.4.0

brendanoh:
@ibarrick I changed both wormhole and basic dropdown and no change in behavior in either my app or the dummy app.

ibarrick:
@brendanoh My bad I did move down to e-p-s 1.0.0-beta.4. I made a lot of tweaks to get it to work alright. I'd love to have ember-paper depending on a stable version of these :/

brendanoh:
@ibarrick I changed that as well to no avail in both but thanks a lot for your help. 

At this rate your going to paste in your package.json before it works for me ;-)

miguelcobain:
This branch is under a lot of changes. Me and cibernox did some chatting about this integration.
Yesterday was my birthday, so I couldn't advance this very much yet. 😅

brendanoh:
@miguelcobain I understand completely. I converted an app that I had started in the previous Ember Paper version and I got to the point where I needed Autocomplete so I took the plunge. Just throwing what I find here in a hope it will help.

app/templates/components/paper-select-content.hbs
```diff
@@ -0,0 +1,20 @@
...
+      {{/if}}
+    --}}
+    {{paper-backdrop class="md-select-backdrop md-click-catcher"}}
+    <div
+      id={{dropdownId}}
+      class="ember-basic-dropdown-content {{class}} {{if renderInPlace 'ember-basic-dropdown-content--in-place'}} {{if hPosition (concat 'ember-basic-dropdown-content--' hPosition)}} {{if vPosition (concat 'ember-basic-dropdown-content--' vPosition)}}
+      md-select-menu-container md-clickable {{if isActive "md-active"}} {{unless isActive "md-leave"}}"
+      style={{style}}
+      dir={{dir}}>
+      {{#paper-select-menu-inner width=width dense=dense dropdown=dropdown searchEnabled as |innerContentHash|}}
```
brendanoh: Should this be `searchEnabled=searchEnabled`

xomaczar:
Yes this should be fixed in #481

miguelcobain:
@xomaczar @ibarrick i've merged master and your pr into this branch. Let's work from here.

**Completed:**
- [x]  You should be able to find `Portugal` now - the dummy app was not working properly.
- [x]  Added an animated reset/close button (had to create another component and use the transition mixin => in order to disabled tab focus (`tabindex="-1"`)
- [x] fix jscs errors

**Remaining:**
- [ ]  "no results" message is shown in a large dropdown. The containing element must set the height explicitly based on # of items shown - the `dropdown.content` has an absolute positioning.
- [ ] ensure "create" functionality works correctly. I played with AM "create" functionality and I couldn't figure out how "it's supposed to work".
- [ ] `autocomplete with label and query` pushes `paper-progress-linear` down where it's not visible due to the rendered `dropdown`. We need to figure out how make `ember-basic-dropdown.trigger` wrap the input and not the whole paper-input element.
- [ ] fix virtual-repeat deprecations @ibarrick 
- class based flex vs. attribute. Refactoring `flex-mixin` will solve this for all ep components. (I'd prefer a separate PR for this)
- integrate `virtual-repeat` with autocomplete (Maybe a separate PR)

ibarrick:
Are the deprecations in autocomplete already fixed?

miguelcobain:
They aren't, @ibarrick

A ter, 6/09/2016, 23:50, ibarrick <notifications@github.com> escreveu:

ibarrick:
I don't see any in the deprecation tab :/ and @xomaczar pushed a commit 5 days ago which had the commit message: "fixed paper-autocomplete deprecations"

miguelcobain:
I think it was a "fix". Because he basically removed virtual-repeat from
autocomplete to fix them? Correct me if I'm wrong.

A ter, 6/09/2016, 23:56, ibarrick <notifications@github.com> escreveu:

ibarrick:
You're right... Was that just to get rid of the deprecation warnings? seems pretty "scorched earch".

miguelcobain:
This isn't on master yet. It's wip. :)

A qua, 7/09/2016, 00:00, ibarrick <notifications@github.com> escreveu:

miguelcobain:
I assumed that starting with no deprecations would be benefitial to
understand exactly what is causing them. Sorry if too much work was undone.
We can always revert that commit or something.

A qua, 7/09/2016, 00:03, Miguel Andrade <jmandrade0@gmail.com> escreveu:

xomaczar:
The original virtual-repeat is still in the template, don't remember which one, autocomplete-something, I simply commented it out. You should be able to uncomment and see all the deprecation warnings.

xomaczar:
Sorry for the confusing git comment - I tried to emphasize that I did not actually fix it just commented it out. My bad.

xomaczar:
Also, there are a lot of components at play and deprecation traceback was not really helpful in identifying which component was causing the issue. In those cases I usually start by removing components to get a better idea where deprecations are coming from.

xomaczar:
@miguelcobain can you post an update version of the remaining tasks, so we are all on the same page

ibarrick:
I've figured out what causes each of the deprecations and fixed a couple, but the rest are probably going to involve considerable effort. I'll hopefully get through all this tomorrow.

ibarrick:
Did anyone take care of the label not showing up on `paper-select` after something has been selected?

xomaczar:
I don't think so.

pauln:
Is there anything I can do to help get this ready for integration?  I'm working on a project which needs autocomplete, and while I can use the paper-menu branch for dev at the moment, it'd be great to see this stabilise and land - and I'm happy to help.

I'm also wondering whether there are any current plans to allow for multiple selections in paper-autocomplete, using chips (PR #344) for selected items and/or allowing them to be customised, as I'm currently working on a UI which would benefit greatly from such a thing.

xomaczar:
As far as I know AM does not allow multiple selections in its autocomplete impl.

ibarrick:
@pauln It will be a part of chips when they land. That's not a priority for 1.0 though. You can always modify the existing PR for chips to be compatible with the new API for autocomplete, that way when 1.0 is ready we can eventually merge in chips.

ibarrick:
@miguelcobain When using paper-select with the async-search feature from e-p-s, the dropdown starts to jump around as the MutationObserver in e-b-d triggeres repositioning. Can I submit a PR to remove that?

miguelcobain:
Why doesn't that also happen in vanilla eps?

ibarrick:
We had to change the positioning logic to match MD spec. It probably has something to do with that but I'm not sure.

addon/components/paper-autocomplete.js
```diff
@@ -2,326 +2,83 @@
...
-        case this.get('constants').KEYCODE.ESCAPE:
-          this.set('searchText', '');
-          this.set('selectedIndex', this.get('defaultIndex'));
-          this.set('model', null);
-          this.set('hidden', this.get('shouldHide'));
-          break;
-        default:
-          break;
+    scrollTo(option /*, e */) {
+      if (!self.document || !option) {
```
brendanoh: self is undefined

addon/components/paper-autocomplete.js
```diff
@@ -2,326 +2,83 @@
...
+        return;
       }
-    },
+      let publicAPI = this.get('publicAPI');
+      let optionsList = self.document.getElementById(`ember-power-select-options-${publicAPI.uniqueId}`);
```
xomaczar: The global self is a shortcut to window.self. They are both the very same object.

xomaczar: See https://developer.mozilla.org/en-US/docs/Web/API/Window/self

brendanoh: Understood. When I pointed my existing project to the latest commit f780b54 I got those 2 lines as lint errors.

miguelcobain:
@xomaczar @ibarrick what is exactly missing?

xomaczar:
@miguelcobain autocomplete is ready to be merged. The scroll mask and searchTextChanged I'm working on can wait. Let's merge it finally - keep the warning sign in demo app - some autocomplete API might change slightly

miguelcobain:
paper-select remaining todos:
- [ ] Keyboard quick filter is something everyone uses on desktop - e.g open states select and type FL should highlight FL
- [ ] Async in demo does not display loading/progress navigation
- [ ] Keyboard up/down keys don't work consistently when paper-select is open

PR #398, opened 2016-06-18
Divider cleanup, docs and test

andybeesee:
Nothing fancy here...

Added documentation and integration test for divider component. 

Removed insetAttr computed property, in favor of explicitly defined inset property which defaults to null.

### DIFF
--- a/addon/components/paper-divider.js
+++ b/addon/components/paper-divider.js
@@ -2,13 +2,7 @@ import Ember from 'ember';
 
 export default Ember.Component.extend({
   tagName: 'md-divider',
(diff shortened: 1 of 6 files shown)


miguelcobain:
@andybeesee Thank you for the PR.

JSCS tests are failing.
What happens when `inset` is `false`? Isn't the attribute added anyway?

andybeesee:
Forgot to run JSCS, will have that cleaned up in a bit. 

I didn't realize that the attribute would be added anyway, I'll revert that and add a test for it too.

miguelcobain:
Thanks for your contribution!

PR #400, opened 2016-06-23
changed paper-select to call onChange instead of setting value directly

ibarrick:
paper-select was never calling the onChange action even though it is required.


miguelcobain:
We will accept immediate fixes for not yet updated components.
Thanks for this.

PR #403, opened 2016-06-27
removed action handlebars helper and replaced it a click event handler inside js

agarcia17:
Fixes issue with action helper not passing the event object


agarcia17:
if you go to the demo page and expand one of the menus that has a submenu you will see the bug in the console: 'Uncaught TypeError: Cannot read property 'stopPropagation' of undefined because the event object was never being passed in

miguelcobain:
Any idea why tests are failing?

dustinfarris:
Should rebase to get CI fixes from master

agarcia17:
@dustinfarris thanks! :)

miguelcobain:
thanks @agarcia17 !

PR #406, opened 2016-06-29
Fixed double sending of onInvalid action

shoxter:
Cleaned up the code a little to use the local var.

Main point: Fix specifically when onInvalid sends.

Because onInvalid is triggered when !== to lastIsInvalid, null and false values triggered two separate onInvalid calls.

As far as validity goes, these two are one in the same. Added !! to the if so that if the case is null !== false, it will now be false !== false and not send the extra onInvalid call.


SaladFork:
Does it make sense for the action to send up the value (`null`, `false`, `true`) or the boolean coercion of it (`false`, `true)`?

shoxter:
We can do either. That's up to you guys. Both can work for me.

I think the boolean coercion would make it simpler for people not reading the source code of the paper-input; however, differentiating null from false in the component helps tell whether it's invalid prior to being touched or after.

shoxter:
Implementing a better solution and then I'll reopen.

shoxter:
Not sure why the tests fail? They pass with ember test...

miguelcobain:
What does this include the paper-form component, just like #408?
Can't we just consider one over the other?

addon/components/paper-form-container.js
```diff
@@ -0,0 +1,20 @@
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+	tagName: '',
+	attributeBindings: ['style'],
+	style: "width:100%",
```
miguelcobain: We're only using single quotes. This should lead to a failing jscs error in the tests.

PR #408, opened 2016-06-29
Finished Paper Form Component

shoxter:
Usage like this:

```
{{#paper-form parentSubmit=(action 'useless') as |form|}}

         {{form.input value=foo onChange=(action (mut foo)) label="Foo" required=true}}
         {{form.input value=bar onChange=(action (mut bar)) label="Bar"}}

         {{#paper-button primary=true raised=true onClick=form.submit disabled=form.isInvalid}}Button{{/paper-button}}

{{/paper-form}}
```

Different Options: 

- **form.input:** Paper inputs inside of the form.
- **parentSubmit:**  The action in the parent component/controller to execute (clear fields, create/save records, etc) with regards to the form prior to the form component resetting the isTouched property on all of the inputs.
- **form.submit:** This should be the onClick method for the intended submit button for the form.
- **form.isInvalid:** This is the property that the form changes as the form's inputs change their validity state. When the inputs are invalid, this will be true and should disabled the form's intended submit button.

The form looks at all the form.inputs with input constraints for invalidity.

Please post any questions. We can create a page on the dummy app to cover the documentation for this component once we cover any questions that should be included.


miguelcobain:
Should I consider this PR or https://github.com/miguelcobain/ember-paper/pull/406 ?

shoxter:
@miguelcobain Both of them.

Multiple input validation in a form setting wasn't possible without the fix to inputs in #406.

I first had to fix the inputs to not send separate isInvalid calls for both 'NULL' and 'TRUE' values.

It was effectively double registering the invalid state of the input. #406 fixes that and this is the form component.

DanChadwick:
Maybe we should have `isValad` so null is better aligned with false.

addon/components/paper-form.js
```diff
@@ -4,7 +4,7 @@ export default Ember.Component.extend({
 	tagName: '',
 	classNames: [],
 	attributeBindings: ['style'],
-	style: "width:100%",
+	style: 'width:100%',
 	numberOfInvalids: 0,
 	isValid: Ember.computed('numberOfInvalids', function() {
 		return this.get('numberOfInvalids') === 0 ? true : false;
```
miguelcobain: `return condition ? true : false;` is equivalent to `return condition;`

addon/components/paper-form.js
```diff
@@ -0,0 +1,41 @@
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+	tagName: '',
+	classNames: [],
+	attributeBindings: ['style'],
+	style: 'width:100%',
+	numberOfInvalids: 0,
+	isValid: Ember.computed('numberOfInvalids', function() {
+		return this.get('numberOfInvalids') === 0 ? true : false;
```
miguelcobain: `return condition ? true : false;` is equivalent to `return condition;`

addon/components/paper-form.js
```diff
@@ -0,0 +1,41 @@
+import Ember from 'ember';
+
+export default Ember.Component.extend({
```
miguelcobain: destructure all the things

addon/components/paper-form-container.js
```diff
@@ -0,0 +1,20 @@
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+	tagName: '',
+	attributeBindings: ['style'],
+	style: 'width:100%',
+	customValidations: [],
+	hasValidation: Ember.computed('min', 'max', 'errorMessages', 'maxlength', 'required', 'customValidations', function() {
```
miguelcobain: destructure all the things

addon/components/paper-form-container.js
```diff
@@ -0,0 +1,20 @@
...
+		return this.get('min') || this.get('max') || this.get('errorMessages') || this.get('maxlength') || this.get('required') || this.get('customValidations');
+	}),
+	didUpdateAttrs: function(params) {
```
miguelcobain: use shorthand syntax `didUpdateAttrs(params) {`

addon/components/paper-form-container.js
```diff
@@ -0,0 +1,20 @@
...
+		if (params.oldAttrs.isTouched === params.newAttrs.isTouched) return;
+		this.set('touched', this.get('isTouched'));
+	},
+	actions: {
+		callOnInvalid: function(status) {
```
miguelcobain: shorthand syntax here as well

addon/components/paper-form.js
```diff
@@ -0,0 +1,46 @@
...
+export default Component.extend({
+  tagName: '',
+  classNames: [],
+  attributeBindings: ['style'],
+  style: 'width:100%',
+  numberOfInvalids: 0,
+  isValid: Ember.computed('numberOfInvalids', function() {
+    return this.get('numberOfInvalids') === 0;
+  }),
+  isInvalid: Ember.computed('isValid', function() {
```
miguelcobain: destructure `computed`

addon/components/paper-form.js
```diff
@@ -0,0 +1,46 @@
...
+    return !this.get('isValid');
+  }),
+  sendToParent: Ember.on('init', Ember.observer('isValid', function() {
```
miguelcobain: destructure `on`

miguelcobain: destructure `observer`

addon/components/paper-form-container.js
```diff
@@ -0,0 +1,24 @@
+import Ember from 'ember';
+const { Component } = Ember;
+
+export default Component.extend({
+  tagName: '',
+  attributeBindings: ['style'],
+  style: 'width:100%',
+  onChange: null,
+  customValidations: [],
+  hasValidation: Ember.computed('min', 'max', 'errorMessages', 'maxlength', 'required', 'customValidations', function() {
```
miguelcobain: destructure `computed`

addon/components/paper-form.js
```diff
@@ -0,0 +1,46 @@
+import Ember from 'ember';
+const { Component } = Ember;
+
+export default Component.extend({
+  tagName: '',
+  classNames: [],
+  attributeBindings: ['style'],
+  style: 'width:100%',
+  numberOfInvalids: 0,
+  isValid: Ember.computed('numberOfInvalids', function() {
```
miguelcobain: destructure `computed`

miguelcobain:
can't you work around the `touchedTrigger` by using `_isTouched` instead of `isTouched` on the templates?

shoxter:
@miguelcobain No because the value will never change in the parent component. The form is never going to set isTouched = true. So oldAttrs will always = newAttrs and will never execute. Unless I call a this.set('isTouched', true) then this.set('isTouched', false) which is more hacky than this solution.

miguelcobain:
Can you edit this twiddle to reproduce that problem?
https://ember-twiddle.com/5d15b9f7ad4092d7bfabc718751c2137?openFiles=components.my-component.js%2Ctemplates.components.my-component.hbs

shoxter:
@miguelcobain https://ember-twiddle.com/5d15b9f7ad4092d7bfabc718751c2137?openFiles=controllers.application.js%2Ctemplates.components.my-component.hbs

miguelcobain:
How do I reproduce?

shoxter:
@miguelcobain It didn't save for whatever reason. 

This is the right one: https://ember-twiddle.com/d7bc618f48a24a89e1bd9411c0d49050/91f4de4c453bcbe8650134d9d72c026ead80cf45?openFiles=controllers.application.js%2C

1.) Toggle controller.
2.) Toggle component.
3.) Toggle controller again.

addon/components/paper-form.js
```diff
@@ -0,0 +1,28 @@
...
+import ParentMixin from 'ember-paper/mixins/parent-mixin';
+
+const { Component, computed, on, observer } = Ember;
+
+export default Component.extend(ParentMixin, {
+  tagName: '',
+  isValid: computed('childComponents.@each.isInvalid', function() {
+    return !this.get('childComponents').isAny('isInvalid');
+  }),
+  isInvalid: computed('isValid', function() {
```
miguelcobain: We can use `computed.not` here

addon/mixins/child-mixin.js
```diff
@@ -0,0 +1,19 @@
+import Ember from 'ember';
+const { Mixin } = Ember;
+
+export default Mixin.create({
+  asdfgh: 'lel',
```
miguelcobain: Maybe we should remove this. :)

shoxter: Fixed.

addon/components/paper-form.js
shoxter: Done.

addon/mixins/child-mixin.js
DanChadwick: OH NO! My code depended upon `asdfgh`. :stuck_out_tongue_winking_eye:

shoxter: Lel!

addon/components/paper-form.js
```diff
@@ -0,0 +1,28 @@
...
+import ParentMixin from 'ember-paper/mixins/parent-mixin';
+
+const { Component, computed, on, observer } = Ember;
+
+export default Component.extend(ParentMixin, {
+  tagName: '',
+  isValid: computed.not('childComponents.@each.isInvalid', function() {
+    return this.get('childComponents').isAny('isInvalid');
+  }),
+  isInvalid: computed.not('isValid', function() {
```
miguelcobain: Please read docs on computed.not. it doesn't take a callback

shoxter: Sorry about that. I was in a rush out the door. I'll fix it when I get back.

dustinfarris:
If I want to use a custom paper-input component, is there an easy way to plug that in here?

shoxter:
`
{{#paper-form as |form|}}
  {{custom-paper-input parentComponent=form}}
{{/paper-form}}
`

DanChadwick:
@dustinfarris: I agree. Any component yielded should be created from an attribute. Contextual components tightly couple. :-1:

shoxter:
@DanChadwick The paper-input is an attribute of the contextual component because it's the most common input yielded in the form.

It's impossible to take into account all possibilities in the contextual component (which I think is what you're saying), but you can make any paper-input based component a part of the form by adding `parentComponent=form` inside of the form component.

dustinfarris:
@shoxter thanks for that example.  what if i'm using an intermediary component that proxies to paper-input?  e.g. my usage here https://github.com/DockYard/ember-changeset/issues/59#issuecomment-231567141

do i pass parentComponent down the chain?  or just to the first component?  is this officially supported?

shoxter:
@dustinfarris No problem.

The reason `parentComponent=form` works is because the form looks at all children components (components that you've specified that their parent is the form) for isInvalid. 

`isInvalid` is a property of paper-input, so unless your proxy component has a property that aliases the isInvalid property of the paper-input in your template, you'll need to have it on the paper-input.

Make sense?

dustinfarris:
@shoxter i think so, i'm just wondering how this will work with an intermediary component (i guess i'll find out soon enough!)

e.g.

#### my-route.hbs

```hbs
{{#paper-form as |form|}}
  {{x-paper-input parentComponent=form}}
{{/paper-form}}
```

#### x-paper-input.hbs

```hbs
{{paper-input parentComponent=parentComponent}}
```

miguelcobain:
@dustinfarris x-paper-input extends paper-input?

dustinfarris:
@miguelcobain no, it is a wrapper

dustinfarris:
@miguelcobain see Slack for extended conversation on this

miguelcobain:
@dustinfarris should be fixed with https://github.com/miguelcobain/ember-paper/commit/35d457fac0f18dd562db25e14d4d7d85e1741620

PR #410, opened 2016-06-30
added paper-select-header component and corresponding example

ibarrick:
Added a component for the select header and an example for it that mirrors the one in Angular Material.


addon/components/paper-select-header.js
```diff
@@ -0,0 +1,6 @@
+import Ember from 'ember';
+
+
+export default Ember.Component.extend({
```
miguelcobain: We're not using any `Ember.` in the codebase.
Please `const { Component } = Ember;`

tests/dummy/app/controllers/demo/select.js
```diff
@@ -7,7 +7,17 @@ export default Ember.Controller.extend({
...
-
+  vegetables: Ember.A([
+    {name: "Corn", checked: false},
+    {name: "Onions", checked: false},
+    {name: "Kale", checked: false},
+    {name: "Arugula", checked: false},
+    {name: "Peas", checked: false},
+    {name: "Zucchini", checked: false}]),
+  searchTerm: "",
+  filteredVegetables: Ember.computed.filter('vegetables',function(vegetable) {
```
miguelcobain: `const { computed } = Ember;` here as well

miguelcobain: space after comma

tests/dummy/app/controllers/demo/select.js
```diff
@@ -7,7 +7,17 @@ export default Ember.Controller.extend({
...
+    return vegetable.name.toLowerCase().indexOf(this.get('searchTerm').toLowerCase()) > -1;
+  }).property('vegetables','searchTerm'),
```
miguelcobain: you can use `computed.filter(...).property(...)`? I didn't know that. :/

shoxter: @miguelcobain Yes. It allows you to trigger the filter based on multiple properties. The only usage rule for this format is that in the .property(...) section, the first property MUST be the property that the .filter('property') is filtering.

ibarrick:
Is there anything required on my end for this?

tests/integration/components/paper-select-header-test.js
```diff
@@ -0,0 +1,24 @@
+import { moduleForComponent, test } from 'ember-qunit';
```
miguelcobain: remove this file, since we're not actually testing anything meaningful here.

tests/dummy/app/templates/demo/select.hbs
```diff
@@ -99,4 +98,40 @@
...
+    value=selectedVegetable
+    itemLabelCallback=vegetableLabelCallback
+    onChange=(action (mut selectedVegetable))
+    as |items|
+  }}
+    {{#paper-select-header class="demo-select-header"}}
+      {{input value=searchTerm
+              type="search"
+              placeholder="Search for a vegetable.."
+              class="demo-header-searchbox _md-text"
```
miguelcobain: I think it's safe to remove `_md-text` class.
It doesn't look like it is used for enything in this case. Also, it may confuse users.

tests/dummy/app/templates/demo/select.hbs
```diff
@@ -99,4 +98,40 @@
...
+    {{! BEGIN-SNIPPET select.select-header }}
+  {{#paper-select
+    placeholder="Vegetable"
+    value=selectedVegetable
+    itemLabelCallback=vegetableLabelCallback
+    onChange=(action (mut selectedVegetable))
+    as |items|
+  }}
+    {{#paper-select-header class="demo-select-header"}}
+      {{input value=searchTerm
```
miguelcobain: We should use 
```hbs
<input
  type="search"
  value={{searchTerm}}
  class="demo-header-searchbox"
  placeholder="Search for a vegetable.."
  oninput={{action "searchKeyPressed" value="target.value"}}>
```

1. `{{input` helpers are being less used because they are two way bound by default
2. we should use `input` instead of `keydown` events because one may input text without typing a key (e.g pasting)

tests/dummy/app/templates/demo/select.hbs
```diff
@@ -99,4 +98,40 @@
...
+              type="search"
+              placeholder="Search for a vegetable.."
+              class="demo-header-searchbox _md-text"
+              key-down="searchKeyPressed"}}
```
miguelcobain: see above comment to use `input` events

tests/dummy/app/controllers/demo/select.js
```diff
@@ -66,6 +82,11 @@ export default Ember.Controller.extend({
 
   vegToppings: Ember.computed('toppings.[]', function() {
     return this.get('toppings').filter((item) => item.category === 'veg');
-  })
+  }),
+  actions: {
+    searchKeyPressed(e) {
+      e.stopPropagation();
```
miguelcobain: Why are you stopping propagation?
Maybe if we use `<input` instead of `{{input` this action may do something meaningful like `this.set('searchTerm', value);`?

tests/dummy/app/controllers/demo/select.js
```diff
@@ -66,6 +82,11 @@ export default Ember.Controller.extend({
 
   vegToppings: Ember.computed('toppings.[]', function() {
```
miguelcobain: we can simplify this using computed macros:
```js
vegToppings: computed.filterBy('toppings', 'category', 'veg')
```

tests/dummy/app/controllers/demo/select.js
```diff
@@ -1,13 +1,27 @@
...
+    { name: 'Onions', checked: false },
+    { name: 'Kale', checked: false },
+    { name: 'Arugula', checked: false },
+    { name: 'Peas', checked: false },
+    { name: 'Zucchini', checked: false }]),
+  searchTerm: '',
+  filteredVegetables: computed('vegetables.[]', 'searchTerm', function() {
+    return this.get('vegetables').filter(function(vegetable) {
+      return vegetable.name.toLowerCase().indexOf(this.get('searchTerm').toLowerCase()) > -1;
+    }.bind(this));
```
miguelcobain: We shouldn't use bind. Use arrow functions instead:
```js
return this.get('vegetables').filter((vegetable) => {
  return vegetable.name.toLowerCase().indexOf(this.get('searchTerm').toLowerCase()) > -1;
});
```

tests/dummy/app/controllers/demo/select.js
```diff
@@ -1,13 +1,27 @@
 import Ember from 'ember';
 
+const { computed } = Ember;
+
 export default Ember.Controller.extend({
```
miguelcobain: destructure `Controller`

tests/dummy/app/controllers/demo/select.js
```diff
@@ -1,13 +1,27 @@
...
 
   userState: '',
   states: Ember.computed(function() {
```
miguelcobain: Use the already destructured `computed` instead of `Ember.computed`.

miguelcobain:
@ibarrick I added some inline comments. Other than those, it looks good to me!

tests/dummy/app/controllers/demo/select.js
ibarrick: Regarding the computed.filterBy, the search should not be case sensitive so that won't work.

tests/dummy/app/controllers/demo/select.js
ibarrick: Without stopping propagation, pressing a key was dismissing the select menu.

miguelcobain:
💯 

Thanks again!

PR #415, opened 2016-07-05
Prevent redundant on invalid

shoxter:
There was an issue with onInvalid being sent multiple times when isInvalid goes from TRUE <=> NULL.

Because these essentially both are invalid states, it's unnecessary to send multiple invalid actions.

### DIFF
--- a/addon/components/paper-input.js
+++ b/addon/components/paper-input.js
@@ -202,11 +202,15 @@ export default BaseFocusable.extend(ColorMixin, FlexMixin, {
 
   notifyInvalid() {
     let isInvalid = this.get('isInvalid');
-    if (this.get('lastIsInvalid') !== isInvalid) {
-      this.sendAction('onInvalid', this.get('isInvalid'));
-      this.set('lastIsInvalid', this.get('isInvalid'));
+    let lastIsInvalid = this.get('lastIsInvalid');
+    let isInvalidChanged = (lastIsInvalid !== isInvalid);
+    let isNotRedundant = !((lastIsInvalid === null && isInvalid) || (lastIsInvalid && isInvalid === null));
+    if (isInvalidChanged && isNotRedundant) {
+      this.sendAction('onInvalid', isInvalid);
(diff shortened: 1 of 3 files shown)


miguelcobain:
I like your recent changes.
However, I'm surprised that all tests have passed. There should be a test ensuring that error messages are not shown when the input isn't touched.
This can only mean that we didn't write enough tests.

So, I will ask you to:

1. write a test that:

  1. renders an invalid paper-input (you can just pass an `errors` array for simplicity)
  2. assert that no error messages are shown
  3. "touch" the input (trigger a blur event on `<input>`)
  4. assert that error messages are shown

2. the above test should fail because of this: https://github.com/miguelcobain/ember-paper/blob/master/app/templates/components/paper-input.hbs#L80
previously `isInvalid` depended on `isTouched`. We now probably need to create something like `isInvalidAndTouched` using `computed.and` or `{{#if isTouched}}{{#if isInvalid}}`.

As usual I'm available on slack.

addon/components/paper-input.js
```diff
@@ -46,14 +46,13 @@ export default BaseFocusable.extend(ColorMixin, FlexMixin, {
...
-   *    null: input has not yet been touched, but would be invalid if it were
+   * @return {boolean} Whether the input is or would be invalid.
    *    false: input is valid (touched or not), or is no longer rendered
    *    true: input has been touched and is invalid.
    */
-  isInvalid: computed('isTouched', 'validationErrorMessages.length', 'isNativeInvalid', function() {
+  isInvalid: computed('validationErrorMessages.length', 'isNativeInvalid', function() {
     let isInvalid = this.get('validationErrorMessages.length') || this.get('isNativeInvalid');
-    return isInvalid && !this.get('isTouched') ? null : !!isInvalid;
+    return isInvalid;
```
miguelcobain: we can make this just:
```js
isInvalid: computed.or('validationErrorMessages.length', 'isNativeInvalid'),
```

miguelcobain:
:100: thanks.

PR #416, opened 2016-07-06
Added keyboard support for paper-menu

ibarrick:
I also made the menu-items contextual components in order to use the ParentMixin and ChildMixin. Obviously neither is really necessary but you already had ParentMixin so I assumed that's the direction you wanted to go.

### DIFF
--- a/addon/components/paper-menu-content-inner.js
+++ b/addon/components/paper-menu-content-inner.js
@@ -1,6 +1,6 @@
 import Ember from 'ember';
 import ParentMixin from 'ember-paper/mixins/parent-mixin';
-const { Component, inject, computed, run: { later } } = Ember;
+const { Component, inject, computed, run } = Ember;
 
 export default Component.extend(ParentMixin, {
   tagName: 'md-menu-content',
@@ -11,7 +11,7 @@ export default Component.extend(ParentMixin, {
   enabledMenuItems: computed.filterBy('childComponents', 'disabled', false),
 
   didInsertElement() {
-    later(() => {
+    run.later(() => {
       let focusTarget = this.$().find('.md-menu-focus-target');
       if (!focusTarget.length) {
         focusTarget = this.get('enabledMenuItems.firstObject.element.firstElementChild');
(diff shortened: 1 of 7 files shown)


addon/components/paper-menu-content-inner.js
```diff
@@ -1,13 +1,24 @@
 import Ember from 'ember';
 import ParentMixin from 'ember-paper/mixins/parent-mixin';
-const { Component, inject } = Ember;
+const { Component, inject, computed, run: { later } } = Ember;
```
miguelcobain: generally, we are not destructuring more than one level to keep some context. So here, we would use `run.later`.

miguelcobain:
@ibarrick great job here. Thanks for pushing ember-paper.

Can I ask you for some tests, please? :pray: 
I'm available on slack if you need something cleared up.

app/templates/components/paper-menu-content-inner.hbs
```diff
@@ -0,0 +1 @@
+{{yield this}}
```
miguelcobain: can't you `yield` the `paper-menu-item` from here instead of yielding `this`.
yielding this is considered an anti pattern because we may unecessarily be leaking context.
These are private components, but still, if we can avoid it, I think we should.

app/templates/components/paper-menu-content.hbs
```diff
@@ -5,8 +5,8 @@
       class="ember-basic-dropdown-content {{class}} {{if renderInPlace 'ember-basic-dropdown-content--in-place'}} {{if hPosition (concat 'ember-basic-dropdown-content--' hPosition)}} {{if vPosition (concat 'ember-basic-dropdown-content--' vPosition)}}
       md-open-menu-container md-whiteframe-z2 md-clickable {{if isActive "md-active"}}"
       dir={{dir}}>
-      {{#paper-menu-content-inner width=width dense=dense dropdown=dropdown}}
-        {{yield}}
+      {{#paper-menu-content-inner width=width dense=dense dropdown=dropdown as |content|}}
+        {{yield (hash menu-item=(component 'paper-menu-item' parentComponent=content))}}
```
miguelcobain: Here we would just "re-yield" using `{{yield content.paper-menu-item}}` or something similar.

PR #418, opened 2016-07-06
Whitelists google font stylesheets in CSP snippet.

patrickberkeley:
The readme is updated, but it looks like the docs' snippet was missed.


miguelcobain:
Thanks!

PR #421, opened 2016-07-08
typo in Input example for errorMessages required

iirving:
minor typo in the documentation of the example app, but might cause some confusion for new users doing a naive copy and paste, which would still build but would display the default required error msg and not the custom errorMessage specified.


miguelcobain:
Thanks!

PR #422, opened 2016-07-08
allow SafeString in paper-inputs validation errorMessages

iirving:
Issue arose when using the ember-i18n add-on with paper, specifically for providing translation for paper input validator errorMessages
```
 {{paper-input
     value=identification
     label=(t "label.email")
     class="md-block"
     onChange=(action (mut identification))
     required=true
     errorMessages=(hash required=(t "paper.required”)) 
}} 
```  
this returned a console error: `vendor.js:151199Exception with validation:  Object TypeError: str.replace is not a function”  `

temporary work around would involve doing ; 
errorMessages=(hash required=(get (t "paper.required") "string"))

(@miguelcobain further suggest looking into Ember.String.loc and why it doesn't work with safe strings :/ )


miguelcobain:
Thanks!

PR #431, opened 2016-07-12
ember-power-select integration

ibarrick:
Extended all of the e-p-s classes and rewrote the demos to have the new API


ibarrick:
This should be ready to go now.

PR #433, opened 2016-07-12
fix to prevent race condition that leaves ghost in the dom

ibarrick:


miguelcobain:
in what browser did you have this problem?

ibarrick:
Google Chrome. It's pretty much the same issue as the one I consulted with you on in e-p-s integration. It's rare to trigger but when it does it requires refresh to remove the clone.

PR #436, opened 2016-07-13
Update latest version copy as 1.0-alpha is default

SaladFork:
Since `ember install ember-paper` now installs 1.0 by default we need to update the copy to reflect this. It looks like the changes have already been made to the snippets but it hasn't been deployed to github pages yet.


DanChadwick:
LGTM

PR #437, opened 2016-07-14
Add YUIDoc scaffolding

dustinfarris:


dustinfarris:
@miguelcobain can we get this in before it becomes a rebase nightmare?

miguelcobain:
Thank you for this.

PR #441, opened 2016-07-20
fixed fastboot service check

davidpett:
using the `FastBoot` global to determine if it is in FastBoot mode


miguelcobain:
Thanks a lot.

PR #446, opened 2016-07-21
Upgrades to latest wormhole version.

jordpo:
I was running into an issue when acceptance testing using the `paper-dialog` component which depends upon `ember-wormhole`.

`Assertion Failed: You have turned on testing mode, which disabled the run-loop's autorun. You will need to wrap any code with asynchronous side-effects`

Upgrading fixes the issue.


miguelcobain:
Thanks!

PR #452, opened 2016-07-26
Fix #448: Sidebar not closable by tapping away on mobile devices

### LINKED ISSUE #448: Sidenav and others broken on iOS
On iPhone I can't close an open sidenav just by tapping away from it.

img_2956

---

Also, on the dummy site, things appear out of whack on an iPad mini:
#### Header is cut off

img_0109

---
#### Sidenav is broken:

img_0110

---
#### In landscape menu button is missing:

img_0111

konradjurk:
Backdrop component now listens to `touchEnd` event and calls `this.click()` when fired.

### DIFF
--- a/addon/components/paper-backdrop.js
+++ b/addon/components/paper-backdrop.js
@@ -34,6 +34,10 @@ export default Component.extend(TransitionMixin, {
   click(e) {
     e.preventDefault();
     this.sendAction('onClick', e);
+  },
+
+  touchEnd(e) {
+    this.click(e);
   }
 
 });


konradjurk:
In TravisCI it seems like  `EMBER_TRY_SCENARIO=ember-beta` is blocking a few PRs. Do you think it makes sense do move this beta test flavor to "allowed failures" group? Ember-canary is already in there :)

miguelcobain:
@konradjurk failing on ember-beta is a good thing. It allows us to fix the problems before that ember version is released and possibly detecting bugs in advance. Ideally we should fix this. :(

miguelcobain:
Also, if I understand correctly, this doesn't fix the whole #448 issue.

miguelcobain:
i thought that click events did trigger in touch devices as well. So I'm not sure why this happens.

@arschmitz, can you help explaining this, please?

konradjurk:
@miguelcobain True! I tried to make that clear in the title by saying which part is fixed. Not sure if thats the right way to do it 💃 :)

miguelcobain:
@konradjurk Perfect. Thanks for taking the time for this.
I just need a clear explanation of why this happened. We had touch event listeners on backdrop before. They were removed because click events also did trigger on touch devices, supposedly. Just trying to make sure we don't fall into this trap again.

dustinfarris:
@konradjurk because of the wording in your commit message, #448 will automatically close when merged to master

miguelcobain:
> @konradjurk because of the wording in your commit message, #448 will automatically close when merged to master

Precisely. Hence my question.

konradjurk:
@dustinfarris Oops, didn't know that! Because of the issue reference or the "Fix" prefix?

dustinfarris:
@konradjurk, the fix prefix.  Check it out: https://help.github.com/articles/closing-issues-via-commit-messages/

konradjurk:
@dustinfarris Thanks for the link!

I could imagine that `click` is not registered on iOS because of this 300ms-trying-to-detect one of these [gestures](https://developer.apple.com/library/ios/documentation/AppleApplications/Reference/SafariWebContent/HandlingEvents/HandlingEvents.html) thingy

miguelcobain:
That's right. We need to use [ember-hammertime](https://github.com/runspired/ember-hammertime).

The problem is that its AST walker doesn't consider `<div onlick="` cases.

@konradjurk can you test if adding `touch-action: manipulation; -ms-touch-action: manipulation; cursor: pointer;` css to the backdrop will make it trigger `click` handler on iOS?

konradjurk:
@miguelcobain Yes

miguelcobain:
"Yes" as in "I can test it" or "The click handler is triggered"?

konradjurk:
@miguelcobain Yes I can test it. Not now though, I need some sleep first :)

arschmitz:
@miguelcobain click should trigger but with the 300ms delay as mentioned above. if your using hammertime `touch-action: manipulation; -ms-touch-action: manipulation; cursor: pointer;` is definitely needed. if this does not work i can look into this further.

miguelcobain:
@arschmitz the problem was that click wasn't triggering at all on iOS devices, only touch events. Android and desktop working fine.

arschmitz:
@miguelcobain if the listener is on a normally unclickable element ( like a div ) on iOS you need to add either a `tabindex` or `cursor: pointer;` or you wont get any click at all.

miguelcobain:
@arschmitz that may very well be the problem. We have a `tagName: 'md-backdrop'` here.

@dustinfarris @konradjurk ^

konradjurk:
@miguelcobain @arschmitz 🎉  Nice, it works with your suggestion to put `touch-action: manipulation; -ms-touch-action: manipulation; cursor: pointer;` on `md-backdrop`!

Unsurprisingly we now have a pointer-cursor all over the screen in small non-touch devices. Do you think we should add this conditionally only for touch devices?

miguelcobain:
@konradjurk instead of that, can you try to set a tabindex attribute of 0 and see if click handler triggers?

That CSS also solves the 300 ms problem, which is also an existing problem, but a different one which should be solved project wide.

konradjurk:
@miguelcobain That does not work

miguelcobain:
This was fixed on master. Thanks. 👍

PR #454, opened 2016-07-28
Fix ripple behaviour for touch devices

konradjurk:
The animation didn't start which made everything feel more slow and static.

Next: To make it more high fidelity and get it closer to the "Google Inbox App" we need a faster animation as the one which is used right now is too slow. But I'll put that into a separate pull request as it seems like we can't change the md-angular stylesheets (?) so I had to make the animation faster by increasing the ripple size 👯  Maybe you don't want such a hack in your app :)

### DIFF
--- a/addon/mixins/ripple-mixin.js
+++ b/addon/mixins/ripple-mixin.js
@@ -126,9 +126,8 @@ export default Mixin.create({
 
   },
   bindEvents() {
-    this.rippleElement.on('mousedown', run.bind(this, this.handleMousedown));
-    this.rippleElement.on('mouseup touchend', run.bind(this, this.handleMouseup));
-    this.rippleElement.on('mouseleave', run.bind(this, this.handleMouseup));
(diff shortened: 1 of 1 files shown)


miguelcobain:
Again, I don't think this is a maintainable solution.

These are the events that angular material uses: https://github.com/angular/material/blob/master/src/core/services/ripple/ripple.js#L245-L248

The solution should be a fastclick solution project wide, that would solve all of these problems at once. Otherwise we're going to have little issues here and there forever.

konradjurk:
Doesn't hammer.js already solve this? If I am not totally wrong that comes with ember-cli by default. At least I got it in my workspace and can't remember that I've installed it

miguelcobain:
@konradjurk looks like ember-hammertime recently landed support for `<div onclick={{action` cases, which should solve pretty much all ember-paper click scenarios.

Could you try adding that addon and check if the problem is solved without changing anything in ember-paper itself?

PR #455, opened 2016-07-28
Faster ripple animation

konradjurk:
Quote from #454 
> To make it more high fidelity and get it closer to the "Google Inbox App" we need a faster animation as the one which is used right now is too slow. But I'll put that into a separate pull request as it seems like we can't change the md-angular stylesheets (?) so I had to make the animation faster by increasing the ripple size 👯 Maybe you don't want such a hack in your app :)

### DIFF
--- a/addon/mixins/ripple-mixin.js
+++ b/addon/mixins/ripple-mixin.js
@@ -215,7 +215,7 @@ export default Mixin.create({
     let height = this.rippleElement.prop('clientHeight');
     let x = Math.max(Math.abs(width - left), left) * 2;
     let y = Math.max(Math.abs(height - top), top) * 2;
-    let size = getSize(this.get('fitRipple'), x, y);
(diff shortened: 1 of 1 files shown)


miguelcobain:
Please help me understand this.
You have a preference for faster ripple animations? We should be sticking with AM's ripples.
Do the ripples feel any different on iOS than on normal desktop browsers?

konradjurk:
Sure, I recorded two videos:

Inbox by Google: https://youtu.be/38zbvKH7X_8
Ember Paper App: https://youtu.be/1pgkRgw3Sek

On the first glance, it might look similar. Both sidebars seem to open equally quick. But googles app feels more responsive as the ripple is inflated quicker and due to the increased size in comparison to Angular/Ember implementations the user can see it while the thumb is still over the screen.

miguelcobain:
https://github.com/miguelcobain/ember-paper/pull/475 now gives us the same ripple as latest AM versions

PR #469, opened 2016-08-20
Set the value before growing the textarea on render

Kerrick:
This fixes an issue with auto-grow when rendering a textarea that already had a `value` passed in (without user input).

### DIFF
--- a/addon/components/paper-input.js
+++ b/addon/components/paper-input.js
@@ -155,9 +155,9 @@ export default BaseFocusable.extend(ColorMixin, FlexMixin, ChildMixin, {
   },
 
   didRender() {
-    this.growTextarea();
     // setValue below ensures that the input value is the same as this.value
     this.setValue(this.get('value'));
+    this.growTextarea();
   },
 
   willClearRender() {


miguelcobain:
@shoxter I think you had this problem as well. Can you confirm this fixes it? It should.
Thanks a lot @Kerrick.

jamesdixon:
@miguelcobain I just ran into this issue as well and can confirm that @Kerrick's fix works as advertised.

miguelcobain:
@jamesdixon isn't it fixed on master? http://miguelcobain.github.io/ember-paper/release-1/#/components/input

jamesdixon:
@miguelcobain sort of, but not completely.

Note that this is for a text area that already has a value. If you go to the demo page and enter some text that spans multiple lines in the text area, navigate to another route and then come back, you'll notice there is a significant delay in the text area growing to accommodate the text. I am able to reproduce this locally as well.

However, if the code is modified to move `this.growTextArea()` to after the value is set, the delay is gone. 

```js
  didRender() {
    this._super(...arguments);
    // setValue below ensures that the input value is the same as this.value
    this.setValue(this.get('value'));
    this.growTextarea();
  },
```

Ericky14:
When is this PR getting merged? This fix works. Without this, textareas with a preset value do not grow when the page is loaded, some input is necessary.

Issue #471, opened 2016-08-24
paper-sidenav: when clicking the menu button so fast, the menu will always show open without backdrop


tmclouisluk:
Even the demo page of paper
when you clicking the menu button in a fast speed
the sidenav will not be closed and always show open with no backdrop

tmclouisluk:
seems paper-sidenav-inner click() function problem

tmclouisluk:
the other way to trigger this bug:
click: menu->backdrop->menu->backdrop->menu->backdrop->menu->backdrop...
many times. Then it comes the same result.

Browser: chrome 52.0.2743.116 m (64-bit)

miguelcobain:
I think using my updated fork of ember-css-transitions will solve these kind of problems.

tmclouisluk:
it doesn't work. The menu now is always open and cannot close :(

miguelcobain:
It is working fine here.

tmclouisluk:
When i dont use ember-css-transitions, then it works :8ball: 
`classNameBindings: ['positionClass','isLockedOpen:md-locked-open', 'closed:md-closed'],`

miguelcobain:
Like that it shouldn't have animations.

miguelcobain:
This should be fixed in latest master after ember-css-transitions update to 0.1.5.

PR #477, opened 2016-08-31
Fixed the sidenav bug

tmclouisluk:
by cancel the animation of open sidenav


tmclouisluk:
Problem: 
https://github.com/miguelcobain/ember-paper/issues/471

miguelcobain:
This fix doesn't make sense to me.
Example: https://github.com/angular/material/blob/82fdd2c7c1cd2eb3b84608e2245a353b561c1c28/src/components/sidenav/sidenav.scss#L35

The animation classes are completely ignored. Do sidenav opening animations work like this? They shouldn't.

Issue #388, opened 2016-05-29
paper-autocomplete suggestion list is not visible


szsoppa:
After switching stable release to newest 1.0.0-alpha.1 my autocomplete suggestion list appends and show itself at the end of body tag which makes it totally unuseful.

DanChadwick:
Can you elaborate about why this would is an issue for your use case?

The autocomplete component is "scheduled" to be refactored using paper-wormhole, which would effectively do the same thing. In what way is the location of the list in the DOM "unuseful"? Locating elements which need to be on top of everything else at the root of the DOM avoid issues where the parent's styling can interfere with the topmost element. This is how menus and dialogs work.

See paper-autocomplete-list.js:didInsertElement() for the relevant code.

Note that there is a partially-complete PR for porting this component to v1.0.

szsoppa:
I think I defined my problem in wrong way. The problem is that autocomplete suggestion list positions itself at the bottom of page. Under my autocomplete input there are some category cards. Suggestion list should be attached to the bottom of autocomplete input but it appears under category cards instead. Because of that user is not able to see suggestion list until he scrolls page down. This is why it's unuseful.

DanChadwick:
Ah, I understand now. Yes, it seems that the pop-up list in v1.0.0-alpha.1 is broken. I suspect the problem is that the styling that was relying on isn't present because we are now using the angular material styles. There is a PR to port autocomplete. You could try that as a starting point or contribute by porting autocomplete yourself to 1.0.

brendanoh:
I am migrating an app to v1 alpha 3. Am I to understand that Autocomplete is totally unusable in the current state?

Because currently none of the autocompletes work in my app and in fact the demo page autocompletes dont seem to work. At least not the way the old versions worked.

Or am I missing something?

brendanoh:
OK I am now looking at https://github.com/miguelcobain/ember-paper/pull/395 so I think I understand.

xomaczar:
Paper-menu branch fixes all the issues mentioned here. Please close

xomaczar:
Can this issue be closed?

PriyangaSG:
This issue still exists in version 1.0.0-beta.3

UserXXXX:
This issue still exists in version 1.0.0-beta.3

xomaczar:
Can you elaborate?

UserXXXX:
[...]

PR #478, opened 2016-08-31
Changed structure of autocomplete suggetion to match angular material css

Foggyman:
Changed structure of autocomplete suggestion to match angular material css


Foggyman:
This should fix the issue #388

PR #480, opened 2016-09-01
Changed structure of autocomplete suggestion to match angular material css

Foggyman:
This should fix the issue #388


miguelcobain:
We're in the middle of a big update to autocomplete. This will be addressed in #395.
Thanks.

PR #481, opened 2016-09-01
Fixed paper-autocomplete

xomaczar:
Paper-autocomplete was not working after upgrading to the latest ember-power-select.


miguelcobain:
I updated paper-select and menu yesterday. Is this the sequel? Great. :D

xomaczar:
Paper-autocomplete does not work with yesterday's update. This PR fixes it. Plus adds paper-autocomplete-options component to allow angular md styles to work.

xomaczar:
There is still an issue with virtual-repeat and keyboard down/up navigation which still needs to be fixed.

xomaczar:
I still need to add meaningful tests

miguelcobain:
Yesterday's update wasn't meant to fix autocomplete. I didn't touch it yet.

xomaczar:
I meant to say that before yesterday's update it was at least working and drop-down was displaying. Does this PR not address some of the needs?

miguelcobain:
Of course it does!
There is paper-menu, paper-select and paper-autocomplete. I updated paper-menu and paper-select.
I was saying I'm happy to see a continuation for my yesterday's update.

ibarrick:
@xomaczar Could you elaborate on the issue with Virtual Repeat? I can go and fix that today probably.

xomaczar:
Great. This is a continuation to your yesterday's work. Let me know when you can merge it. Let's make it a bit stable then add integration/unit tests.

xomaczar:
@ibarrick i think its a power-select issue and not necessary a virtual-repeat problem. When a user presses up/down keys power-select is looking for Dom elements with specific tags, which it then highlights. The issue occurs when one scrolls a long list with scrollbar and then presses up/down key - power-select is unable to find next/previous element to highlight.

ibarrick:
@xomaczar Does the AM component behave the that way?

xomaczar:
@ibarrick don't know - i just noticed it but did not have time to debug - this has a low priority to me. Just wanted a way to track it and fix it before 1.0 release

app/templates/components/paper-autocomplete-content.hbs
```diff
@@ -1,11 +1,12 @@
 {{#if dropdown.isOpen}}
-  {{#basic-dropdown/wormhole to=to renderInPlace=renderInPlace didInsert=(action "didOpen") willRemove=(action "willClose")}}
+  {{#ember-wormhole to=to renderInPlace=renderInPlace didInsert=(action "didOpen")}}
+  {{debugger}}
```
xomaczar: Remove thus

xomaczar:
@miguelcobain 'fixed' deprecation warnings by replacing virtual-repeat with {{each}} wrapped inside divs with AM classes. I'll need @ibarrick help on this.

miguelcobain:
There are some jscs errors on the latest commit.

miguelcobain:
Overall it looks very good!

From my brief manual testing:
  - I couldn't filter anything. e.g type `Portugal` and no results are found. Substrings should also work. It seems to work on the last examples, but not on the first.
  - the "no results" message is shown in a large dropdown
  - is the "create" functionality working?
  - there is a little show/hide animation on the little "x" in AM. Tip: ember-css-transitions provides a `transition-group` component that may help. Alternatively we can use its mixin and create a component for the "x". This is not a priority.
  - we're opting out of attribute-based flex functionality in v1. Instead we'll use classes. This should be simpler, since components can use `class` attribute.

ibarrick:
@xomaczar What were the deprecations?

xomaczar:
@ibarrick there were 2 types of deprecations:
1. Binding style attribute http://emberjs.com/deprecations/v1.x/#toc_warning-when-binding-style-attributes
2. Modifying properties inside didInsertElement hook

ibarrick:
@xomaczar Those seem easy enough to fix, I'll submit a PR. Should I go fix the jscs errors while I'm at it?

miguelcobain:
@ibarrick sure, why not? :)

xomaczar:
@ibarrick fixing deprecations would be awesome. To integrate virtual-repeat with autocomplete we'll need to resolve scroll related issues between the eps and virtual-repeat. ember-power-select assumes all matches/results are rendered in the DOM - which obviously not the case.

xomaczar:
@miguelcobain I'll take a look at the remaining items. I was just testing the bottom inline autocomplete with paper-input trigger. As far as flex attr is concerned I believe we discussed this in the past, and performance was the main issue. I believe paper-input on master also sets flex attr instead of the class, so we need to address this everywhere.

miguelcobain:
@xomaczar @ibarrick there is an eps + ember-collection addon: https://github.com/cibernox/ember-power-select-collection

Maybe we can take some ideas from there?

ibarrick:
@miguelcobain Maybe, but last time I looked at ember-collection for VR it was a bad fit. The VR implementation I have is extremely similar to AM so I'm guessing AM source has a viable solution

xomaczar:
@miguelcobain can you merge/rebase this branch against master - ripple fixes are not included in this branch and it hurts my eyes when I'm navigating the dummy app.

xomaczar:
**Completed:**
- [x]  You should be able to find Portugal now - the dummy app was not working properly.
- [x]  Added an animated reset/close button (had to create another component and use the transition mixin => in order to set tabindex on that element)

**Remaining:**
- [ ]  "no results" message is shown in a large dropdown. The containing element must set the height explicitly based on # of items shown - the dropdown.content is position absolutely.
- [ ] ensure "create" functionality works correctly. I played with AM "create" functionality and still don't understand how it works.
- [ ] class based flex vs. attribute - like I mentioned other components are suffering the same issue. Refactoring flex-mixin will solve this for others. (I'd prefer a separate PR for this)
- [ ] fix jscs errors @ibarrick

PR #482, opened 2016-09-07
Match paper-sidenav width with AM@1.0.6

xomaczar:


miguelcobain:
Nice!
The sidenav size is indeed smaller like in AM docs. But the scrollbars look different for me. Can you confirm that?

I'd like to understand why.

xomaczar:
Yes the webkit browsers will have different scrollbars. You don't like 'em?

xomaczar:
They kind of compliment the md style, nah? I am on ubuntu & chrome my scrollbars take away from this polished ember-paper demo ui.

miguelcobain:
@xomaczar i meant that with this PR I'm seeing different scrollbars from the angular material website. Can you confirm that?

xomaczar:
I confirm - I changed it on purpose

xomaczar:
And this is not part of AM

miguelcobain:
I think I didn't make myself clear.

Angular Material (1.0.6) scrollbar:
screen shot 2016-09-10 at 00 34 27

this PR's scrollbar:
screen shot 2016-09-10 at 00 34 33


Why are they different? This was my initial question.

xomaczar:
Cause I did not like how the native scrollbars looked in chrome/Firefox on ubuntu/windows. I can remove the scrollbars styles ...

miguelcobain:
I'm not opposed to any scrollbar in particular. I'm just trying to understand.

But doesn't this PR include the same styles as AM? That was my understanding. If it includes, why are they different?

xomaczar:
Yes this PR includes some of the AM demo/docs style - specifically width of the nav container - I added my scrollbar styles to further enhance demo app. I have no problem removing the scrollbar styles - to match am demo. Sorry for the confusion...

miguelcobain:
Ah, ok. I thought these were scrollbar styles from AM. Didn't understand why the scrollbars were different. Got it. :)

xomaczar:
What are we doing with this PR?

miguelcobain:
Merging. :)

xomaczar:
Done... :)

PR #487, opened 2016-09-07
make progress-linear component visible when querying autocomplete w/ label

xomaczar:


xomaczar:
Can you guys review this PR. The first commit fixes progress-linear component visibility issues when  autocomplete with floating label is used to make xhr queries. The second commit fixes the dropdown trigger target - this allows a user to clear selection and set focus on input element - identical to am latest version. I know we've talked about matching the behavior of am@1.0.6, however I believe this was the intended behavior all along.

xomaczar:
I probably need to fix failing tests first. Thoughts?

xomaczar:
Ping @miguelcobain
I am not really happy with trigger target just wasn't bale to find a better solution yet.

xomaczar:
@miguelcobain Where did md-backdrop idea come from? It seems that AM is using a different approach. I can't keep autocomplete open and click on a trigger multiple times w/o it trying to close the dropdown.

xomaczar:
I think I figure it out. Don't worry - AM does similar thing...

xomaczar:
@ibarrick @miguelcobain last commit integrates paper-virtual-repeat into autocomplete. No deprecations and height adjust to its content - similar to AM. 
- Remaining items:
 1. Scroll the virtual-repeat/dropdown with keyboard UP/DOWN keys

miguelcobain:
I really like this PR, but it seems that selection is broken when the dropdown opens upwards.
When the autocomplete is on the lower edge of the viewport, the dropdown should open upwards.

Can you review that, @xomaczar? That was working.
Also, what exactly is the reason for the backport here? 

Other than that, it looks good!

xomaczar:
Backport?

miguelcobain:
@xomaczar yes, `app/styles/backports/paper-autocomplete.scss` file

xomaczar:
I believe it was related to linear-progress not being visible when using autocomplete with floating label

xomaczar:
I need to diff these two to be certain. Also, the old one needs to be excluded from importing

xomaczar:
Added: 
 1. Really using virtual-repeat (Inspect the DOM to see number of  <li> elements)
2. Keyboard UP/DOWN navigation.
3. Tab selection when an option is highlighted
4. Find & scroll to the selected option within virtual-repeat
cc @miguelcobain  @ibarrick

ibarrick:
@xomaczar autocomplete is working on the branch for my PR, did you duplicate that work or do it yourself?

xomaczar:
All those points I listed works on your branch? I thought you just fix deprecation warnings...

xomaczar:
@ibarrick I did it myself

ibarrick:
@miguelcobain @xomaczar We should decide on an implementation of VR before we do anything else. As of now there are 2 different implementations with different feature sets in PR's and we can't be very productive if we're all writing different implementations for the same things. :/

Can we merge either my PR or @xomaczar's so that we're all on the same page? I think it would be worth merging an imperfect version if it means we aren't stepping on each other's toes.

xomaczar:
Agree

ibarrick:
@xomaczar not all of them (missing the keyboard thing for sure) but I have it working in my own app with paper-autocomplete. I've even extended the new paper-select and successfully got VR working inside of that. In order to do that I had to add some things to VR which I'll have to add to yours to get similar things working.

If yours is stable then it may be the more complete implementation so once Miguel weighs in I may close my PR and submit some changes to your implementation. You should take a look at my PR in the mean time anyways because miguel had some comments that I think apply to your code (in particular the bit about computed properties) and we should make his life a little easier xD

xomaczar:
I'll take a look at your PR?

xomaczar:
Was your work if paper-select merged in, I need it for my app as well

Andrey S. Khomenko

ibarrick:
@xomaczar Since it wasn't part of AM at all I decided to do it separately in my own app. It relies on my version of VR though so once yours is merged I'll get it working with your version of VR and then extract it to an addon for you and anyone else who wants to use it.

xomaczar:
Why is select is not part of AM? Do you mean paper-select with virtual-repeat? I don't need that yet, just a stand alone paper-select with label.

ibarrick:
Oh I assumed you meant with VR. The paper-select in paper-menu should have labels now and be pretty usable. I think my work was not merged but Miguel accomplished the task all the same, if I'm remembering correctly.

xomaczar:
@miguelcobain tests pass in chrome

xomaczar:
Fixed virtual-repeat

xomaczar:
Now failing on beta -- unrelated to ember-autocomplete, virtual-repeat.

xomaczar:
@miguelcobain ready for review

miguelcobain:
🎉  thanks!

PR #488, opened 2016-09-08
Added labels to paper-select

ibarrick:
The failed tests are the paper-autocomplete tests. paper-select now supports placeholders or labels the same way paper-input does.

### DIFF
--- a/app/templates/components/paper-select-trigger.hbs
+++ b/app/templates/components/paper-select-trigger.hbs
@@ -9,5 +9,7 @@
   {{/if}}
 {{else if placeholder}}
   <span class="ember-power-select-placeholder">{{placeholder}}</span>
+{{else if label}}
+  <span class="ember-power-select-placeholder">{{label}}</span>
 {{/if}}
 <span class="md-select-icon" aria-hidden="true"></span>
--- a/addon/components/paper-menu-content-inner.js
+++ b/addon/components/paper-menu-content-inner.js
@@ -24,7 +24,9 @@ export default Component.extend(ParentMixin, {
       if (!focusTarget.length) {
         focusTarget = this.get('enabledMenuItems.firstObject.element.firstElementChild');
       }
-      focusTarget.focus();
+      if (focusTarget) {
+        focusTarget.focus();
+      }
     });
   },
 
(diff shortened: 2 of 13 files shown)


miguelcobain:
But what is the difference between label and placeholder?

ibarrick:
Same as the difference with paper-input. Placeholder doesn't float when you the input has value.

xomaczar:
what's the reason for not using paper-input as a trigger for paper-select where we need floating label? It would solve case for required=true and errors in general.

ibarrick:
@xomaczar The rendered DOM is different between the two and setting the "value" of the paper-input to be formatted like the block supplied by the user (the current way the e-p-s works) seems nontrivial. All in all it seems like it would be MUCH less effort to just extract the validation to a mixin so that it could later be added to things like the chips component, or any custom component.

It certainly isn't worth it to do all that just for labels (since I've already done the labels), and mixins seem like a better fit for validation so that users aren't coupled to adding paper-input to every single component they want validation for.

xomaczar:
Great. Just wanted to make sure we considered the alternative.

app/templates/components/paper-select-trigger.hbs
```diff
@@ -9,5 +9,7 @@
   {{/if}}
 {{else if placeholder}}
   <span class="ember-power-select-placeholder">{{placeholder}}</span>
+{{else if label}}
+  <span class="ember-power-select-placeholder">{{label}}</span>
```
miguelcobain: maybe something like this reads better
```hbs
...
{{else}}
  <span class="ember-power-select-placeholder">{{if placeholder placeholder label}}</span>
{{/if}}
```

miguelcobain:
I've added labels in paper-menu branch (https://github.com/miguelcobain/ember-paper/commit/d8f06dec87da8e3c2430d4454cdbb3e059620d78) with inspiration from this PR.
Thanks a lot!

I did not add searchMessage, which corresponds to your commit: https://github.com/miguelcobain/ember-paper/pull/488/commits/0773c797e68c5deb7c53ddb837f04823371b2b08

There seems to exist multiple places where we can specify the message.

I'll close this for now, but please give me some feedback on the searchmessages issue, either here or on slack. Thanks again.

PR #483, opened 2016-09-07
Fix deprecations in Virtual Repeat and reintegrate with paper-autocomplete

ibarrick:
Fixed the deprecations in Virtual Repeat and reintegrated with paper-autocomplete

PR #490, opened 2016-09-09
extracted validation to a mixin and added the mixin to paper-select

ibarrick:
Since it is no longer possible for paper-select to inherit paper-input like it was before, I extracted the validation logic that wasn't specific to paper-input into a mixin. Components only need to add the mixin and set validationProperty to be the name of the property validations will be run against. Components that use the mixin will also be responsible for calling this.notifyValidityChange() whenever they want to recheck for validity.

I think moving validation to a mixin will be a big help in including validation in future components such as chips and date picker. This PR is independent of the PR for adding labels so it will have to be adjusted slightly once that is merged.


app/templates/components/paper-virtual-repeat.hbs
```diff
@@ -1,6 +1,6 @@
 {{#paper-virtual-repeat-scroller onScroll=(action 'onScroll')}}
 	<div class="md-virtual-repeat-sizer" style={{contentStyle}}></div>
 	<div class="md-virtual-repeat-offsetter" style={{concat "transform: " (if horizontal "translateX(" "translateY(")  _marginTop "px);"}}>
```
miguelcobain: This `style={{concat ...` triggers a warning for not using htmlSafe in style attribute.

app/templates/components/paper-select.hbs
```diff
@@ -42,6 +42,24 @@
...
+      <div class="md-input-messages-animation md-auto-hide" ng-messages>
+        {{#each validationErrorMessages as |error index|}}
+          <div id="error-{{inputElementId}}-{{index}}"
+          class="paper-input-error ng-enter ng-enter-active" ng-message>
+            {{error.message}}
+          </div>
+        {{/each}}
+      </div>
+    {{/if}}
+  {{/unless}}
```
miguelcobain: I think we could benefit from creating a `{{paper-validation-errors` component with this logic.
As I understand, it is shared with paper-input, and possibly others in the future?

addon/mixins/validation-mixin.js
```diff
@@ -0,0 +1,117 @@
...
+/**
+ * @class ValidationMixin
+ * @extends Ember.Mixin
+ */
+export default Mixin.create({
+  validationErrorMessages: null,
+  lastIsInvalid: undefined,
+  validationProperty: '', // property that validation should be based on
+  init() {
+    this._super(...arguments);
```
miguelcobain: Assert here that `validationProperty` isn't empty.

ibarrick: @miguelcobain I'll just merge paper-menu in again and that will fix the style issue (I tried to keep the PR's independent of eachother).

What is the benefit of making validation a component instead of a mixin like I did?

ibarrick:
@miguelcobain I'll just merge paper-menu in again and that will fix the style issue (I tried to keep the PR's independent of eachother).

What is the benefit of making validation a component instead of a mixin like I did?

ibarrick:
@miguelcobain Actually I misread my notifications, it looks like the style issue will be fixed once #483 is merged. I thought that was one of the ones that you merged today.

ibarrick:
Fixed issues that you addressed, tests and style issue are waiting on #483. Thanks for reviewing!

miguelcobain:
I probably wasn't clear regarding `{{paper-validation-errors` component.
We have the same error message markup here: https://github.com/miguelcobain/ember-paper/blob/master/app/templates/components/paper-input.hbs#L74-L90

How about extracting that in a `{{paper-validation-errors` component and use something like:
```hbs
{{#unless hideAllMessages}}
  {{paper-validation-errors
    maxlength=maxlength
    renderCharCount=renderCharCount
    isInvalidAndTouched=isInvalidAndTouched
    inputElementId=inputElementId
    validationErrorMessages=validationErrorMessages}}
{{/unless}}
```

DRY makes refactoring and testing easier.

Also not that in your PR you use `inputElementId` which isn't defined, afaik?

miguelcobain:
Good job!

PR #493, opened 2016-09-12
referencing the correct API

SirZach:
without this change the browser throws an error when jquery tries to do a .indexOf on an anonymous function instead of a string


miguelcobain:
Thanks!
I'm going to ask you to reference this PR on CHANGELOG.md.

miguelcobain:
Thanks a lot!

SirZach:
Thanks!

PR #494, opened 2016-09-14
Call power-select's contentFor hook for body-footer

pauln:
Without this, attempting to use paper-autocomplete in an app (or other addon) results in:
ember-wormhole failed to render into '#ember-basic-dropdown-wormhole' because the element is not in the DOM

See https://github.com/cibernox/ember-power-select/issues/145#issuecomment-170012836 for more detail

### DIFF
--- a/index.js
+++ b/index.js
@@ -20,7 +20,7 @@ module.exports = {
     }
   },
 
-  contentFor: function(type) {
+  contentFor: function(type, config) {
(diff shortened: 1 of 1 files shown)


miguelcobain:
Thanks @pauln! This should solve a problem @xomaczar reported about not being able to install this branch in an app.

xomaczar:
Awesome

PR #498, opened 2016-09-19
Select disabled styling

ibarrick:
Had to extend basic-dropdowns trigger in order to add the attributeBinding for disabled to get the styling of disabled selects correct.

### DIFF
--- a/addon/components/paper-select-menu-trigger.js
+++ b/addon/components/paper-select-menu-trigger.js
@@ -1,0 +1,26 @@
+import Ember from 'ember';
+import BasicTrigger from 'ember-basic-dropdown/components/basic-dropdown/trigger';
+
+const { computed } = Ember;
+
+export default BasicTrigger.extend({
+  tagName: 'md-select',
+  attributeBindings: [
+    'role',
+    'tabIndex:tabindex',
+    'dropdownId:aria-controls',
+    'ariaLabel:aria-label',
+    'ariaLabelledBy:aria-labelledby',
+    'ariaDescribedBy:aria-describedby',
+    'aria-disabled',
+    'aria-expanded',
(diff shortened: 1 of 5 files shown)


addon/components/paper-select-menu-trigger.js
```diff
@@ -0,0 +1,26 @@
...
+    'ariaLabel:aria-label',
+    'ariaLabelledBy:aria-labelledby',
+    'ariaDescribedBy:aria-describedby',
+    'aria-disabled',
+    'aria-expanded',
+    'aria-haspopup',
+    'aria-invalid',
+    'aria-pressed',
+    'aria-required',
+    'disabled:disabled'
```
miguelcobain: `attributeBindings` is a concatenated property. This means that it will be concatenated when we `extend` the base class.
We just need to specify `attributeBindings: [ 'disabledAttr:disabled']` in this case, with the following CP:
```js
disabledAttr: computed('disabled', function() {
  return this.get('disabled') ? 'disabled' : null;
})
```

app/templates/components/paper-select-menu.hbs
```diff
@@ -3,7 +3,7 @@
   disabled=publicAPI.disabled
   actions=publicAPI.actions
   trigger=(component triggerComponent
-    tagName="md-select"
+    disabledProxy=publicAPI.disabled
```
miguelcobain: We can pass in `disabled=publicAPI.disabled` here. Why the need for `disabledProxy` naming?

miguelcobain:
I did the changes in https://github.com/miguelcobain/ember-paper/commit/311fa4bb57574e64d7567ceb3f05a6d9db14807f. Thanks!

PR #503, opened 2016-09-26
Upgrade paper-menu branch to ember-power-select 1.0.0-beta.19

pauln:
This fixes a memory leak, as well as fixing and improving a few other things.  Also bump minimum version of ember-basic-dropdown in core requirements to match that in e-p-s beta.19, to get some of the same fixes when using e-b-d directly.


miguelcobain:
Thanks!

PR #504, opened 2016-09-27
Add onBlur support to paper-autocomplete

pauln:
onFocus was already supported, but not onBlur.  Both are useful for handling focus events, so this adds basic onBlur support.  onBlur does need to be handled with some care though, as clicking on items within the dropdown causes the autocomplete trigger to blur.


mellatone:
@pauln several tests are failing due to this change. Can you take a look at those to make sure there aren't any regressions?

pauln:
@mellatone This PR is against the paper-menu branch - the failures should all be preexisting on that branch (see PR #395).  I've just done a quick diff and can't see any new failures compared to #395; if you can see any new failures which I've overlooked, please let me know and I'll gladly investigate.

app/templates/components/paper-autocomplete.hbs
```diff
@@ -34,7 +34,7 @@
       listboxId=(readonly optionsId)
       onFocus=(action "onFocus")
       activate=(action "activate")
-      onBlur=(action "deactivate")
```
miguelcobain: this means we're no longer calling `deactivate`? Can't that be problematic since eps does call it?

pauln: I couldn't see anything using `deactivate`, but if it's used somewhere I'm sure we can figure out a workaround.  I see we have `activate=(action "activate")` - would adding `deactivate=(action "deactivate")` do the trick here, or do we need to fire both events from `onBlur` for some reason?

pauln:
@miguelcobain I've had a bit of a play with this, and confirmed that my change was preventing the eps `deactivate` action from firing, which left the eps state as "active".  I've just pushed another commit which fixes this similarly to how `onFocus`/`activate` work together; note that `onBlur` doesn't call `this.publicAPI.actions.close(event)` as it's not necessary and prevents the user from choosing an item (the dropdown closes prematurely).

addon/components/paper-autocomplete.js
```diff
@@ -32,6 +32,13 @@ export default PowerSelect.extend({
         action(this.publicAPI, event);
       }
     },
+    onBlur(event) {
+      this.send('deactivate');
```
miguelcobain: We should use `sendAction` and not `send`

miguelcobain:
@pauln thanks for this.

@cibernox I wonder if such an action should be supported in core eps. The `onfocus` counterpart.

PR #506, opened 2016-09-28
Add super calls inside paper-input lifecycle hooks.

jordpo:
* ensures hooks higher up in mixinx will fire properly

this was causing an error when we were using the `paper-form` input and dynamically adding/removing its child components in the form based on the current form data.

### DIFF
--- a/addon/components/paper-input.js
+++ b/addon/components/paper-input.js
@@ -155,14 +155,12 @@ export default BaseFocusable.extend(ColorMixin, FlexMixin, ChildMixin, {
   },
 
   didRender() {
-    this._super(...arguments);
     this.growTextarea();
     // setValue below ensures that the input value is the same as this.value
     this.setValue(this.get('value'));
   },
 
   willClearRender() {
-    this._super(...arguments);
     this.sendAction('onValidityChange', false);
   },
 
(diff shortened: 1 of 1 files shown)


shoxter:
@miguelcobain I was also experiencing this. 

This should fix errors like ```Uncaught Error: Assertion Failed: calling set on destroyed object: <client@component:paper-input::ember1779>.isTouched = true```

I haven't tested his fix, but I suspect this fixed the issue.

miguelcobain:
There is a form related error that is failing. Mind checking that?

jordpo:
@miguelcobain will do later today

jordpo:
the only thing I can think of is the super call inside the `didRender` hook was screwing up the `onValidityChange` action from getting properly fired.

shoxter:
@jordpo @miguelcobain The `onValidityChange` was firing 3 times (as opposed to the 2 expected). This is what your failed buil reason was. The `assert.expect(2)` is the faulty part in my opinion. I would expect there to be 3 assertions (one for each input on render, and then a third when setting the errors array).

jordpo:
the default scenario fails if we update the assert.expect to 3. I wonder if we can just take the assert.expect out?

shoxter:
@jordpo I really have to wonder why the default scenario fails at 3. I can't think out why it would be anything but 3. If we remove the `assert.expect(n)` we'd essentially just have to remove the test.

Maybe @miguelcobain can rectify why the default build returns anything but 3 assertions?

miguelcobain:
@jordpo thanks for the PR. @shoxter was right. The action should be triggered three times. The test was wrong.

jordpo:
@miguelcobain 👍

PR #510, opened 2016-10-05
Fixed problem with paper-select-options in Ember 2.7+

ibarrick:
They added assertions that were failing for paper-select options. I was able to fix the issue by setting the attributeBindings and tagName to `undefined` and `''` respectively before calling init, rather than after. Now the assertions pass and the component works as expected.


miguelcobain:
Thanks for looking in to this, @ibarrick.

Issue #512, opened 2016-10-06
ember-css-transitions version v0.1.5 breaks ember paper v0.2.x


chbonser:
A few days ago ember-css-transitions published a breaking change in v0.1.5.  ember-paper specifies a dependency on this addon using `^` which resulted in picking up the newly published version.  The net effect is the following error when using paper-select (and maybe other components).

```
TypeError: Cannot read property 'addEndEventListener' of undefined
at exports.default._ember.default.Component.extend.hideWrapper (paper-menu-container-abstract.js:59)
```

I’ll submit a PR shortly (to the release-0-2 branch?) to pin the ember-css-transitions to the previously published v0.1.2 version.

miguelcobain:
I released version `0.2.15` with this fix. Thanks again.

PR #513, opened 2016-10-06
pinning the version of ember-css-transistions as v0.1.5 was a breaking change

chbonser:
v0.1.2 was the most recent non-breaking version published to npm.

This fixes https://github.com/miguelcobain/ember-paper/issues/512


miguelcobain:
Thanks a lot for this!

PR #514, opened 2016-10-07
paper-icon: Set width as well as height / font-size

pauln:
When specifying a size for paper-icon, only the height and font-size were being set - the width was always left at 24px.

### DIFF
--- a/addon/components/paper-icon.js
+++ b/addon/components/paper-icon.js
@@ -40,7 +40,7 @@ let PaperIconComponent = Component.extend(ColorMixin, {
     let size = this.get('size');
 
     if (size) {
-      return Str.htmlSafe(`height: ${size}px; font-size: ${size}px;`);
+      return Str.htmlSafe(`height: ${size}px; width: ${size}px; font-size: ${size}px;`);
     }
   })
 });


miguelcobain:
I'm not sure if we want fixed width icons all the time. Font awesome, for example, has an option to have fixed width icons.

I'd have to know more about the rationale of this.

pauln:
The problem is that they're currently fixed to 24px no matter what size you specify, which can result in other elements overlapping the icons (in the case of icons >24px) or a large gap between the icon and following content (for icons <24px).

The fixed width of 24px appears to be [coming from AM](https://github.com/angular/material/blob/master/src/components/icon/icon.scss), which sets both width and height to `$icon-size` on `md-icon` elements unless they have an `md-icon-font` attribute. in which case they become `width: auto`.  Unless I'm missing something, `paper-icon` never uses the `md-icon-font` attribute, so we always end up with the fixed width of 24px.

If we want to have `width: auto` rather than setting the width to equal the height/size, I guess we'd have to override the AM styles; if you'd rather take that approach, please let me know what the preferred way to override the AM styles for an e-p component is, and I'll prep a replacement PR.

miguelcobain:
@pauln I see. I think the correct way would be to use an attributeBinding to bind the icon name to `md-icon-font`.
Would you mind adding that attribute and a test for it?

pauln:
@miguelcobain Done - see #515.

PR #515, opened 2016-10-07
paper-icon: Set md-font-icon attribute

pauln:
Following discussion in #514, set md-font-icon attribute on paper-icon so that the icon doesn't end up with a fixed width of 24px no matter what size is specified.


miguelcobain:
Perfect. Thanks a lot.

PR #517, opened 2016-10-07
Legible paper-radio demo

xomaczar:


miguelcobain:
Thanks for this. Looks great!

We will need to buff up the docs very soon. 👍

PR #519, opened 2016-10-10
paper-autocomplete: don't highlight options by default

pauln:
Use ember-power-select 1.0.0-beta.23's new `defaultHighlighted` option to override its default highlighting (on open and filter/search) to match AM's behaviour - not highlighting anything.  Keyboard or mouse can be used to select an option from the dropdown.


miguelcobain:
Thanks. :+1:

Issue #420, opened 2016-07-08
Error watching angular-material-source dependency directory when using npm 3.x


tzellman:
I'm using npm 3.x and just started getting an error recently when building (recently, possibly due to caching).

I originally thought it came down to this change: 32fee2c21c6e6762d81e484887b4596f4cf9e462

> The Broccoli Plugin: [Funnel: AngularScssFunnel] failed with:
> Error: Attempting to watch missing directory: /my/path/node_modules/ember-paper/node_modules/angular-material-source/src
> at EventEmitter.Watcher_addWatchDir as addWatchDir

Basically, npm 3.x pulls the angular source down to a peer level to the ember-paper directory under node_modules, but does not put it under the sub node_modules directory within ember-paper, which is where the addon appears to be watching.

Switching to npm 2.15.8 causes my projects to build/work successfully.

jpaas:
Same here. Running npm 3.10.3.

jamesdixon:
For anyone else that runs into this, I had to remove the module from `package.json` and then run `ember install ember-paper`. Just scorching `node_modules` and then running `npm i` didn't work.

bdmac:
@jamesdixon that's fine for local but our deploy process isn't working either now. :-/

jerel:
I'm getting this same error when switching to `yarn`. Using ember-paper 1.0.0-alpha-4

miguelcobain:
We need to fix this logic: https://github.com/miguelcobain/ember-paper/blob/master/index.js#L119

miguelcobain:
Fixed by #521.

BobrImperator:
Heroku build would fail for ``npm 5.6.0`` specyfing ``npm 6.0.1`` did the trick.

PR #521, opened 2016-10-13
Update Angular SCSS import to work with yarn's flat node_modules

jerel:
I don't have a good way to test compatibility with all node and npm versions but this change should only be additive and does work for me with npm v3 and with yarn.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,6 +1,7 @@
 /* jshint node: true */
 'use strict';
 
+var fs = require('fs');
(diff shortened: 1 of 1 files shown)


miguelcobain:
@jerel this awesome. We had some problems in npm 3.
Thanks a lot.

This fixes #420.

v3ss0n:
thanks a lot @jerel  and @miguelcobain  thats awesome!

PR #522, opened 2016-10-13
Moved "hide-gt-sm" class in sidenav demo.

jerel:
The component `paper-sidenav-toggle` now has a `tagName` of `''` so adding a class to it as shown in the demo doesn't work and `Toggle` is visible on all screen sizes.


miguelcobain:
Yeah, makes sense.

These are probably remnants from the old toggle.

Thanks. 👍

PR #524, opened 2016-10-14
Autocomplete API change

ibarrick:
search text should now be tracked and initializeable. Typing after something is already selected clears selection.


ibarrick:
This should be ready now.

app/templates/components/paper-autocomplete.hbs
```diff
@@ -40,6 +40,7 @@
       onKeydown=(action "onKeydown")
       searchEnabled=(readonly searchEnabled)
       searchField=(readonly searchField)
+      searchText=searchText
```
xomaczar: Can we be explicit here and show that's searchText is one way with readonly helper

addon/components/paper-autocomplete-trigger.js
```diff
@@ -66,11 +74,14 @@ export default Component.extend({
     },
 
     handleInputLocal(e) {
+      // If something is already selected when the user types, it should clear selection
+      if (this.get('selected')) {
+        this.get('select').actions.select(null);
```
xomaczar: Does it auto search or we need to call actions.search

ibarrick: the onInput fires and calls search automatically.

app/templates/components/paper-autocomplete.hbs
ibarrick: just added commit to fix this, should be good now.

addon/components/paper-autocomplete-trigger.js
xomaczar: Great

xomaczar:
LGTM - let's merge it.
cc @miguelcobain

PR #525, opened 2016-10-18
Don't show dropdown if paper-autocomplete has a selected value and

xomaczar:
receives a focus.

### DIFF
--- a/addon/components/paper-autocomplete-trigger.js
+++ b/addon/components/paper-autocomplete-trigger.js
@@ -6,24 +6,30 @@ export default Component.extend({
   tagName: 'md-autocomplete-wrap',
   classNameBindings: ['noLabel:md-whiteframe-z1', 'select.isOpen:md-menu-showing'],
 
-  text: computed('selected', 'extra.labelPath', 'searchText', {
+  noLabel: computed.not('extra.label'),
+  _innerText: computed.oneWay('searchText'),
+
+  text: computed('selected', 'searchText', '_innerText', {
     get() {
-      if (this.get('selected')) {
+      let {
+        selected,
+        searchText,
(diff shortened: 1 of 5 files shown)


xomaczar:
@miguelcobain - should be all fixed.

miguelcobain:
@xomaczar here are my thoughts:
We shouldn't force the user to provide both `onSearchTextChange` **and** `onSelectionChange`. The reason is that the searchText or selection may not be relevant to the user at all.

- [ ] we should assert that either `onSearchTextChange` or `onSelectionChange` is present
- [ ] if `onSearchTextChange` is absent, the input text should be updated normally as you type. Implementing this will probably lead to an internal property that holds the current input text?
- [ ] if `onSelectionChange` is absent, don't do anything?
- [ ] in both cases, `searchText` or `selected` should **never** be updated directly, otherwise the default two-way nature of bindings would propagate changes up (violating DDAU, hence the reason for an internal property)
- [ ] I think we can rename `onSearchTextChange` to `onTextChange`

xomaczar:
- [x] assert that either onSearchTextChange or onSelectionChange is present
- [x]  if onSearchTextChange is absent, the input text should be updated normally as you type. Implementing this will probably lead to an internal property that holds the current input text?
- [x]  if onSelectionChange is absent, don't do anything?
- [x] in both cases, searchText or selected should never be updated directly, otherwise the default two-way nature of bindings would propagate changes up (violating DDAU, hence the reason for an internal property)
- [ ]  I think we can rename onSearchTextChange to onTextChange

xomaczar:
@miguelcobain ready for review

miguelcobain:
🎉 
very nice.

PR #350, opened 2016-04-28
Feature/toast

brandonjmckay:
Continuation of https://github.com/miguelcobain/ember-paper/pull/141

PR #141, opened 2015-07-31
Feature: Toasts & Snackbars

peec:
Demo: http://jensb.home.pkj.no/paper-toast/index.html#/toast

- Full featured toasts.
- Angular material scss only
- Actions up, data down driven
- Animations, Swipe and all event based features works.

PR #529, opened 2016-10-28
Feature/toast

pauln:
Continuation of PR #350 (itself a continuation of PR #141).  Rebased and updated to work on latest master, plus some minor fixes.


tests/unit/components/paper-toast-test.js
```diff
@@ -0,0 +1,35 @@
+import { moduleForComponent, test } from 'ember-qunit';
```
miguelcobain: This should be converted to an integration test.
This is the test that is failing (`needs: ['service:transition-events']` no longer makes sense in the new ember-css-transitions version we're using).

brandonjmckay:
May want to check out [ember-notify](https://github.com/aexmachina/ember-notify). I know I revised the original work, but that's no quite some time ago, could probably use a revisit to the implementation.

xomaczar:
I'm needing some toasts in my app - what's the general direction we want to go? Continue work based on this PR or use external addons and why? I'm willing to drive this to completion

miguelcobain:
@xomaczar thanks again for your interest. I actually have what I think to be a good plan to tackle toasts.

### Step 1

Make a `{{#paper-toast` component. This will be our basic building block. It would work pretty much like a `{{#paper-dialog` and it would be public in case people want to make more customised things in a toast. A basic example:

```hbs
{{#if isToastOpen}}
  {{#paper-toast onClose=(action (mut isToastOpen) false)}}
    Toast was shown successfully!
  {{/paper-toast}}
{{/if}}

{{#paper-button onClick=(action (mut isToastOpen) true)}}
  Open toast
{{/paper-button}}
```

API:
- `hideDelay` - default `3000` - number of ms before triggering the `onClose` action. Note that the action is responsible for closing the toast, DDAU style, pretty much like dialogs. we could support a special value like `false` to make a toast that never automatically closes itself.
- `parent` - default `paper-wormhole` - this is the id of the parent of the toast we will render into, again very similar to dialogs. I'm not 100% sure if we can render toasts on paper-wormhole as well, but I don't see why we can't.
- `position` - default `'bottom left'` - possible values are `'bottom left'`, `'bottom right'`, `'top left'`, `'top right'`
- `capsule` - default to `false` - Whether or not to add the `md-capsule` class to the toast to provide rounded corners

The component will probably yield some util components that hide the material markup underneath:

```hbs
{{#paper-toast onClose=(action (mut isToastOpen) false) as |toast|}}
  {{#toast.text}}
    Custom toast!
  {{/toast.text}}
  {{#toast.close-button}}
    Close
  {{/toast.close-button}}
  {{!--
    for maximum flexibility, we should also yield the close action.
    users may trigger it in any way they want
  --}}
  <button onclick=(action toast.closeAction)>Close using ember</button>
{{/paper-toast}}
```

Notes:
- the component would include the ColorMixin to support `accent=true`
- I believe we'll need to use ember-css-transitions for animating the toast
- We need this building block, and there's no reason we shouldn't make it public (that is, we should export it to the consuming app). It may allow more complex toasts the service won't. Speaking of that...

### Step 2

Make a service so we can show toasts more easily.
I'll lay out the API and the approach here soon.

xomaczar:
Thanks Miguel for details description - this smells like the first ember-paper RFC

miguelcobain:
closed in favor of https://github.com/miguelcobain/ember-paper/issues/706

PR #531, opened 2016-10-31
paper-select focus fix

ibarrick:
Should not try to call focusTarget.focus if there is no focusTarget (which is perfectly possible).


miguelcobain:
Thanks again. 👍

PR #532, opened 2016-10-31
validation for autocomplete and change to form

ibarrick:
Added contextual components to paper-form for select and autocomplete. Interesting issue with autocomplete is that there are two properties that we may want to watch for validation over but currently validation-mixin only supports a single validation property. For now I set it to do validation on searchText but that can be changed by the user by passing in validationProperty='selected'.

Validation messages for autocomplete will only show up for the floating label version for obvious reasons.

### DIFF
--- a/addon/components/paper-autocomplete.js
+++ b/addon/components/paper-autocomplete.js
@@ -28,8 +28,13 @@ export default PowerSelect.extend(ValidationMixin, ChildMixin, {
   onblur: computed.alias('onBlur'),
   onchange: null,
   oninput: null,
-  validationProperty: 'searchText',
-
+  validationProperty: computed('onSearchTextChange', 'onSelectionChange', function() {
+    if (this.get('onSearchTextChange')) {
+      return 'searchText';
+    } else {
+      return 'selected';
+    }
+  }),
   searchText: '',
   _onChangeNop: emberNop,
 
--- a/addon/components/paper-autocomplete.js
+++ b/addon/components/paper-autocomplete.js
@@ -3,6 +3,8 @@
  */
 import Ember from 'ember';
 import PowerSelect from 'ember-power-select/components/power-select';
+import ValidationMixin from 'ember-paper/mixins/validation-mixin';
+import ChildMixin from 'ember-paper/mixins/child-mixin';
 import { indexOfOption } from 'ember-power-select/utils/group-utils';
 
 const { assert, computed, inject, isNone, defineProperty, K: emberNop } = Ember;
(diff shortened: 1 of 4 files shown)


ibarrick:
This should be set to go now.

addon/components/paper-autocomplete.js
```diff
@@ -28,8 +28,13 @@ export default PowerSelect.extend(ValidationMixin, ChildMixin, {
   onblur: computed.alias('onBlur'),
   onchange: null,
   oninput: null,
-  validationProperty: 'searchText',
-
+  validationProperty: computed('onSearchTextChange', 'onSelectionChange', function() {
+    if (this.get('onSearchTextChange')) {
```
miguelcobain: With this test, we're using `searchText` if both `onSearchTextChange` and `onSelectionChange` are present, which is not what we want to.

miguelcobain:
@ibarrick should we add this to paper-autocomplete's template?
https://github.com/miguelcobain/ember-paper/blob/master/app/templates/components/paper-select.hbs#L51-L67

ibarrick:
@miguelcobain I don't think so, that should be taken care of by the paper-input component that the trigger renders if label is defined. We only had to do that in paper-select because it no longer extends paper-input.

miguelcobain:
🤔  hmmm

Why then does paper-complete need ValidationMixin in the first place?
It doesn't seem to do a lot of what it should. selected and searchtext are never validated, it seems to me. There is now way to render errors for those.

ibarrick:
I see what you're saying. I can pass `validationErrorMessages` down to the paper-input instead of `required` and `customValidations` and it should display the error messages as if they originated from the paper-input.

miguelcobain:
That makes sense.
Also, with that scenario, paper-autocomplete doesn't exactly need the validationmixin?

ibarrick:
In that scenario it would because the `validationErrorMessages` array is computed by the mixin and should be computed for `searchTerm` or `selected`, rather than the value of the input.

We could remove the validation mixin right now if we wanted but I think that would be a mistake since it is conceivable that a user may want to do more complicated validation on `selected`, such as computing the validation based on properties of `selected` other than what is rendered in the paper-input.

miguelcobain:
I see. So we would be essentially computing the errors on autocomplete and passing them to paper-input for it to render.
I wouldn't want paper-input to compute any kind of validation. `validationProperty=null` doesn't seem possible because we assert its presence.

ibarrick:
not passing in any validations should ensure that it doesn't compute any of it's own it should just render the messages it's given

miguelcobain:
Great. So it seems we have a path forward.
Unfortunately I've already merged this, so we need a new PR.

PR #533, opened 2016-10-31
VR: ignore offsetter parameter

ibarrick:
sometimes it is desirable to have the list items render outside of the virtual repeats offsetter. In particular, using VR in conjunction with tables (such as paper-data-table) requires the table to be rendered as a sibling of the offsetter to work correctly.

I'd also like to remove the virtual-each class from the paper-virtual-repeat, I'm not sure how to do that though since it's being added by the parent class and doesn't appear to be overridden by the paper-virtual-repeat implementation.


ibarrick:
@miguelcobain I don't imagine this change will effect many people since it is strictly an addendum to the api, could you take a quick look when you get the chance?

miguelcobain:
Will merge it as an "undocumented" API for now.
As you say, it won't affect people.

PR #536, opened 2016-10-31
Validation Messages fix

ibarrick:


SirZach:
any chance to bump this to a beta7 release for this PR? otherwise I need to use master in my package.json so tests will pass which is fine but not ideal

miguelcobain:
@SirZach `1.0.0-alpha.7` was released.

SirZach:
thanks!

PR #537, opened 2016-10-31
Gave paper-form its own section

shoxter:


miguelcobain:
This is a very important part of the documentation. paper-form isn't very evident in the current docs.

- [ ] mention that this works with paper-select and paper-autocomplete.
  - say it in text with `link-to`s to each page and use them in the example as well
- [ ] I think this should have its own section called "Forms", next to Typography. I think of "Components" as something more visual, i.e AM components ports, essentially. I intended to make this section a bit more "meta", like a guide of "how you make forms in ember-paper".
- [ ] delete the old paper-form section in the paper-input page (since paper-form is no longer related to just paper-input). That text looks like an interesting intro text to paper-form.

PR #541, opened 2016-11-01
allow touch devices to interact with autocomplete

ibarrick:
Currently, touch devices are unable to focus on the input in paper-autocomplete because ember-basic-dropdown does a preventDefault on the touch event that prevents the focus event from being called on the input. This corrects that issue.

### DIFF
--- a/addon/components/paper-autocomplete-trigger-container.js
+++ b/addon/components/paper-autocomplete-trigger-container.js
@@ -20,7 +20,18 @@ export default BasicTrigger.extend({
     }
     return tabindex;
   }),
-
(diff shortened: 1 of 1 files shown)


miguelcobain:
Can you clarify why paper-select doesn't need this?

ibarrick:
@miguelcobain Because there is no need to focus on any of the children of the trigger in the case of paper-select because there is no text input underneath the trigger that we're trying to access. In autocomplete we need to focus on the input in order to type, so it's problematic that the event is not allowed to go past the trigger container.

miguelcobain:
I see. Thank you.

PR #547, opened 2016-11-03
ember-power-select compat - fix paper menu positioning (fixes #546)

### LINKED ISSUE #546: paper-menu positioning off
The positioning of the paper-menu content doesn't attach to the element and instead always shown in the upper left-hand corner of the page. I've tracked this down to the `top` and `left` properties having an  extra `px` attached to the properties. It appears that `ember-basic-dropdown` already adds the unit and with that, it's being added twice resulting in the property becoming invalid.

Note that I'm running against master.

I'll submit a PR with a fix.

Relevant code from `ember-basic-dropdown`: https://github.com/cibernox/ember-basic-dropdown/blob/4295170a2e53277b6b0af4503b6e2168a7d79d25/addon/components/basic-dropdown.js#L180

jamesdixon:
This resolves an issue with paper-menu positioning being off due to the `px` unit being added twice to the `top` and `left` properties.


miguelcobain:
I am aware of this change in ember-basic-dropdown, but I thought we wouldn't pull that version yet, according to the package.json we're using.
What is the version of ember-basic-dropdown you're using?

jamesdixon:
@miguelcobain it appears to be 0.16.4

miguelcobain:
@jamesdixon The functionality you're describing landed in 0.17.0: https://github.com/cibernox/ember-basic-dropdown/blob/master/CHANGELOG.md#0170

:thinking:

jamesdixon:
@miguelcobain aye, you're correct. Not sure what's going on. In `node_modules/ember-basic-dropdown`, I see 0.16.4. Going to try blowing things away and see what I end up with. I'll keep you posted

jamesdixon:
Not sure what's going on. Blew away `node_modules` and I'm still seeing 0.16.4 😕

miguelcobain:
@jamesdixon Do you have any `ember-power-select` entry in your `package.json`, i.e did you install `ember-power-select` on yourself on your application?

cibernox:
Maybe some other dependency (namely EPS) is pulling a newer version of ember-basic-dropdown, and since npm's idea of duplicating dependencies when there is a mismatch is stupid AF, perhaps you have two different versions installed, and there is no way of knowing which one is going to be included.

jamesdixon:
@miguelcobain @cibernox I do have EPS installed (latest version). I think you're right; because of NPM, the wrong version is being pulled.

cibernox:
Either way, this fix is good 👍 
You should also bump the version of ember-basic-dropdown in the package.json to ^0.17

miguelcobain:
Other people had issues when using ember-paper + ember-power-select.
  1. styles. `paper-select` includes classes from ember-power-select. It means that if people include the ember-power-select styles, they will collide.
  2. javascript. If people include a different version of ember-power-select, that one will be used. I don't see what can be done about this, besides trying to keep ember-paper updated.

@cibernox is there any way we can solve the first problem?
Would we need to manually prevent ember-power-select from applying its classes everywhere?

cibernox:
1. If people use ember-paper they probably want to use the select in ember-paper, right? If so people using ember-paper shouldn't include EPS styles.

For sass users that is easy, because the inclusion of the styles is explicit (@import 'ember-power-select'). For non-sass users, they can opt out to ANY styles by doing:

```js
var app = new EmberApp(defaults, {
  'ember-power-select': {
    theme: false
  }
});
```

miguelcobain:
> If people use ember-paper they probably want to use the select in ember-paper, right? If so people using ember-paper shouldn't include EPS styles.

Supposedly, yes. But it looks like some people might want to use ember-power-select **and** paper-select separately.

cibernox:
In that case I know an "easy" fix we can do in Ember Power Select, but will only work for `sass` users. Which it might be good enough.

Basically right now in sass, you import the default styles by doing `@import 'ember-power-select';`
If that file instead of just importing the styles imported a mixin instead, the users would be able to do:
```scss
@import 'ember-power-select';
.eps-select-trigger {
  @include ember-power-select-trigger();
}
.eps-select-dropdown {
  @include ember-power-select-dropdown();
}
```

so the styles are scoped to that class only. If someone wants the exact existing behaviour where styles are global, they can import and use the mixing without scoping it.

```scss
@import 'ember-power-select';
.ember-power-select-trigger {
  @include ember-power-select-trigger();
}
.ember-power-select-dropdown {
  @include ember-power-select-dropdown();
}
```

This solution means that if users want EPS styles to coexist, they need to add the special classes to all their selects:

```hbs
{{#power-select options=... onchange=... triggerClass="eps-select-trigger" dropdownClass="eps-select-dropdown" as |opt|}}
 ...
```
This can be tiring but people can just an initializer to reopen the default trigger/dropdown and avoid adding this class all the time.

This is the solution I can think about, and it is something I was considering anyway so people can opt-in to styles in a more controlled way that just importing a file.

jamesdixon:
FWIW, I've opted to use EPS over paper-select (for the time being) due to more options. Plus, it was already embedded throughout my codebase and haven't had time to change it :)

miguelcobain:
@jamesdixon what options do you think are missing in paper-select? Multiple selection?

jamesdixon:
@miguelcobain multiple-select was the option that kept me from the switch. I have a couple of filtering interfaces that utilizes the multi-select version of EPS.

Glennvd:
+1 for this PR

miguelcobain:
This should be fixed on master. We updated to eps 1.0. Thanks!

PR #551, opened 2016-11-08
[paper-input] preventing exception when typing with mobile keyboard

Juraci:
I got the following exception on paper-input when typing with the Android keyboard:

 [paper-input error](https://drive.google.com/file/d/0B_CrumEitcQDREFQaVRLc1UtczA/view?usp=sharing)
Cannot read property 'val' of undefined.

Steps to reproduce:

requirements:
Android smartphone with google chrome

1. Open Chrome dev tools and setup it for [remote debugging](https://developers.google.com/web/tools/chrome-devtools/remote-debugging/)
2. On the phone open this page [dialog example](http://miguelcobain.github.io/ember-paper/release-1/#/components/dialog)
3. In the prompt dialog example type some text with the phone keyboard (has to be with the phone keyboard)
4. Click on confirm action on the dialogue 

Expected:  Dialog closes
Actual: Dialog remains opened  with the error linked above
 
The fix in the pull request prevents this exception while maintaining the current behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ Contributions and pull requests are always welcome. Contributors may often be fo
 ### master
 - [#542](https://github.com/miguelcobain/ember-paper/pull/542) paper-form now yields a submit button
 - [#549](https://github.com/miguelcobain/ember-paper/pull/549) Allow `title` attribute in paper-button
+- [#551](https://github.com/miguelcobain/ember-paper/pull/551) avoids calling setValue when the component is destroyed
 
 ### 1.0.0-alpha.7 (November 2, 2016)
 - [#531](https://github.com/miguelcobain/ember-paper/pull/531) paper-select focus fix - doesn't call focusTarget.focus if there is no focusTarget (which is perfectly possible)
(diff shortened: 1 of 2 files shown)


waldemarnt:
👍

miguelcobain:
@Juraci I wonder why this fix is needed. I'm generally opposed to merging fixes without knowing what is the root cause of it. Do you have any idea?

Juraci:
@miguelcobain I did not have the time to dig deeper into why this is happening on the mobile only. I do suspect that the setValue method is being called after a dialog action has been made, but I do not know why this happens when using the mobile keybord only.

A fix is definitely needed because the prompt dialog is currently unusable in mobile platforms

miguelcobain:
@Juraci I think I know the root cause of the problem.
We're using `run.next` with `setValue` here: https://github.com/miguelcobain/ember-paper/blob/master/addon/components/paper-input.js#L144-L146

What I suspect is happening is that, by the time `setValue` runs, the template is already destroyed (especially on android due to low performance, I assume).
The proper fix would be to test if the component is destroyed, before calling `setValue`. We can test that checking `this.isDestroyed` value.

The code would become somthing like:
```js
run.next(() => {
  if (this.isDestroyed) { return; }
  this.setValue(this.get('value'));
});
```

Could you please test this, and if it solves the issue, update the PR?

Juraci:
@miguelcobain, sure I will do that.

Juraci:
@miguelcobain you are correct, this solves the problem, just tested it here.

miguelcobain:
@Juraci Great. While you're at it, could you please add an entry to the changelog?

miguelcobain:
Thanks a lot for this!

PR #554, opened 2016-11-09
[DEMO]: remove form from components

BennyAlex:
there is no form route so I removed it

### DIFF
--- a/tests/dummy/app/templates/application.hbs
+++ b/tests/dummy/app/templates/application.hbs
@@ -23,7 +23,6 @@
             {{#submenu-item active=(is-active "demo.checkbox" currentRouteName) onClick=(transition-to "demo.checkbox")}}Checkbox{{/submenu-item}}
             {{#submenu-item active=(is-active "demo.dialog" currentRouteName) onClick=(transition-to "demo.dialog")}}Dialog{{/submenu-item}}
             {{#submenu-item active=(is-active "demo.divider" currentRouteName) onClick=(transition-to "demo.divider")}}Divider{{/submenu-item}}
-            {{#submenu-item active=(is-active "demo.form" currentRouteName) onClick=(transition-to "demo.form")}}Form{{/submenu-item}}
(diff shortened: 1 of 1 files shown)


miguelcobain:
We want it to be a Forms page for ember-paper. So, we want the ending S

BennyAlex:
But where it is the plural? 
And do you want to have a Forms Demo and a Forms Page?

tests/dummy/app/templates/application.hbs
```diff
@@ -41,7 +40,7 @@
              {{#submenu-item active=(is-active "demo.virtual-repeat" currentRouteName) onClick=(transition-to "demo.virtual-repeat")}}Virtual Repeat{{/submenu-item}}
           {{/liquid-if}}
         </div>
-        {{#menu-item active=(is-active "forms" currentRouteName) onClick=(transition-to "forms")}}Forms{{/menu-item}}
+        {{#menu-item active=(is-active "forms" currentRouteName) onClick=(transition-to "forms")}}Form{{/menu-item}}
```
miguelcobain: You've replaced `Forms` with `Form`. What is the reason for that?
We thought of `Forms` to be a general purpose sectionabout "Forms" in ember-paper, not just about paper-form.

miguelcobain:
Thanks!

BennyAlex:
I just renamed it because the other navbar item names are singular

BennyAlex:
Theres is also a bug on the forms site, showing "comonents >" before "form"
image

miguelcobain:
Nice catch. :)
Can you PR that as well, please?

BennyAlex:
As far i can see the isDemo attribute is already false. 
So probably ypu have to update the demo page?

miguelcobain:
That's right. :face_with_head_bandage:

BennyAlex:
I dont know exactly how the demo page works, but would it be possible to use the master branch for the page? so you dont need to update it manually

DanChadwick:
@BennyAlex Seen "Building and deploying the dummy app to github" in [contributing.md](https://github.com/miguelcobain/ember-paper/blob/master/CONTRIBUTING.md)

BennyAlex:
@DanChadwick I dont want to build or deploy this, in the end I will break something. I just thought if it is possible to do this automatic, based on the master branch

PR #556, opened 2016-11-10
[DEMO] width should be height

BennyAlex:
the max-width is set to 100% so max-height is the the height of the items in the container, not the width


miguelcobain:
Thanks!

PR #559, opened 2016-11-13
Fix old/missing sidenav attributes

knownasilya:
Not really sure, but this looks right.

### DIFF
--- a/app/templates/components/paper-sidenav.hbs
+++ b/app/templates/components/paper-sidenav.hbs
@@ -1,6 +1,6 @@
 {{#unless closed}}
   {{paper-backdrop
-    locked-open=isLockedOpen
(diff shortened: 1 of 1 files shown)


knownasilya:
On another note, we should probably use ember-responsive or just have the user use it instead of specifying a string e.g. `lockedOpen=media.isMobile` vs `lockedOpen='gt-sm'` this way you can check if it should be open and not show a toggle button.

miguelcobain:
Thanks a lot for this!

We already support boolean type `lockedOpen`: https://github.com/miguelcobain/ember-paper/blob/master/addon/components/paper-sidenav-inner.js#L63-L68

Does this match the functionality you described?

PR #562, opened 2016-11-15
Close autocomplete drop-down on tabKey

xomaczar:
To reproduce:
1. In e-p current demo (v1.0.0-alpha.8), click on any autocomplete, then hit 'tab' key w/o selecting/highlighting any items
2. The autocomplete drop-down stays open

Fix:
- When tab is invoked, just delegate to e-p-s component, which closes the drop-down by default.


miguelcobain:
👍

xomaczar:
thx

PR #563, opened 2016-11-15
Make autocomplete similar to paper-input (required star & passThru hash)

xomaczar:
1. Support required attribute on autocomplete similar to paper-input (shown
with a start (*))
2. Allow passThru hash to passed in.


miguelcobain:
Arigato.

PR #564, opened 2016-11-17
Error after removal of a paper item from View

A-Helberg:
When an action on a sub-element of a paper-item causes the item to be removed from the view, an error is caused because of trying to set mouseActive properties on the destroyed component.

### DIFF
--- a/tests/integration/components/paper-item-test.js
+++ b/tests/integration/components/paper-item-test.js
@@ -176,3 +176,32 @@ test('Item checkbox with secondary action and primary action dont bubble primary
     assert.notOk(this.get('secondaryValue'));
   });
 });
+
+test('An Item that is removed from the view should not cause errors in the runLater loop', function(assert) {
+  assert.expect(0);
+
+  this.set('items', [1]);
+  this.set('deleteItem', () => {
+    this.set('items', []);
+  });
+
+  this.render(hbs`
+    {{#paper-list}}
+      {{#each items as |item|}}
+        {{#paper-item}}
+          {{item}}
+          {{#paper-button class="md-secondary" iconButton=true onClick=(action deleteItem)}}
+            {{paper-icon "message"}}
+          {{/paper-button}}
+        {{/paper-item}}
+      {{/each}}
+    {{/paper-list}}
+  `);
+
+  let button = this.$('.md-secondary');
+
+  wait().then(() => {
+    button.click();
+    button.trigger('mousedown');
+  });
+});


A-Helberg:
I also noticed that the same error occurs when a click on the item causes the page to navigate away from the page where the list was on (subsequently causing the list to be removed from the view and destroyed)

addon/components/paper-item.js
```diff
@@ -82,10 +82,12 @@ export default Component.extend(RippleMixin, ProxyMixin, {
         let el = component.$();
         set(this, 'mouseActive', false);
         el.on('mousedown', ()=> {
-          set(this, 'mouseActive', true);
-          run.later(()=> {
-            set(this, 'mouseActive', false);
-          }, 100);
+          if (!this.get('isDestroyed')) {
```
arschmitz: You can't just check this here. You need to check inside the run later or the error will still occur if the action is async which is most cases.

arschmitz: I believe you will also want to check `isDestroying` and should probably be consistent with the rest of the file and use `get` not `this.get`

A-Helberg: @arschmitz Good catch! I'll update it shortly.

addon/mixins/ripple-mixin.js
```diff
@@ -136,24 +136,27 @@ export default Mixin.create({
...
-    // When jQuery is loaded, we have to get the original event
-    if (event.hasOwnProperty('originalEvent')) {
-      event = event.originalEvent;
-    }
-    this.mousedown = true;
-    if (this.get('center')) {
-      this.createRipple(this._container.prop('clientWidth') / 2, this._container.prop('clientWidth') / 2);
-    } else {
+    // We shouldn't do anything if the view is destroyed
+    if (!this.get('isDestroyed')) {
```
arschmitz: i believe you will also want to check `isDestroying` here as well

addon/components/paper-item.js
```diff
@@ -82,10 +82,14 @@ export default Component.extend(RippleMixin, ProxyMixin, {
         let el = component.$();
         set(this, 'mouseActive', false);
         el.on('mousedown', ()=> {
-          set(this, 'mouseActive', true);
-          run.later(()=> {
-            set(this, 'mouseActive', false);
-          }, 100);
+          if (!get(component, 'isDestroyed') && !get(component, 'isDestroying')) {
```
arschmitz: rather than component i think this should just be `this`

A-Helberg: Now that you mention it, @xomaczar @miguelcobain Is it the intended behavior that the ripple extend from the clicked component, across the entire paper item? 

In the case of clicking on a button it doesn't seem to make much sense?

xomaczar: Why not? clicking on paper-item should create a ripple effect according to the md spec & AM impl?

A-Helberg: So I did a bit of research and I couldn't come to a conclusive answer. The material design spec only states "Ink ripples confirm user input by immediately expanding outward from the point of touch"

On http://inbox.google.com, as well as the android and iOS apps, they don't use any ripple effect, anywhere. 

On Android, in the settings menu, the ripple effect does expand the entire equivalent of a `paper-item` but only if the right item is a toggle, such as the wifi menu option.

In the Sheets app, clicking on a card does produce a ripple effect, but clicking on the `more vert` icon in the card, only ripples the icon, and not the entire card.

I will fix the PR to any way you are happy. I guess there are two cases, a toggle makes sense to ripple the entire `paper-item` but a delete and a `vert more` does not.

What do you guys, think?

miguelcobain:
@A-Helberg I refactored lists in ember-paper.
Could you please test if the problem persists in master. In any case, this test would be welcome.

A-Helberg:
@miguelcobain The test and manual testing passes 😄 
Are you happy for me to have this PR only include the test?

miguelcobain:
Yes please. Lots of tests aren't enough.

A-Helberg:
Mmm it didn't fail locally, I'll try and find the cause/solution to the 
`writeDeps` issue

PR #566, opened 2016-11-17
Autocomplete should support ObjectProxy as selected value

### LINKED ISSUE #565: Autocomplete clears search text if the selected value is an ember-data belongsTo relationship
To reproduce:
1. Bind a selected value to a resolved DS.PromiseObject (a belongsTo on some model)
2. Place the focus on autocomplete trigger then backspace once (or however many times you want)
3. Notice that the searchText has been completely cleared out

xomaczar:
Closes #565

### DIFF
--- a/addon/components/paper-autocomplete-trigger.js
+++ b/addon/components/paper-autocomplete-trigger.js
@@ -17,12 +17,10 @@ export default Component.extend({
         _innerText
       } = this.getProperties('selected', 'searchText', '_innerText');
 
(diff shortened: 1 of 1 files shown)


miguelcobain:
Did you solve this problem meanwhile? I think ObjectProxy should be transparent and we shouldn't need to care about them.
Also, at most, this would be a change upstream in ember-power-select. I'm sure eps has tests to cover ember-data relationships as it seems to be your original problem in #565

PR #567, opened 2016-11-18
Autocomplete highlight non Text

A-Helberg:
When the label in a `paper-autocomplete-highlight` points to a property that is not Text the call to `replace` on  the property will fail.

An example use case is having an autocomplete of some sort of number. eg. Product codes, employee numbers...


xomaczar:
Thx. LGTM

miguelcobain:
Thanks a lot!

Issue #460, opened 2016-07-30
submitting paper-form with enter key


jpaas:
I just spent a lot of hours trying to make my paper-form component using 1.0.0.alpha.3 submit with the enter key and in the end I found the magical incantation, but I'm left wondering if maybe paper-form should make this easier, or at least document it.

Here is the only way I could make it work:

``` handlebars
  {{#paper-form onSubmit="submit" as |form|}}
    <form {{action form.onSubmit on="submit"}}>
        {{paper-input label="Email" onChange=(action (mut identification)) value=identification}}
        {{paper-input label="Password" onChange=(action (mut password)) type="password" value=password}}
        {{#paper-button disabled=form.isInvalid type="submit"}}Signin{{/paper-button}}
    </form>
  {{/paper-form}}
```

BennyAlex:
this causes that the action 'submit' will fired two times

miguelcobain:
```hbs
<form onsubmit={{action form.onSubmit}}>
```
should also work.

I'm open to suggestions on how this could be made easier.
paper-form is tagless, which makes it difficult to listen for key events. Maybe we should make it have a tag and listen for enter key presses?

jpaas:
It would be nice to have enter key submission "just work" by default. If that means generating a form tag, is there any downside?

jpaas:
If you're unsure, then just maybe just add a section to the docs?

miguelcobain:
I think the benefits outweigh the cons (if any). So, I could take a PR for this + tests. :)

jpaas:
@miguelcobain Sorry but I don't think I'll have time. I've moved onto a new project.

miguelcobain:
@shoxter what is your opinion on this?

shoxter:
@miguelcobain I think it's a good idea.

What if we add an action to the form that looks for the enter key being pressed (the event is passed in) and set the input(s) the user wants to have access to it to having the submit action (i.e. `form.submit`)?

miguelcobain:
@shoxter another idea would be to use `tagName: 'form'` and then use the `onSubmit` event.
Is there any harm that could come from this?

`onSubmit` triggers on enter keypresses and on button submits.

shoxter:
@miguelcobain The primary issue I foresee as being problematic is styling.

For example, in my personal application I wrap entire `paper-dialog`s in `paper-form` for validation purposes.

It's very flexible without a tag. This goes away when we add one.

[...]

PR #568, opened 2016-11-21
automatically set the type of submit-button to submit

panthony:
Specify the type of paper-form button as submit so that the form can be submitted using enter key

See #460


miguelcobain:
Thanks a lot!

PR #570, opened 2016-11-23
remove destructuring from paper-chips

ibarrick:


miguelcobain:
Thanks for the quick fix, @ibarrick.

PR #571, opened 2016-11-24
can add CSS classes to paper dialog container

panthony:
Add the capability for the user to specify CSS classes to the paper dialog container if necessary.

### DIFF
--- a/app/templates/components/paper-dialog.hbs
+++ b/app/templates/components/paper-dialog.hbs
@@ -5,8 +5,9 @@
     fixed=(unless parent true)
     class="md-dialog-backdrop"
     onClick=(action "outsideClicked")}}
-  {{#paper-dialog-container outsideClicked=(action "outsideClicked")}}
+  {{#paper-dialog-container class=(readonly dialogContainerClass) outsideClicked=(action "outsideClicked")}}
     {{#paper-dialog-inner
+      class=(readonly class)
       origin=origin
       defaultedParent=defaultedParent
       defaultedOpenFrom=defaultedOpenFrom
--- a/tests/integration/components/paper-dialog-test.js
+++ b/tests/integration/components/paper-dialog-test.js
@@ -252,3 +252,23 @@ test('opening gives focus', function(assert) {
   });
 
 });
+
+test('can specify dialog container classes', function(assert) {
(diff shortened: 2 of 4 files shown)


app/templates/components/paper-dialog.hbs
```diff
@@ -5,7 +5,7 @@
     fixed=(unless parent true)
     class="md-dialog-backdrop"
     onClick=(action "outsideClicked")}}
-  {{#paper-dialog-container outsideClicked=(action "outsideClicked")}}
+  {{#paper-dialog-container class=(readonly dialogContainerClasses) outsideClicked=(action "outsideClicked")}}
```
miguelcobain: I think that `class` would be a better API than `dialogContainerClasses`, no?
Since `paper-dialog` is a tagless component, we can just transport the passed in classes to `paper-dialog-container`.

tests/integration/components/paper-dialog-test.js
```diff
@@ -252,3 +252,13 @@ test('opening gives focus', function(assert) {
   });
 
 });
+
+test('can specify dialog container classes', function(assert) {
+  this.render(hbs`
+    <div id="paper-wormhole"></div>
+    {{paper-dialog dialogContainerClasses='flex-50 my-dialog-container'}}
```
miguelcobain: For consistency project-wide, please use double quotes in everything template related.

app/templates/components/paper-dialog.hbs
panthony: I named it `dialogContainerClasses` for two reasons:
 - consistency with other components (ex: `paper-select`: `class=(readonly concatenatedTriggerClasses)`)
 - avoid collision because there is multiple component we can pass class to (container, inner, ..)

miguelcobain: IMO:
  - `class` for `paper-dialog-inner`, since this is the dialog itself
  - `dialogContainerClass` for `paper-dialog-container`

paper-select component renders two things, a trigger and a dropdown. That's why we need to be more specific.
In this case this is just a "proxy" to render a single component, so there shouldn't be any confusion.

miguelcobain:
Perfect PR. 💯

PR #579, opened 2016-11-29
remove overflow:hidden directive on body tag that breaks test app

panthony:
Because the test app share its css with the dummy app. It had the 'overflow:hidden' css property which prevent the user from scrolling.

This property seems unnecessary as it does not appear* to affect the dummy app in anyway.

* tested on chrome only


miguelcobain:
👍

miguelcobain:
Didn't break anything here as well.

PR #580, opened 2016-11-29
Paper-autocomplete options must only be shown when interacted with

xomaczar:
Fixes a bug where clicking on paper-select component would cause paper-autocomplete w/o a label to also display its own dropdown options.


xomaczar:
^ @miguelcobain

miguelcobain:
Thanks a lot @xomaczar. Sorry for the delay.

PR #581, opened 2016-11-30
Add `openFrom` to paper-dialog docs

joukevandermaas:
Also add default value for `clickOutsideToClose`.

I can also add an example with a different `openFrom` and `closeTo` value if you feel that it's valuable.


miguelcobain:
@joukevandermaas such an example would be great.

miguelcobain:
Perfect!

PR #582, opened 2016-12-02
Update Ember, Ember CLI and Ember Data to 2.10

joukevandermaas:
Also includes `ember init`, which changed the ember try config slightly. Let me know what you think.

I tested the dummy app using 2.9 and 2.10 and found no differences. 2.10 felt a little faster on pages with a lot of components (like the input demo), but that's subjective.

### File sizes of the app bundle in a production build
|gzipped|size 2.9 | size 2.10 | reduction |
|---|---|---|---|
no |988.15 KB |465.31 KB | 53%
yes |103.81 |64.74 KB | 38%

Probably slightly skewed because the dummy app almost entirely consists of templates.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,7 +14,7 @@
 /connect.lock
 /coverage/*
 /libpeerconnection.log
-npm-debug.log
+npm-debug.log*
 testem.log
 
 # editors
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,11 +7,12 @@ sudo: false
 
 cache:
   directories:
-    - node_modules
+    - $HOME/.npm
+    - $HOME/.cache # includes bowers cache
 
 env:
-  - EMBER_TRY_SCENARIO=default
-  - EMBER_TRY_SCENARIO=ember-1.13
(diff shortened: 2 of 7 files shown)


xomaczar:
Awesome Thx

miguelcobain:
Tests are failing. :(

joukevandermaas:
I'll see if I can fix it later today. It looks like it's the new ember 2.4 LTS try scenario that's failing.

joukevandermaas:
The only tests that are failing are related to paper-dialog and paper-autocomplete. In both cases the tests run fine when run individually, but not when run in sequence. 

For both, the tests time out due to a `wait()` that never resolves. This is why autocomplete has more failing tests; more tests use `wait()`. I couldn't exactly figure out why this happened in the time that I had today. I'll probably have to dive into the implementation of these two components to figure out what exactly is going wrong.

If you want to merge this PR already, I can add the 2.4 LTS scenario to allowed failures.

miguelcobain:
prior to ember 2.6 there was a bug with `run.later` with a `NaN` timeout.
I made sure that at least we set `0` and it fixed the tests.

Thanks for this!

PR #585, opened 2016-12-07
Expose and rename constants service to 'paper-contants'

panthony:
The goal of this PR is to:

- Let the client import this service if required
- Avoid clash between the `ember-paper` constants service and the client own `constants` service (if any)

Use case:

I use `ember-paper` in my app and I wanted a `constants` service. I could not create one because it would override the one from `ember-paper` and I could not extend the one from `ember-paper` because it was not exposed.


miguelcobain:
@panthony I planned for a long to convert his service into a vanilla es6 module.
There is no need for this to be a service, I think.
It could be a module in `utils`.

panthony:
Ok, shall I close this PR then?

No need to leave it open if it won't get merge.

miguelcobain:
@panthony I think if we're going to refactor this, it should be to its final form, yes.
Migrating this to a module shouldn't be anything more difficult than what you've done.

PR #588, opened 2016-12-14
Blockless autocomplete

xomaczar:


miguelcobain:
I think I see the point of this. We have to pass in `labelPath` anyway, so a blockless version might make sense.
I wish we could get rid of `labelPath`, but I don't think we can because of the input.

xomaczar:
Agree, labelPath is only needed when binding to the list of objects. The blockless version should cover the 80% of all the use cases.

xomaczar:
@miguelcobain  ^

xomaczar:
Ping @miguelcobain

miguelcobain:
I went ahead and added these changes myself.
Sorry for taking 2 years for something that makes total sense and is easy to review.

PR #593, opened 2016-12-22
Allow external validations to return array or string

lukaiser:
Some external validators just provide a error string.
Allowing them to pass a string doesn’t seem have any side effects.
Therefor I propose to add this option. As it would make my life way easier.

An other option would be to add an 'error' property for strings and not use the 'errors' property.


lukaiser:
It finally passes. Sorry I missed that.

PR #595, opened 2016-12-23
Update ember-cli-sass to v6 (node-sass 4.1.0)

xomaczar:


xomaczar:
This should allow us to run in twiddle

v3ss0n:
wow nice we can twiddle with ember-paper now ??

miguelcobain:
Would be great to do so!

miguelcobain:
We're not using any sass from eps or ebd, but I'm not sure if those need to be updated as well.

xomaczar:
I know - maybe PRs to both

PR #599, opened 2016-12-23
Grid fix

sl249:
Removed all tiles with the state of _destroying to fix issue when rendering new tiles.


miguelcobain:
@sl249 Can you describe the issue? Steps to reproduce, etc?

PR #600, opened 2016-12-27
search fix

ibarrick:
When using paper-select with async search and no options specified, the menu will reposition itself when the loading spinner shows. This makes it look janky to do async searches so we should ignore the spinner change.


miguelcobain:
Thanks a lot. 👍

PR #608, opened 2017-01-04
[WIP] Datepicker component

eduardoweiland:
This is work in progress, but the basic functionality is working. I am sending this to get some feedback.

TODO:
- [ ] Add tests
- [ ] Allow changing date by typing in the input
- [ ] Allow customization of the trigger
- [ ] Allow the user to clear the selected date


miguelcobain:
I'm sure this involved a lot of effort and I thank you, but we were more inclined to build this feature extending ember-power-calendar. Have you considered this?

eduardoweiland:
I didn't know about this addon. But from what I could see, ember-power-calendar does not have an infinite scroll mode like Angular Material. Anyway, I will look into it and see what we can do.

miguelcobain:
To be honest, the Angular Material implementation looks completely off the spec: https://material.io/guidelines/components/pickers.html#pickers-date-pickers

miguelcobain:
@eduardoweiland template files are now on the `addon` folder.

addon/components/paper-calendar-days.js
```diff
@@ -0,0 +1,10 @@
+import Ember from 'ember';
+import PowerCalendarDays from 'ember-power-calendar/components/power-calendar/days';
+
+const { computed } = Ember;
+
+export default PowerCalendarDays.extend({
+  weekdaysLetters: computed('weekdaysNames', function() {
```
miguelcobain: can replace with

```js
weekdaysLetters: computed.map('weekdaysNames', function(name) {
  return name.charAt(0);
})
```

eduardoweiland:
Sorry @miguelcobain, I didn't see your edited comment with the Angular Material discussions around this problem. I have reworked the design of the calendar to match the one from the material website, but I haven't thought about how to open it.

I think this new design fits better with power-calendar, by not using an infinite scroll to navigate the months. Maybe we can make it work like angular/material#5971, opening it inside a dropdown instead of a modal?

I am trying to isolate the date picker from the trigger, so it could be opened by a text field, a select, a button, or anything else. What do you think?

eduardoweiland:
Sorry @miguelcobain, I didn't see your edited comment with the Angular Material discussions around this problem. I have reworked the design of the calendar to match the one from the material website, but I haven't thought about how to open it.

I think this new design fits better with power-calendar, by not using an infinite scroll to navigate the months. Maybe we can make it work like angular/material#5971, opening it inside a dropdown instead of a modal?

I am trying to isolate the date picker from the trigger, so it could be opened by a text field, a select, a button, or anything else. What do you think?

miguelcobain:
@eduardoweiland I think that's probably a bit against the spec.
What I'd prefer would be to reuse the AM styles, but to not use infinite scrolling if it makes things easier. What do you think?

eduardoweiland:
I will be back in office on Monday, then I will see what I can do. I think we can reuse the styles from AM for the input and dropdown, as I've already done in the first version. But for the calendar I think we need to rewrite, since power-calendar uses divs with buttons and AM uses tables (most of the rewrite is already done).

eduardoweiland:
A quick mockup of what this would look like:

paper-datepicker

miguelcobain:
@eduardoweiland looks nice!

miguelcobain:
@eduardoweiland we're using DockYard's code styling on this project, and that requires hbs to use double quotes instead of single quotes everywhere.

Can you please rebase/merge? I'd like to see the progress. Looks good!

eduardoweiland:
Hi @miguelcobain, unfortunately we decided not to use ember-paper in our project, so we will not invest more time on this component. Maybe, when we have some free time we can get back to work on this, or someone else can continue working on it.

miguelcobain:
@eduardoweiland sorry to hear that.

Can you tell us the reason for that decision, for educational purposes?
Maybe we can improve on the adoption story, and it could help us to better prioritise work.

eduardoweiland:
Sure @miguelcobain. It was mostly by performance reasons. The select component is very slow to open, even with just a few items, and our use case requires fast user input.

We decided to use [Semantic UI](https://github.com/Semantic-Org/Semantic-UI-Ember). It's not fully Ember, but it fulfills our needs.

v3ss0n:
can you point out some performance analysis so we can benchmark and improve?

v3ss0n:
for us we dumped semantic-ui because both its UI framework and Ember Addon is pretty much dead, not longer actively maintained

eduardoweiland:
@v3ss0n We haven't done any deep performance analysis. We have simply put a paper-select and an ui-dropdown on screen, with the same 100 options each, and tried to click on them. ui-dropdown opened instantly, while paper-select took about 2 seconds to open.

xomaczar:
100 options? I don't think paper-select uses virtual-repeat - try paper-autocomplete. Do you really think a user will scroll through so many options to find whatever he/she is looking for? Autocomplete is meant to address this use case, however we can make ep select use virtual-repeat which will speed it up substantially.

miguelcobain:
paper-select should be more or less as slow as ember-power-select is.

v3ss0n:
@eduardoweiland  always keep lower than 10 options in a select box. It is a UX Smell to have 100 options. Please use paper-autocomplete , it is a lot faster. 
@xomaczar  , making it use virtual-repeat will be really cool.

But still , more than 10 options in a combo box is not a good idea, even performance is good.

miguelcobain:
@eduardoweiland this was probably the cause of your issues.

lukaiser:
Hey guys!
What is the current status? Is anybody still working on this?

miguelcobain:
@lukaiser I don't think so, but I plan to tackle this at some point.
This is exact direction the project will take to implement this.

lukaiser:
@miguelcobain All right. I am interested in the feature and could help. I currently don't know when and how much time I have, as it doesn't have top priority, but I will look in to it.

cs3b:
@lukaiser recently I wrote my own date picker (exactly 7 lines of custom js code :-)) It's not something that could be merged into ember paper repo, still solve the problem, with low cost of side effects.

https://medium.com/@cs3b/a-component-in-7-lines-of-javascript-17207f6e58db#.2ueotm38i

bbooth:
Looking into helping pick up this up and help out.  Does anyone know where this work finished up in the grand scheme?

miguelcobain:
Closing this as this PR is now obsolete.

The plan is to implement the datepicker in ember-paper using ember-power-calendar under the hood and customizing it to render the markup that angular material styles expect.

I'll probably be able to do it in about a month or so.

PR #609, opened 2017-01-04
fixed async search behavior for autocomplete

ibarrick:
Behavior of async searches more closely matches AM now. I had to override either the `handleAsyncSearchTask` or the `search` action and I opted to overriding the task. I don't anticipate either of them changing substantially in the future and I would have had to add a task to make it work the other way. I can switch it if you want though.

Wait for Travis to finish though because my phantomjs was throwing undefined symbol errors at me but I think it's a local problem, we'll see if Travis breaks.

### DIFF
--- a/addon/components/paper-autocomplete.js
+++ b/addon/components/paper-autocomplete.js
@@ -6,8 +6,14 @@ import PowerSelect from 'ember-power-select/components/power-select';
 import ValidationMixin from 'ember-paper/mixins/validation-mixin';
 import ChildMixin from 'ember-paper/mixins/child-mixin';
 import { indexOfOption } from 'ember-power-select/utils/group-utils';
+import { task } from 'ember-concurrency';
+import { countOptions } from 'ember-power-select/utils/group-utils';
 
 const { assert, computed, inject, isNone, defineProperty } = Ember;
+
+function toPlainArray(collection) {
+  return collection.toArray ? collection.toArray() : collection;
+}
 
 /**
  * @class PaperAutocomplete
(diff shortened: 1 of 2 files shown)


ibarrick:
Looks like it's good :+1:

miguelcobain:
I'm concerned that we're overriding so much of eps stuff. Updates are more difficult that way.
Can you think of anything that eps could provide to make this easier?

ibarrick:
beforeSearch and afterSearch hooks would mean I didn't have to override anything (except the hooks). All I need is to run some code before the search fires and after it finishes. In order to wait on it finishing I either have to make a new task that calls the old one or just override the old one. With hooks I could just override the hooks to do what I need.

xomaczar:
Any updates on this? cc @cibernox

cibernox:
No, I didn't look into this when I had the chance and now I don't even remember what the desired behaviour was 😭 

I think I remember something like that the select shouldn't open until there is results to show. Is that the desired behaviour?

xomaczar:
@ibarrick we need those hooks from eps to hide dropdown content while async is pending right?

cibernox:
I'd like to focus on the desired behaviour, because if that is the case I've achieved it already in a different way somewhere else and it might be reusable for this situation: https://github.com/cibernox/ember-power-select-typeahead/blob/master/addon/components/power-select-typeahead/trigger.js#L42-L52

But I want to understand exactly the desired behaviour to if it really requires reaching in the internals. Perhaps we can do it already with the publicAPI and `didReceiveAttrs`

xomaczar:
@ibarrick was this ever addressed? Can it be closed?

PR #615, opened 2017-01-08
Chips: pass event into focusMovingTo

### LINKED ISSUE #611: paper-chips throws error on blur on Firefox
This seems similar to #606 but it's a bit different. Can be reproduced on the [demo page](http://miguelcobain.github.io/ember-paper/release-1/#/components/chips) when using Firefox (this error does not occur in Chromium). Just clicking on the field then clicking out of it gives the error `ReferenceError: event is not defined`. The field keeps the focused state after loosing focus.

Seems like it is caused by this line https://github.com/miguelcobain/ember-paper/blob/7ba0c103ebd093792c60241000eae1f0df4fe29c/addon/components/paper-chips.js#L232.

```javascript
if (!isEmpty(event) && !isEmpty(event.relatedTarget) && this.$(event.relatedTarget).is(selector)) {
  return true;
}
```

The variable `event` seems to be global. Replacing it with `window.event` seems to solve the issue.

pauln:
Fixes #611 (paper-chips throws error on focus out on Firefox)

### DIFF
--- a/addon/components/paper-chips.js
+++ b/addon/components/paper-chips.js
@@ -80,8 +80,8 @@ export default Component.extend({
       }
     },
 
-    inputBlur() {
-      if (this.focusMovingTo('.ember-power-select-option')) {
+    inputBlur(_, event) {
+      if (this.focusMovingTo('.ember-power-select-option', event)) {
         // Focus has shifted to an item - don't mess with this event.
         return true;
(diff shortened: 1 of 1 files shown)


miguelcobain:
Thanks for this and sorry for the terrible delay!

Issue #616, opened 2017-01-10
Chips demo page is horribly mangled in IE11


pauln:
The Chips demo page is completely unusable in IE11:
![image](https://cloud.githubusercontent.com/assets/574359/21826733/7fdd7dcc-d7ed-11e6-88db-a6dd291d0bcb.png)

Removing `class="flex"` from the `paper-card-content` elements fixes this (PR inbound).

pauln:
There are other compatibility issues with IE which may or may not need to be addressed, but now that this has landed, it's actually possible to see the Chips in IE11 in order to do so.

PR #617, opened 2017-01-10
Demo - Chips: Remove class="flex" from paper-card-content elements

pauln:
This fixes display in IE11 (see #616).  Note that there are other compatibility issues with IE11, but this at least allows you to see the chips demos in order to work on them.


miguelcobain:
Thanks!

PR #621, opened 2017-01-12
add setAllChilds function for paper-form

BennyAlex:


BennyAlex:
@miguelcobain Can you merge it please? Its really usefull to sett all inputs to invalid-touched. I do the input validation with ember-cp-validators and if the value isnt valid, I want to set the fields to invalid-touched....

PR #629, opened 2017-01-18
Modification to the index.js include directory

3tarazona:
With this little fix the ember-addon will be able to retrieve the parent application while being used from another addon. 

This also provides compatibility for ember version lower than 2.5 which don't provide the findHost helper


miguelcobain:
Thanks!

PR #646, opened 2017-02-15
Fixes bug in disabled slider

JustInToCoding:
Fixes the possibility to change the slider through clicking while it's disabled.


miguelcobain:
Thanks!

PR #647, opened 2017-02-15
paper-form enterToSubmit option

tyleryasaka:
This is a possible solution to https://github.com/miguelcobain/ember-paper/issues/460.

Triggers the `onSubmit` action when a user presses the enter key in an input field, if `enterToSubmit=true` is set on `paper-form`. This replicates expected html form behavior.

Of course, this only works when using the contextual `form.input` component. It doesn't do anything for `paper-autocomplete` or `paper-select` right now. Implementing that gets complex, and I'm not sure that it would even be desired behavior.

### DIFF
--- a/addon/components/paper-form.js
+++ b/addon/components/paper-form.js
@@ -15,6 +15,7 @@ const { Component, computed } = Ember;
 export default Component.extend(ParentMixin, {
   layout,
   tagName: '',
+  enterToSubmit: false,
   isValid: computed.not('isInvalid'),
   isInvalid: computed('childComponents.@each.isInvalid', function() {
     return this.get('childComponents').isAny('isInvalid');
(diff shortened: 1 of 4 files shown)


miguelcobain:
@tyleryasaka After a lot of thought, I'm more and more inclined to make `paper-form` to have a tag of `<form`. Here are my reasons:

1. we would get the benefits of the default browser behaviour, like enter to submit
2. html-wise, the app is semantically more meaningful
3. hbs-wise, I think it is expected that a paper-form generates a form tag
4. I don't think this would affect people's styles, especially because you can put your paper-form higher in the hierarchy, if needed
5. if people really want a tagless component, that can easily do `{{#paper-form tagName=""`

tyleryasaka:
@miguelcobain Sounds good! I agree that a form tag makes more sense.

Is this something I can take a stab at implementing? Or is someone else already on it?

miguelcobain:
@tyleryasaka You can go ahead, sure. :)

Please add a test for the "enter to submit" case.
If you have any questions feel free to show up at #e-paper channel in ember's slack.

Thank you!

PR #649, opened 2017-02-17
Makes paper-form an actual html form

tyleryasaka:
My second attempt at solving https://github.com/miguelcobain/ember-paper/issues/460, this time by making `paper-form` an html form.

Some things to note:
- Pressing the enter key to select an item in a `paper-autocomplete` (as well as a `paper-select` I assume) will submit the form - probably not ideal behavior. I attempted to find a simple way to prevent the `enter` keypress event from propagating when an item in the dropdown is selected, but I wasn't able to find a solution without hacking into `ember-power-select`. But maybe I missed something.
- I wasn't able to actually write a test for pressing the enter key in a form. From what I can tell this isn't possible with jQuery. Instead, I tested that the `onSubmit` action is called when an input of type submit within the form is clicked. This gives us some test coverage to make sure `paper-form` is in fact a real html form.

### DIFF
--- a/tests/integration/components/paper-form-test.js
+++ b/tests/integration/components/paper-form-test.js
@@ -182,3 +182,23 @@ test('form submit button is of type submit', function(assert) {
 
   assert.equal(this.$('button').attr('type'), 'submit');
 });
+
+test('form `onSubmit` action is invoked when form element is submitted', function(assert) {
+  assert.expect(1);
+
+  this.set('onSubmit', () => {
+    assert.ok(true);
+  });
+
+  this.render(hbs`
+    {{#paper-form onSubmit=(action onSubmit) as |form|}}
+      {{form.input value=foo onChange=(action (mut foo)) label="Foo"}}
+      {{form.input value=bar onChange=(action (mut bar)) label="Bar"}}
+
+      <input type="submit" value="Submit">
+
+    {{/paper-form}}
+  `);
+
+  this.$('input').click();
+});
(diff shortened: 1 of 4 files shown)


miguelcobain:
@tyleryasaka this looks good.

Can you please update the docs to reflect this change?

tyleryasaka:
@miguelcobain updated the docs. How does that work?

tests/integration/components/paper-form-test.js
```diff
@@ -50,12 +50,12 @@ test('form `onSubmit` action is invoked', function(assert) {
       {{form.input value=foo onChange=(action (mut foo)) label="Foo"}}
       {{form.input value=bar onChange=(action (mut bar)) label="Bar"}}
 
-      <button onclick={{action form.onSubmit}}>Submit</button>
+      <a {{action form.onSubmit}}>Submit</a>
```
miguelcobain: Do we need to change this to an `<a`?

tyleryasaka: @miguelcobain It needs to be something besides `button`, because html buttons submit forms automatically. When it's a button here, the `onSubmit` action is called twice. Once because the action is assigned directly to the button, and once because of the form's `submit` handler.

miguelcobain: @tyleryasaka That's what happens by default, but I changed it to `<button type="button"`.

tyleryasaka: Ah ok - I didn't realize that could be done. 👍

miguelcobain:
It was merged. Thanks. :)

tyleryasaka:
😃

PR #656, opened 2017-02-23
Enhance customizability paper-{form, input, radio-group}

LevelbossMike:
Hi and thanks again for this great addon! As recently discussed in slack I looked across existing components and added the ability to customize the sub-components that are yielded by parent-components. 

Autocomplete sub-components were already configurable. I left out the card-components because as far as I get it they are only provided because of styling concerns and extending them should be enough when you need to customize those.

### DIFF
--- a/tests/integration/components/paper-form-test.js
+++ b/tests/integration/components/paper-form-test.js
@@ -1,5 +1,8 @@
+import Ember from 'ember';
 import { moduleForComponent, test } from 'ember-qunit';
 import hbs from 'htmlbars-inline-precompile';
+
+const { Component } = Ember;
 
 moduleForComponent('paper-form', 'Integration | Component | paper form', {
   integration: true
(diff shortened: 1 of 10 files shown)


tests/integration/components/paper-form-test.js
```diff
@@ -182,6 +185,22 @@ test('form submit button is of type submit', function(assert) {
...
 
+test('form submit button component can be customized by passing `submitButtonComponent`', function(assert) {
+  assert.expect(1);
+
+  this.register('component:custom-submit-button', Component.extend({
+    classNames: ['custom-submit-button']
+  }));
+
+  this.render(hbs`
+    {{#paper-form submitButtonComponent='custom-submit-button' as |form|}}
```
miguelcobain: Can we change to double quotes in hbs code to keep the consistency with other tests/templates?

miguelcobain:
@LevelbossMike Can we have an entry in the changelog for this, please?

Other than that, this looks great and ready to merge!

LevelbossMike:
@miguelcobain all done

miguelcobain:
@LevelbossMike thanks!

PR #663, opened 2017-02-28
Fix binding style attribute warning in paper-backdrop

### LINKED ISSUE #659: [paper-backdrop] Warning: Binding style attributes
paper-backdrop causes a warning as `backdropStyle` is not considered safe
https://ember-twiddle.com/dcde26757f2b4b94acc8e724eae08877?openFiles=templates.application.hbs%2C
(i couldn't use ember-paper in the twiddle as an addon, so i added the ember-backdrop file directly in ember-twiddle)

Changing
```javascript
backdropStyle: computed('fixed', function() {
    return this.get('fixed') ? 'position:fixed;' : null;
  }),
```
to
```javascript
backdropStyle: computed('fixed', function() {
    return this.get('fixed') ? Ember.String.htmlSafe('position:fixed;') : null;
  }),
```
seems to fix the problem. Am I missing something else?

staskij:
Fixes #659


miguelcobain:
Thank you!

PR #670, opened 2017-03-16
use computed.bool instead of directly messages length

panthony:
because `computed.or` does not coalesce to bool, we otherwise end up with the number of errors instead of true in `isInvalid` and `isInvalidAndTouched`

Signed-off-by: Anthony Pessy <anthony.pessy@gmail.com>

### DIFF
--- a/addon/components/paper-input.js
+++ b/addon/components/paper-input.js
@@ -38,8 +38,7 @@ export default Component.extend(FocusableMixin, ColorMixin, ChildMixin, Validati
 
   iconComponent: 'paper-icon',
 
-  hasErrorMessages: computed.bool('validationErrorMessages.length'),
-
+  // override validation mixin `isInvalid` to account for the native input validity
   isInvalid: computed.or('hasErrorMessages', 'isNativeInvalid'),
 
   hasValue: computed('value', 'isNativeInvalid', function() {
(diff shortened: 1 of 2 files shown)


miguelcobain:
Can you convert these to integration tests, please?

miguelcobain:
Can we promote

```js
hasErrorMessages: computed.bool('validationErrorMessages.length'),
```

to the Mixin

and then paper-input would just override `isInvalid` to account for native invalids?

panthony:
@miguelcobain In `ValidationMixin` we would have:

```javascript
  hasErrorMessages: computed.bool('validationErrorMessages.length'),

  isInvalid: computed.bool('validationErrorMessages.length'),
```

And in paper-input:

```javascript
  isInvalid: computed.or('hasErrorMessages', 'isNativeInvalid'),
```

Is that correct?

miguelcobain:
I was thinking, for consistency

```
// validation-mixin.js
  hasErrorMessages: computed.bool('validationErrorMessages.length'),
  isInvalid: computed.bool('hasErrorMessages'),
```
and
```
// override validation mixin `isInvalid` to account for the native input validity
isInvalid: computed.or('hasErrorMessages', 'isNativeInvalid'),
```

miguelcobain:
Thanks!

PR #674, opened 2017-03-17
replace flex attribute by flex class

### LINKED ISSUE #673: What should I use: flex=50 or class="flex-50"
In the demo page is a paper-input with flex=50, but there also elements with class="flex-50".

In my application it looks like that flex=50 doesnt work.

So can you provide any information on this please?

panthony:
The `flex` attribute was replaced in favor of `flex-X` classes.

Closes #673


miguelcobain:
These were leftovers from when I removed flex attribute styles. Thank you!

PR #683, opened 2017-03-24
Use ember-power-select publicAPI in autocomplete

JustInToCoding:
Fixes a bug where paper-autocomplete wouldn't resolve promises passed into 'selected' property.

Basically comes down to using ember-power-select's publicAPI object instead of using the 'selected' property directly.


BryanHunt:
@miguelcobain Can we get this merged?  This is a big deal for me.

miguelcobain:
Thanks a lot for this. It was indeed a problem.
I'm very sorry for taking so long to merge this.

PR #689, opened 2017-04-13
Upgrade dependencies

bjornharrtell:
Except Ember and its core plugins, will try that in a separate PR. These upgrades passes tests locally and the dummy application seems to run fine.


bjornharrtell:
Seems `paper-autocomplete` is broken after the upgrade, i suspect the `ember-power-select` upgrade. Will try to fix.

bjornharrtell:
Seems not trivial to upgrade `ember-power-select` and the related `ember-basic-dropdown`. Reverted to hopefully handle that separately.

miguelcobain:
Thanks! 👍

PR #690, opened 2017-04-13
Upgrade Ember

bjornharrtell:
Depends on #689.


bjornharrtell:
Rebased.

bjornharrtell:
Another rebase with additional upgrades.

bjornharrtell:
With the latest round of updates including adaptions with a `ember init` run, the default package manager is now yarn and the qunit problem that was worked around by using an older version (#694) is no longer needed.

bjornharrtell:
Appears I made a premature conclusion, the qunit issue is still there. The yarn.lock was not upgraded.

miguelcobain:
Thanks!

PR #693, opened 2017-04-13
Consistent SASS theme colors

wmadden:
For some reason the foreground colors are set during preprocessing to fixed values, making it impossible to alter them using SASS variables. This PR replaces the generated constants with SASS expressions, making it possible to override them, and strips dead code from `lib/angular-scss-filter.js`.

### DIFF
--- a/app/styles/color-palette.scss
+++ b/app/styles/color-palette.scss
@@ -7,7 +7,7 @@ $strong-light-contrast-color: rgb(255, 255, 255) !default;
 $foreground-light: (
   '1': rgba(255, 255, 255, 1.0),
   '2': rgba(255, 255, 255, 0.7),
-  '3': rgba(255, 255, 255, 0.3),
+  '3': rgba(255, 255, 255, 0.5),
   '4': rgba(255, 255, 255, 0.12),
   'shadow': '1px 1px 0px rgba(0, 0, 0, 0.4), -1px -1px 0px rgba(0, 0, 0, 0.4)'
 ) !default;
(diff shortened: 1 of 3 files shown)


wmadden:
@miguelcobain ping

miguelcobain:
@wmadden sorry for the late reply. I had to do some changes to the filter and didn't know if they influenced this.
Can you please rebase and fix conflicts?

wmadden:
@miguelcobain done

miguelcobain:
@wmadden looks like the CI failed: https://travis-ci.org/miguelcobain/ember-paper/jobs/230832609

wmadden:
Ah nuts 😞 I'll fix it later

miguelcobain:
Here's an interesting regex AM itself is using: https://github.com/angular/material/blob/master/src/core/services/theming/theming.js#L889

Falibur:
@wmadden hope that you will and can fix this 👍

wmadden:
@miguelcobain I've fixed the issue on this branch but unfortunately the build is broken due to [this issue](https://github.com/yarnpkg/yarn/issues/3485) with Yarn and node-sass.

You've probably seen it before - locally it's fixed by running `npm rebuild node-sass`, after which the tests all pass. On CI it might be fixed by just rerunning the build and clearing the cache, I'm not sure, but you'll probably see the same failure on master.

jiaweihli:
You can also add `npm rebuild node-sass` to CI after updating npm dependencies - this should fix it.

There a few suggestions floating around to add this to the `postinstall` script in `package.json`, but IIRC this was buggy in some situations (may have been fixed since).

wmadden:
@miguelcobain looks like this issue is fixed in https://github.com/miguelcobain/ember-paper/pull/709

miguelcobain:
#709 is not ready to merge. It still has issues (only #705)

miguelcobain:
I added the workaround for the yarn bug on master. This PR looks like it simplifies things a lot while providing more flexibility as well. 👍

wmadden:
Nice work @miguelcobain, looks like the build is passing now 👍

Falibur:
Nice work you all :thumbsup:

v3ss0n:
Nice !!! that is awesome!

PR #694, opened 2017-04-13
Specifies exact version of qunitjs.

tyleryasaka:
`qunitjs` is a dependency of `ember-cli-qunit`. By not explicitly stating the version we want, we get the latest 2.x version. The problem is that anything after 2.2.1 breaks our tests at the moment.

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,8 @@
     "ember-transition-helper": "0.0.6",
     "eslint-plugin-ember-suave": "^1.0.0",
     "liquid-fire": "0.27.0",
-    "loader.js": "^4.0.10"
+    "loader.js": "^4.0.10",
+    "qunitjs": "2.2.1"
   },
   "dependencies": {
     "angular-material-source": "angular/material#v1.0.9",


bjornharrtell:
Funny observation: It is always 38 failing cases and they all involve explicitly calling click() on an element.

miguelcobain:
Thank you for this. I'll merge this, but it would be nice to find the exact cause of the failures.

wmadden:
@miguelcobain it's probably related to this issue in `ember-cli-qunit` (https://github.com/ember-cli/ember-cli-qunit/issues/162), which was an effect of an API change in QUnit (https://github.com/qunitjs/qunit/issues/1145) which broke a lot of Ember projects.

bjornharrtell:
I also thought that was it, but it seems explicitly setting `qunitjs` to 2.3.2 (clearing out `node_modules` and `npm install`) still fails on the 38 cases. The error message is slightly different (`Assertion occured after test had finished` vs. `Assertion after the final...`) which comes from https://github.com/qunitjs/qunit/blob/master/src/test.js#L375.

PR #699, opened 2017-04-21
Removed paper-wormhole initializer

### LINKED ISSUE #665: #paper-wormhole div not cleaned up in acceptance tests
In [this initializer](https://github.com/miguelcobain/ember-paper/blob/master/addon/initializers/paper-wormhole.js), Ember Paper adds a `div#paper-wormhole` to the DOM when the app boots. However, when the app shuts down this `div` is never removed. This happens with every acceptance test, leaving a bunch of `div#paper-wormhole`'s in the DOM at the end of a test run.

I propose Ember Paper simply only adds the div if it's not already there:

```js
if (document.getElementById('paper-wormhole')) {
  return;
}
```

morschtk:
Added functionality to rendering in contentFor when build env is not test, and rending paper-dialog inside of #ember-testing when build env is test.

Fixes #665.


miguelcobain:
@morschtk Thank you! I've been wanting this for a while.

PR #703, opened 2017-04-28
Autocomplete inner text

ibarrick:
pass the attrs block into the trigger so that if the user ops into tracking searchText manually, the component will have referential transparency.

### DIFF
--- a/addon/components/paper-autocomplete-trigger.js
+++ b/addon/components/paper-autocomplete-trigger.js
@@ -8,7 +8,7 @@ export default Component.extend({
   tagName: 'md-autocomplete-wrap',
   classNames: ['md-show-clear-button'],
   classNameBindings: ['noLabel:md-whiteframe-z1', 'select.isOpen:md-menu-showing'],
-
+  searchText: computed.oneWay('select.searchText'),
   noLabel: computed.not('extra.label'),
   _innerText: computed.oneWay('searchText'),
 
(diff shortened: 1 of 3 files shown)


ibarrick:
@xomaczar That's probably why the tests for earlier versions failed. Any idea what else I can use? I need to see if the user passed in searchText and ignore innerText if they did.

xomaczar:
Why do we define searchText as an empty string in autocomplete component? If we set it to undefined you can just compare against that.

ibarrick:
@xomaczar That works to change the text underlying the input but now I'm running into an issue where the options array is still filtered in an unexpected manner. We need an e-p-s way of initializing searchText (which is stored in publicAPI) since that governs what options show up.

ibarrick:
@miguelcobain I changed the component to use the `searchText` variable that was already built into e-p-s which helped resolve a lot of other issues I discovered such as the options array not being updated when the `searchText` attribute changed. I'm not sure why we implemented it outside of e-p-s in the first place but it seems to work this way.

xomaczar:
So is searchText two way bound or not - We are still exposing onSearchTextChanged and user must mutate the searchText himself

ibarrick:
I found all kinds of issues with this once I started using it in production. I'm going to try to work through them and hopefully what comes out will be mergeable.

xomaczar:
Yeah I knew there was a reason we did not use it before :)

xomaczar:
@ibarrick can we close this. Was addressed in #956

PR #707, opened 2017-05-11
Clean up event handlers in paper dialog

joukevandermaas:
paper-dialog registered event handlers for each image in the dialog, because the size of the content can change when they load. It never cleaned up these handlers, which could cause tests to fail if the dialog in the test contained an image.


miguelcobain:
Thanks. 👍

Issue #705, opened 2017-05-01
Error using chips on Ember (CLI) 2.13: expected an implementation for 6


backspace:
Hey, thanks for all the work on this, it’s been helpful for me to have a lot of UI taken care of so I can focus on business logic.

I tried updating an application to Ember 2.13 today and ran into this error in the console, with part of the UI failing to render:

`ember.debug.js:19829 Error: expected an implementation for 6`
<details>
<summary>Stacktrace</summary>
<pre>
    at Object.debugAssert [as assert] (ember.debug.js:1498)
    at Compilers.compile (ember.debug.js:1201)
    at compileStatement (ember.debug.js:1200)
    at ember.debug.js:1200
    at Array.forEach (<anonymous>)
    at WrappedBuilder.compile (ember.debug.js:1200)
    at ComponentLayoutBuilder.compile (ember.debug.js:1173)
    at Object.compileLayout (ember.debug.js:1173)
    at Cache._emberMetal.Cache.owner [as func] (ember.debug.js:11330)
    at Cache.get (ember.debug.js:18578)
</pre>
</details>

I tracked it down to [this line](https://github.com/backspace/prison-rideshare-ui/commit/f91401396271aa284524a2d93bdb12d931f5e280) in my header template:

```
{{#paper-chips readOnly=true content=chips as |item|}}<span title='{{item.title}}'>{{item.label}}</span>{{/paper-chips}}
```

I’m admittedly using chips outside their intended purpose, just to show an indicator in the header when you’re viewing a “sandbox” site. I can accomplish this differently without using chips, I was just being lazy about CSS etc.

I’ve removed the use of `paper-chips` for now but hopefully I’ll find some time to look into this and propose a fix.

My limited investigation has shown me that the failure is unrelated to the block use; even when I changed the template to this it still produced the same error:

```
{{paper-chips readOnly=true content=chips}}
```

It’s my belief that this relates to Babel 6, based on the text of the error, but it’s pretty mysterious!

bjornharrtell:
The same error message popped up in https://github.com/miguelcobain/ember-paper/pull/578 but was worked around.. would be good to find out what the real cause is though.

backspace:
Thanks for that connexion, @bjornharrtell. I wonder if this is actually a missing tag from [the new Glimmer that uses integers to represent tags](https://github.com/glimmerjs/glimmer-vm/pull/379/files)?

backspace:
[...]

PR #709, opened 2017-05-18
Upgrade to Ember 2.13.3

bjornharrtell:


miguelcobain:
Is https://github.com/miguelcobain/ember-paper/issues/705 fixed on this ember version?

bjornharrtell:
Should be yes, this one: https://github.com/emberjs/ember.js/pull/15247.

bjornharrtell:
It also appears as if https://github.com/miguelcobain/ember-paper/pull/694 is no longer an issue. Removed tentatively and I'll see how Travis feels about all this.

bjornharrtell:
Getting `ENOENT: no such file or directory, scandir '/home/travis/build/miguelcobain/ember-paper/node_modules/node-sass/vendor'` failure at Travis.. which seem unrelated to this, not sure what is going on.

bjornharrtell:
Still not sure what is going on at Travis but it appears I was too fast again to draw conclusions about #694 no longer being an issue. Seems like the `yarn.lock` is confusing me. After cearing out `node_modules` and installing with `yarn install --no-lockfile` the qunitjs related problem manifested itself again. So I have reverted that change and squashed this PR to only involve the Ember upgrade again.

bjornharrtell:
I suspect another "compatible" silent semver upgrade (I'm really starting to unlike ^ and ~ for deps), in this case of `node-sass`. Will investigate.

bjornharrtell:
Downgrading `node-sass` to 4.5.2 resolved the Travis CI issue. Should not be related to the Ember upgrade, and `node-sass` 4.5.3 which was released two days ago will likely cause build failures for any new commit. @miguelcobain let me know if you want me to move this fix to a separate PR.

bjornharrtell:
I have created https://github.com/sass/node-sass/issues/1971 for the upstream issue.

bjornharrtell:
While testing the dummy application I discovered a runtime problem. When navigating away from the Grid List demo page I get the following exception:

`calling set on destroyed object: <dummy@component:paper-grid-list::ember1849>.layoutInvalidated = false`

After which stuff is broken.

bjornharrtell:
Fixed by checking for `isDestroyed` and if `true` bailing the custom `doLayout` implementation. Running through the dummy demo pages seems fine now.

miguelcobain:
That bug was already there.

bjornharrtell:
Ok, should I remove, keep or extract the grid fix to another PR? And about `node-sass` upstream claims it's probably an issue with the Yarn cache but removing the use of Yarn cache from the travis yml doesn't help, so I'm not sure how to proceed with that either. :(

bjornharrtell:
Verified that the `node-sass` thing is really a Yarn issue. Applied workaround from https://github.com/yarnpkg/yarn/issues/1981.

steven-ferguson:
I am still encountering #705 after upgrading to Ember 2.13.2

bjornharrtell:
If that's the cause it's unfortunate and unexpected. :( @steven-ferguson are you able to supply a test case?

bjornharrtell:
Can reproduce it when running the Chips demo locally. Sorry for missing it before. Have noted this in https://github.com/glimmerjs/glimmer-vm/issues/453.

rwjblue:
If there are more scenarios that throw errors (missing implementation of x type things), please open issues for each over on the glimmer-vm repo with an example template snippet that triggers the error. I'm happy to dig in and fix the various cases that have been mentioned...

bjornharrtell:
Remaining known regressions are now tracked by https://github.com/emberjs/ember.js/issues/15258.

bjornharrtell:
Upgraded to Ember 2.13.3 and can confirm the Chips demo now works without apparent issues. :heart: @rwjblue. :)

rwjblue:
Awesome, thank you for confirming!

bjornharrtell:
@miguelcobain this should now be ready for review/merge.

miguelcobain:
@bjornharrtell thanks a lot 👍

v3ss0n:
i can't find ember 2.13.3 , how to install it ?

PR #714, opened 2017-05-26
Small correction to the doc (dummy app)

plcarmel:


miguelcobain:
I think `onSubmit` is the correct name: https://github.com/miguelcobain/ember-paper/blob/master/addon/components/paper-form.js#L36

plcarmel:
Yes, never mind, my mistake. Sorry about that.

PR #716, opened 2017-05-28
Removes jQuery from paper-slider

NLincoln:


miguelcobain:
Thank you! Other components could benefit from PRs like this. 😅

PR #718, opened 2017-05-30
Only call `setValueFromEvent` if the event has a center

NLincoln:
This is because sometimes we can "coerce" the slider to receive events
that don't have this value at all, and it causes errors and is
generally not a good time

### DIFF
--- a/addon/components/paper-slider.js
+++ b/addon/components/paper-slider.js
@@ -141,7 +141,9 @@ export default Component.extend(FocusableMixin, ColorMixin, {
       return;
     }
 
-    this.setValueFromEvent(event.center.x);
+    if (event && event.center) {
+      this.setValueFromEvent(event.center.x);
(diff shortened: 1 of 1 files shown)


miguelcobain:
@NLincoln in what cases the center property is not present?

miguelcobain:
I've seen this happening but on old ember-paper versions. I remember taking a look at those cases and fixing them. That's why I'm curious.

NLincoln:
Hmmmm.... That is very interesting. I can try and find a stable reproduction then and maybe see about writing a test.

NLincoln:
I've found a stable reproduction for this issue on http://miguelcobain.github.io/ember-paper/release-1/#/components/slider

1. Highlight across that first slider (so that the R and value are highlighted)
1. Try to drag the slider
1. Console fills with errors

PR #720, opened 2017-06-01
can disable paper-dialog's backdrop opaqueness

panthony:
Add a way to disable the opaqueness of the paper-dialog.

Combined with `origin` and little css on `md-dialog` you can do nice "popover panels" next to clicked buttons, etc.


miguelcobain:
Awesome! Can we have an entry in the changelog for this?

Issue #706, opened 2017-05-10
paper-toast


miguelcobain:
> @xomaczar thanks again for your interest. I actually have what I think to be a good plan to tackle toasts.

### Step 1 - `paper-toast` component

Make a `{{#paper-toast` component. This will be our basic building block. It would work pretty much like a `{{#paper-dialog` and it would be public in case people want to make more customised things in a toast. A basic example:

```hbs
{{#if isToastOpen}}
  {{#paper-toast onClose=(action (mut isToastOpen) false)}}
    Toast was shown successfully!
  {{/paper-toast}}
{{/if}}

{{#paper-button onClick=(action (mut isToastOpen) true)}}
  Open toast
{{/paper-button}}
```

API:
- `hideDelay` - default `3000` - number of ms before triggering the `onClose` action. Note that the action is responsible for closing the toast, DDAU style, pretty much like dialogs. we could support a special value like `false` to make a toast that never automatically closes itself.
- `parent` - default `paper-wormhole` - this is the id of the parent of the toast we will render into, again very similar to dialogs. I'm not 100% sure if we can render toasts on paper-wormhole as well, but I don't see why we can't.
- `position` - default `'bottom left'` - possible values are `'bottom left'`, `'bottom right'`, `'top left'`, `'top right'`.
- `capsule` - default to `false` - Whether or not to add the `md-capsule` class to the toast to provide rounded corners.

The component will probably yield some util components that hide the material markup underneath:

```hbs
{{#paper-toast onClose=(action (mut isToastOpen) false) as |toast|}}
  {{#toast.text}}
    Custom toast!
  {{/toast.text}}
  {{#toast.close-button}}
    Close
  {{/toast.close-button}}
  {{!--
    for maximum flexibility, we should also yield the close action.
    users may trigger it in any way they want
  --}}
  <button onclick=(action toast.closeAction)>Close using ember</button>
{{/paper-toast}}
```

Notes:
- the component would include the ColorMixin to support `accent=true`.
- I believe we'll need to use ember-css-transitions for animating the toast.
- We need this building block, and there's no reason we shouldn't make it public (that is, we should export it to the consuming app). It may allow more complex toasts the service won't. Speaking of that...

### Step 2 - `paper-toast` service and `paper-toast-container`

Make a `paper-toast` service so we can show toasts more easily in js land.
[...]

PR #722, opened 2017-06-02
Adding Snackbar to Ember paper

henrixapp:
+ GestureMixin for swipe-detection
+ paper-snackbar-inner
+ paper-snackbar-text
+ Demo app
TODO: tests


miguelcobain:
@henrixapp thank you for your contribution!

We imagined toast support in an RFC: https://github.com/miguelcobain/ember-paper/issues/706

Also, we include hammer js which should be more than enough for gesture support. No need to port anything for that.

And I think paper-toast is the standard name AM uses.

henrixapp:
Yes, probably the name should be paper toast, because it does not strictly follow the rules of a snackbar. I tried to follow #706 during implementation :grin:. I did not notice hammerjs, so I ported the angular implementation. Should I rename it?

miguelcobain:
Yes, please.

henrixapp:
Should it yield the button, because it only adds `accent=true` by default?

miguelcobain:
This doesn't include the service mentioned in the rfc, correct?

henrixapp:
Yes. Didn't have time to implement that yet.

henrixapp:
What's missing:

- [ ]  Tests
- [ ] Animation
- [ ] Service

Have you any ideas for animations?

PR #726, opened 2017-06-09
Upgrade eps@1.8.5

xomaczar:


miguelcobain:
@xomaczar we should bump ember-basic-dropdown version as well, to the same version that eps uses, right?

xomaczar:
Forgot about that -- absolutely.

miguelcobain:
I'm concerned if this upgrade is just a matter of bumping versions. Please test if this upgrade breaks any functionality.

xomaczar:
You don't trust our test suite?

miguelcobain:
@xomaczar unfortunately some things are still poorly tested, e.g paper-select, only two tests there

miguelcobain:
But overall, I think we have very good tests.

xomaczar:
Upgrading to the latest ember-basic-dropdown breaks a bunch of things -- so
don't be concerned :), more fixes to come.

bjornharrtell:
As I have personal interest in this upgrade and looked into upgrading but @xomaczar got there first with what looks like clean changes and nice removals of no longer needed code. Tried out the branch locally and tried the autocomplete, menu and select components without any issues. Great work @xomaczar! :)

miguelcobain:
Great PR. 👌 
Thanks!

PR #728, opened 2017-06-09
Add NAME_KEY property to components and mixins.

cah-danmonroe:
This will fix most of the '(unknown_mixin)' labels shown currently in Ember Inspector.


miguelcobain:
This PR is great because it documents where the problem is.
However this 1) doesn't seem to be a very maintainable solution (more stuff to think about when creating components) 2) we should be adding code that only helps in dev mode and fixes tooling problems.

Spoke with folks from ember inspector channel on slack. This should be addressed by a babel plugin or some broccoli funnel, etc. 

Thanks for this anyway!

PR #730, opened 2017-06-10
Prepare for FastBoot 1.0

kratiahuja:
This change wraps the browser specific libraries with if (typeof FastBoot === 'undefined') {...} check so that they are eval'd and run in browser but only eval'd and skipped in Node.

This prepares the addon for the upcoming FastBoot 1.0. See ember-fastboot/ember-cli-fastboot#387

cc: @miguelcobain


miguelcobain:
@kratiahuja again, thanks, and we can update ember-paper entry in the list

v3ss0n:
i gotta test this about time we need it!

PR #733, opened 2017-06-13
use ember-basic-dropdown / ember-power-select test helpers for paper-select/paper-menu

### LINKED ISSUE #586: paper-select test helpers
I'm trying to write some tests on some of my components that uses a `paper-select`.

Since this is not a real `select` component it is quite difficult to interact with it.

I found in the `paper-select` integrations tests helpers to focus, open and click on it.

Maybe they should be provided to the client application to ease the manipulation in integration tests ?

panthony:
This PR factorize and expose test helpers to open and select options for `paper-select` and `paper-menu`.

I was not sure about the API and/or naming, open to suggestions if any.
 
closes #586

### DIFF
--- a/tests/integration/components/paper-menu-test.js
+++ b/tests/integration/components/paper-menu-test.js
@@ -1,60 +1,14 @@
 import { moduleForComponent, test } from 'ember-qunit';
 import hbs from 'htmlbars-inline-precompile';
 import wait from 'ember-test-helpers/wait';
-import run from 'ember-runloop';
+import { clickTrigger, nativeClick, fireKeydown } from '../../../tests/helpers/ember-basic-dropdown';
 import $ from 'jquery';
 
 moduleForComponent('paper-menu', 'Integration | Component | paper menu', {
   integration: true
 });
 
-function focus(el) {
-  if (!el) {
-    return;
-  }
-  let $el = $(el);
-  if ($el.is(':input, [contenteditable=true]')) {
-    let type = $el.prop('type');
-    if (type !== 'checkbox' && type !== 'radio' && type !== 'hidden') {
-      run(null, function() {
-        // Firefox does not trigger the `focusin` event if the window
-        // does not have focus. If the document doesn't have focus just
-        // use trigger('focusin') instead.
-
-        if (!document.hasFocus || document.hasFocus()) {
-          el.focus();
-        } else {
-          $el.trigger('focusin');
-        }
-      });
-    }
-  }
-}
-
-function nativeClick(selector, options = {}) {
-  let mousedown = new window.Event('mousedown', { bubbles: true, cancelable: true, view: window });
-  let mouseup = new window.Event('mouseup', { bubbles: true, cancelable: true, view: window });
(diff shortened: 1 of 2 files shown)


knownasilya:
@miguelcobain would love to see this merged

knownasilya:
@panthony I'd love to see this use https://github.com/cibernox/ember-native-dom-helpers

panthony:
@knownasilya Those test helpers looks neat! I think I'm gonna start using them in my own projects alongs with https://github.com/simplabs/ember-test-selectors 👍  :)

test-support/helpers/paper-basic-dropdown.js
```diff
@@ -0,0 +1,20 @@
+import wait from 'ember-test-helpers/wait';
```
xomaczar: Use ```wait``` from ```ember-native-dom-helpers```

test-support/helpers/paper-basic-dropdown.js
```diff
@@ -0,0 +1,20 @@
...
+
+export function clickTrigger(scope, options = {}) {
+  if (scope && scope.classList.contains('ember-basic-dropdown-trigger')) {
+    return click(`#${scope.getAttribute('id')}`, options);
+  } else {
+    return click('.ember-basic-dropdown-trigger', scope, options);
+  }
+}
+
+export function selectOption(scope, index) {
```
xomaczar: I think this makes it easier to read.

```js
export async function selectOption(scrope, index) {
   await clickTrigger(scope);
   await wait();
   await clickOption()
   return wait();
}
```

panthony: Agree that async/await is easier to read but not sure it was allowed since there is 0 occurence in ember-paper.

I started with a wait() because it was already there in tests (a reason why remove it?) and I didn't add wait() after click() because it returns it already returns it ( https://github.com/cibernox/ember-native-dom-helpers/blob/master/addon-test-support/click.js#L37 )

Maybe this instead ?

```javascript
export async function selectOption(scope, index) {
   await wait();
   await clickTrigger(scope);
   return clickOption();
}
```

panthony:
@miguelcobain I tried to use test helpers provided by power-select & dropdown for ember-paper tests but I failed to make the 'focus' test works on paper-menu.

The 'focusIn' method of `focusable-mixin` is just not triggered although `paper-menu-content-inner` calls focus on the correct item...

Otherwise, people using `{{paper-select}}` can directly use the helpers of power-select, maybe it is worth a note somewhere?

```javascript
import { clickTrigger, selectChoose } from 'my-app/tests/helpers/ember-power-select';


test('it can select an option', async function(assert) {
  this.render(hbs`{{my-component}}`);

  await clickTrigger('md-input-container');

  await selectChoose('md-input-container', 'value');

  assert.equal(find('.md-select-value').textContent.trim(), 'value');
});
```

No need to provide additional helpers from ember-paper.

panthony:
closing in favor of #780 & #781 to split what works and what does not.

PR #736, opened 2017-06-24
WIP Upgrade basic-dropdown and power-select

bjornharrtell:
The usage of basic-dropdown and power-select is quite complex and seems hard to upgrade. These are the components that either fork, extend or use stuff from these dependencies:

* paper-autocomplete-content
* paper-autocomplete-dropdown
* paper-autocomplete-options
* paper-autocomplete-trigger-container
* paper-autocomplete
* paper-menu-content
* paper-select-options
* paper-select-menu-trigger
* paper-select-menu-inner
* paper-select-options
* paper-select-search
* paper-select-trigger
* paper-select

Noted breaking upstream changes:

### ember-basic-dropdown

* The positioning logic now accounts for the position of the parent of container if it has position relative or absolute. The breaking part is that custom calculatePosition functions now take the destination element as third arguments, and the object with the options has now been moved to thr forth position.

* Passing to="id-of-elmnt" to the {{#dropdown.content}} component is deprecated. The way to specify a different destination for ember-wormhole is now by passing destination=id-of-elmnt to the top-level {{#basic-dropdown}} component. The old method works but shows a deprecation message.

* Unify calculatePosition and calculateInPlacePosition. Now the function receives an renderInPlace flag among its options, and based on that it uses a different logic. This reduces the public API surface. This new function is the default export of /utils/calculate-position. The individual functions used to reposition when rendered in the wormhole or in-place are available as named exports: calculateWormholedPosition and calculateInPlacePosition

* It is a problem for a11y to have aria-owns/controls to an element that it's not in the DOM, so now there is a stable div with the right ID that gets moved to the root of the body when the component opens.

* Don't display aria-pressed when the component is opened. The attribute is not present by default, but can be bound from the outside.

* Don't display aria-haspopup by default, but display it if passed in a truthy value.

* Use aria-owns instead of aria-controls to link trigger and content together.

* Allow to customize the ID of the trigger component. Now the dropdown uses a new data-ebd-id attribute for query the trigger reliably. Unlikely to be breaking tho.

### ember-power-select

* ember-basic-dropdown has improved the experience with A11y for screen readings. Some aria-* have changed and there is an invisible div that wasn't there before. EPS doens't rely on those attributes and it's unlikely that this will break for anyone, but just in case I'll bump a minor version number and keep it in beta for some days.


xomaczar:
This work has already been done and pending reviews/merge - #726

bjornharrtell:
Ah nice work @xomaczar and thanks for letting me know. :)

xomaczar:
Please review if u have time

xomaczar:
I like your write up - can u please very my PR addresses all of your points?

bjornharrtell:
I made a pretty quick review so might have missed something, but it does appear that your PR should take care of the breaking changes, notably that it removes "forked" code that would have been affected and seems to address the change for `calculatePosition`.

Issue #738, opened 2017-06-27
Whats left to drop Alpha , and Move to Beta?


v3ss0n:
Since we are using e-p alpha without problem in Production - serving over 9000 monthly active users now, i guess its time to announce Beta.
Whats left in features that needed to be matured to announce this a beta? Lets focus on those.

miguelcobain:
@v3ss0n I wanted to update grid-list component at least to have an API that is coherent with the new convention: http://miguelcobain.github.io/ember-paper/release-1/#/components/grid-list

Essentially, convert dash-cased attributes to camelCased. Also remove `md-` prefix.

Also `{{#paper-grid-list` should yield a `grid` hash, such that we can use it like
```hbs
{{#paper-grid-list}}
```

miguelcobain:
beta.1 was released 👍

PR #739, opened 2017-06-27
Update grid list

Subtletree:
* Use camelCase attribute names
* Use contextual component api.
* Use composability mixins
* Add basic render tests
* Remove demo grid-list warning

Attribute names are now camelCased e.g `rowHeight`. However the media suffixes are still dasherized `rowHeight-gt-xs`. Thought this would be inline with the `flex-gt-xs` class names and easier to read imo than `rowHeightGtXs`. Let me know if you want it all camelCase though.

Completes #738
p.s sorry about 'remote-tracking branch' commits


v3ss0n:
Awesome

tests/dummy/app/templates/demo/grid-list.hbs
```diff
@@ -1,49 +1,50 @@
...
-      {{#paper-grid-tile class="yellow"}}
-        {{#paper-grid-tile-footer}}
-            <h3>#3: (1r x 1c)</h3>
-        {{/paper-grid-tile-footer}}
-      {{/paper-grid-tile}}
-
-      {{#paper-grid-tile class="blue" md-rowspan="2"}}
-        {{#paper-grid-tile-footer}}
+    {{#paper-grid-list
+      cols-sm="1" cols-md="2" cols-gt-md="6"
```
miguelcobain: Did you keep dash-cased attributes here on purpose? It is true that size prefixes (`-sm`, `-xs`, `-lg`, etc) usually are dash-cased. However, I think consistency wins here. Trying to think of a good api for this.

Let me know what you think.

miguelcobain: Maybe we could take inspiration from https://flexi.readme.io/docs/attributes

Subtletree: Yes, mentioned it in the PR note.

Looks to me that flexi's api is essentially what we have now except we can't use simple names like `sm` and `md` as we have to differentiate between 3 properties with media options instead of their 1.

Only reasonable options I can think of are:
a) `cols="1"` `cols-gt-xs="2"`
b) `cols="1"` `colsGtXs="2"`
c) `cols="1 gt-xs-2"`

panthony: I find `colsGtXs` very hard to read

miguelcobain: @Subtletree how hard could it be to implement c) ?

Subtletree:
Ok ready for another review:
* Now uses attrs like `cols=“1 gt-xs-2 gt-md-4”`.
* Re worked internals, using computed properties and `didUpdateAttrs`
instead of observers.
* Fixed `undefined` error messages when tile `colspans` weren’t
compatible with grid `cols`.
* Fixed `removeListener` to properly remove event listeners.
* Should be some performance increase as it no longer re calculates
everything all the time.

Subtletree:
Think I know what's causing these errors, update soon.

Subtletree:
Passes locally 😕

Subtletree:
This appears to be hitting the qunit problem in #694 also.

@bjornharrtell Looks like you've fixed it in #747? Hopefully once that is merged this will pass.

miguelcobain:
Thanks a lot for this, @Subtletree !

PR #747, opened 2017-07-08
Upgrade dependencies incl. Ember 2.14

bjornharrtell:
Appears to be working fine except the virtual repeat.

Remaining tasks:

* [x] Fix lint problems
* [x] Fix "Assertion Failed: Cannot call get with 'options' on an undefined object." in the virtual repeat sample
* [x] Fix qunitjs upgrade issues


bjornharrtell:
Ready for review.

bjornharrtell:
Forgot to check qunit issues. Will do so and return with a new reviewable state.

bjornharrtell:
Hit by the qunit issue as fixed/described by https://github.com/miguelcobain/ember-paper/pull/694 only this time it does not appear to help to pin the qunitjs version. :(

bjornharrtell:
I just now noticed that running a failing test in isolation makes it pass.

bjornharrtell:
Think I've found the test case issue and fixed it. The Ember beta build fails at CI but I suspect it has nothing to do with this PR.

bjornharrtell:
Rebased. CI fails completely but I don't understand why, it seems it's not using the `.travis.yml` that's in this branch (unchanged from master).

bjornharrtell:
Rebased again. Still failing at CI but it runs as expected, failure is with ember-beta which is the same on the master branch.

bjornharrtell:
Rebased and squashed.

miguelcobain:
beta and canary are expected to fail. It's an upstream problem.
Thanks for this. Will merge once tests finish running.

PR #763, opened 2017-07-30
Added isTouched and isInvalid to paper-autocomplete

betocantu93:
Realized it wasn´t possible, and I thought it could be useful for some scenarios where you need to manually show validity from inputs, for merely UX purposes.

### DIFF
--- a/addon/templates/components/paper-autocomplete-trigger.hbs
+++ b/addon/templates/components/paper-autocomplete-trigger.hbs
@@ -5,6 +5,8 @@
     flex=true
     required=(readonly required)
     passThru=(readonly passThru)
+    isTouched=(readonly isTouched)
+    isInvalid=(readonly isInvalid)
     validationErrorMessages=(readonly validationErrorMessages)
     disabled=(readonly disabled)
     onChange=(action "handleInputLocal")
--- a/addon/templates/components/paper-autocomplete.hbs
+++ b/addon/templates/components/paper-autocomplete.hbs
@@ -33,6 +33,8 @@
       allowClear=(readonly allowClear)
       required=(readonly required)
       passThru=(readonly passThru)
+      isTouched=(readonly isTouched)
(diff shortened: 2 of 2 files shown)


backspace:
This would be useful for me too, would you be into merging this (post-rebase), @miguelcobain?

miguelcobain:
Can someone explain exactly what does this PR address?
Does autocomplete never shows error UI feedback at the moment?

backspace:
In my case, it’s because I’m converting a field from a plain text input to an autocomplete. I’m using the `isTouched` trick you described [here](https://github.com/miguelcobain/ember-paper/issues/216#issuecomment-302350507) to show validation errors that I get back from JSON API even when the field has received no input. After having converted, my acceptance test that shows the validation errors upon submission of a blank form is failing because the autocomplete field will only show errors if I’ve entered something into it.

Does that make sense? The code is open source so I could deploy it somewhere if that helps.

betocantu93:
Yes, in my case I need autocomplete to show validation errors when the user hasn't entered any input, for this specific case isRequired is not enough because it's not inside a paper-form (it's a paper-tabbed-ish form, which each tab has its own validations and logic) in the end, when I the users wants to submit the "whole paper-tabbed-ish form" I notify which tabs have an error, and if they select the tab they should see only the inputs with error, so if for example, there is tab A, B, C, D, only tab A and C are fully required to finish the whole form so if the user doesn't complete tab C (which has a paper-autocomplete)  I activate C tab and I need to show that the paper-autocomplete has an error. In Tab A I have normal paper inputs so I can easily just use something like isRequired=(readonly isTabAInvalid) and isTouched=(readonly (and isTabAInvalid (not model.temperature)) isTabAInvalid is only computed to true if it has an error while trying to end the tabbed form. 

Hope it make sense.

xomaczar:
This can be closed as #795 covers this case.

PR #765, opened 2017-08-01
add hint about wormhole for integration tests using paper-dialog

### LINKED ISSUE #630: [Missing information - dialog] ember-wormhole failed to render into '#paper-wormhole'
While creating a integration test for a dialog that was wrapped by my own component I got this error:

`Error: ember-wormhole failed to render into '#paper-wormhole' because the element is not in the DOM`

Could not find any issue or documentation on why this was happening for the Dialog component in particular. 

By reading these [tests](https://github.com/miguelcobain/ember-paper/blob/master/tests/integration/components/paper-dialog-test.js#L11)  I learned that for the Dialog tests to work properly I need to manually add this element `<div id="paper-wormhole"></div>` to my component test. Then it started to work fine.

Would be nice to have a place in the documentation with those special cases, something like "tips and tricks". I'm adding this issue hoping to help anyone that might have stumbled on the same problem.

panthony:
fix #630

### DIFF
--- a/tests/dummy/app/templates/cookbook.hbs
+++ b/tests/dummy/app/templates/cookbook.hbs
@@ -7,6 +7,11 @@
     Material Design <a href="https://material.google.com/components/text-fields.html#text-fields-character-counter">requires that a character count is displayed</a> for input elements which have a length restriction. However, when forms have many input elements, some may prefer to display the count only when the element is focused or invalid. This CSS achieves this.
   </p>
   {{code-snippet name="cookbook.no-char-count.scss"}}
+  <h3>Integration tests with <code>\{{paper-dialog}}</code></h3>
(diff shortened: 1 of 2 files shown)


miguelcobain:
@cibernox I'd like to borrow your knowledge on this one. How was this solved in ember-power-select et al?

cibernox:
In tests it uses the qunit-container (I believe it also works in mocha) https://github.com/cibernox/ember-basic-dropdown/blob/master/addon/components/basic-dropdown.js#L242-L248

miguelcobain:
@panthony aren't we doing the same thing here https://github.com/miguelcobain/ember-paper/blob/master/addon/components/paper-dialog.js#L35 ?

PR #770, opened 2017-08-04
Chips: Pass search and matcher through to autocomplete

pauln:
This allows for customisable searching in Chips autocomplete (beyond just specifying the `searchField`).  I would have done this sooner, but I've only just hit the need for it and realised that these aren't already being passed through...


miguelcobain:
Thanks. 👍

PR #775, opened 2017-08-08
Yield the dropdown public API from paper-menu

joukevandermaas:
This can be very useful in more advanced scenarios. The yielded actions are practically useless without the dropdown's unique id.


joukevandermaas:
The beta build failed somehow. It doesn't look related to this PR tho.

PR #776, opened 2017-08-08
Update paper-menu-item.hbs for href

RedTn:
I would like to be able to pass in the href for the paper-menu-item button that gets generated, our team is using the dropdown as a router navigator.


Subtletree:
Looks like `paper-menu-item` will need something like https://github.com/miguelcobain/ember-paper/blob/0ae981e1826cb7a404b005e485021ba53d0a1760/addon/components/paper-item.js#L47

Then `{{#if onClick}}` would need to be `{{#if hasPrimaryAction}}`

RedTn:
Thanks @Subtletree , updated

RedTn:
@miguelcobain Hi, I'm hoping I can still get my PR reviewed and merged. I added a test now but feel free to let me know if there is anything else you would like me to add/change. Thanks!

miguelcobain:
It seems like paper-menu-item already supports an `href` attribute.
Closing this. Thank you.

PR #780, opened 2017-08-10
use ember-power-select test helpers

panthony:
Use ember-power-select test helpers.

Those helpers are also usable directly in an application by importing:

```
import { clickTrigger, selectChoose } from 'my-app/tests/helpers/ember-power-select';
```


panthony:
This was fixed by merging #944

PR #782, opened 2017-08-10
fix(ac): remove md-show-clear-button style when disabled

xomaczar:
Currently ```paper-autocomplete-trigger``` component has ```md-show-clear-button``` class always set regardless of the state of the ```paper-autocompelete```  component (allowClear). This class sets padding on the inner input element to make sure that the text does not over-run the clear-button. This PR address 2 things:
1. If the ```paper-autocomplete``` is disabled, remove that class so more of the selected content is visible
2. If only apply ```md-show-clear-button``` class if allowClear is truthy.

### DIFF
--- a/addon/components/paper-autocomplete-trigger.js
+++ b/addon/components/paper-autocomplete-trigger.js
@@ -13,8 +13,10 @@ const { Component, isPresent, isBlank, run, get, computed } = Ember;
 export default Component.extend({
   layout,
   tagName: 'md-autocomplete-wrap',
-  classNames: ['md-show-clear-button'],
-  classNameBindings: ['noLabel:md-whiteframe-z1', 'select.isOpen:md-menu-showing'],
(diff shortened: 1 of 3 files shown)


xomaczar:
@miguelcobain ping

miguelcobain:
@xomaczar this PR adds custom styles. Are we extending Angular Material functionality here?

xomaczar:
Just a little bit - seemed like the cleanest way w/o changing AM styles...

xomaczar:
The main reason this style exists because toggling disabled true/false - if bound - creates weird side effects. Please checkout PR and remove this custom CSS rule

miguelcobain:
Will do. But does AM have this issue? If not, how did they solve it?

xomaczar:
They have the same issue - there is no fix there that I know of

miguelcobain:
This was fixed on https://github.com/miguelcobain/ember-paper/commit/41765e69991dff1fa1d8c70aa5f197c688ae0650

Thanks for bringing this up.

PR #793, opened 2017-08-22
add hint in docs for routable paper-tabs active state (fix #789)

### LINKED ISSUE #789: Routable tabs don't highlight the right route if you hard refresh
Hi, this isn't really a bug per-se. The solution just isn't in the docs yet, so I wanted to share how I did it.

The problem: If you have tabs for nested routes, then clicking on a tab will highlight it and transition to that route. However, if happen to transition here from somewhere else, the first tab is selected, not the one matching the current route.

The solution is to use `selected` and `value` as such:

```htmlbars
{{#paper-tabs selected=currentRouteName as |tabs|}}
  {{#tabs.tab class="flex" value="foo.nested-one" href=(href-to "foo.nested-one")}}
    Nested One
  {{/tabs.tab}}
  {{#tabs.tab class="flex" value="foo.nested-two" href=(href-to "foo.nested-two")}}
    Nested Two
  {{/tabs.tab}}
  {{#tabs.tab class="flex" value="foo.nested-three" href=(href-to "foo.nested-three")}}
    Nested Three
  {{/tabs.tab}}
{{/paper-tabs}}
```

And here is how I get the current route name in my controller:

```javascript
  applicationController: Ember.inject.controller("application"),
  currentRouteName: Ember.computed.reads("applicationController.currentRouteName"),
```

panthony:
This PR shows how to handles active state for routable `paper-tabs`.


miguelcobain:
Thanks a lot!

PR #795, opened 2017-08-23
Make externally provided errors work with paper-autocomplete

lucacorti:
Pass isTouched from paper-autocomplete to paper-autocomplete-trigger
and its inner paper-input.

External errors are not otherwise rendered because isTouched is always false on the inner `paper-input` component.


xomaczar:
@miguelcobain ping. Can this be merge in.

miguelcobain:
Thank you. 👍

PR #810, opened 2017-09-12
[WIP] Look ma, no components!

cibernox:
@miguelcobain This is just a 30m hack so you get the drill of how this can be done. 

Fun fact: I wrote everything in one go and worked on the first try 🔥 

I'll continue tonight. I already found one area where I should BuildTimeComponent: Components with templates.


cibernox:
It's important to note that this approach has one very significant caveat. It only works in Ember 2.10 and beyond, and node 6.

Making it work in node 4 is doable, I just didn't try yet. Making it work in Ember 2.9 and below might be too hard to be worth it.

cibernox:
Master:
```
 - dist/assets/dummy-d07754501a220f535cac21c72fdf9dc4.css: 460.13 KB (44.63 KB gzipped)
 - dist/assets/dummy-ee7bf31c31d29e9283e48ad84a76c4c0.js: 416.51 KB (65.76 KB gzipped)
 - dist/assets/dummy-fastboot-79227cfae53af54d4330b2e960cd1956.js: 1.32 KB (680 B gzipped)
 - dist/assets/vendor-75f40878596e268f96dd0e4a24b078bb.css: 1.5 KB (622 B gzipped)
 - dist/assets/vendor-afde3a38d4ecab6bf83d30b87841fe54.js: 1.76 MB (494.46 KB gzipped)
```

This branch:
```
 - dist/assets/dummy-d07754501a220f535cac21c72fdf9dc4.css: 460.13 KB (44.63 KB gzipped)
 - dist/assets/dummy-dfe0f6ddf5c0828913320e22bc72d694.js: 412.03 KB (65.15 KB gzipped)
 - dist/assets/dummy-fastboot-79227cfae53af54d4330b2e960cd1956.js: 1.32 KB (680 B gzipped)
 - dist/assets/vendor-75f40878596e268f96dd0e4a24b078bb.css: 1.5 KB (622 B gzipped)
 - dist/assets/vendor-d26f8a020ca154809e4c5afa2efa0379.js: 1.74 MB (493.48 KB gzipped)
```

The dummy app hasn't grown. In fact is 1kb smaller.

miguelcobain:
I assume that the performance benefits aren't at all proportional to the size gains. They should much higher.

cibernox:
Time to render the `#components/paper-card` URL of the dummy app (in development)

Before: 120ms
After: 85ms

Which I think it's pretty good perf boost because it only has 12 cards and also because this is the time that takes to render the entire page, which contains many more things than just the cards. The perf difference of the cards along is probably much bigger.

cibernox:
The downside is that the build time has increased significantly.
I haven't dedicated any time to improve the performance of the library yet tho.

PR #811, opened 2017-09-14
Fix paper-progress-circular infinite loop in acceptance tests

tyleryasaka:
I recently upgraded ember paper for my project and found that acceptance tests were failing left and right. I finally isolated the issue to this component, and found that renderCircle was getting caught in a recursive loop once the component started being destroyed. The tests were waiting on this component to finish rendering, and they would wait until the tests timed out.

At first glance it looks like I'm doing a lot here, but I'm just wrapping the contents of `renderCircle` inside an `if` statement (in fact, the same check that's being performed inside the `renderFrame` function defined inside this function).

### DIFF
--- a/addon/components/paper-progress-circular.js
+++ b/addon/components/paper-progress-circular.js
@@ -155,39 +155,41 @@ export default Component.extend(ColorMixin, {
 
   lastAnimationId: 0,
   renderCircle(animateFrom, animateTo, ease = linearEase, animationDuration = 100, iterationCount = 0, dashLimit = 100) {
-    let id = ++this.lastAnimationId;
-    let startTime = now();
-    let changeInValue = animateTo - animateFrom;
-    let diameter = this.get('diameter');
-    let strokeWidth = this.get('strokeWidth');
-    let rotation = -90 * iterationCount;
+    if (!this.isDestroyed && !this.isDestroying) {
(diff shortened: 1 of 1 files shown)


xomaczar:
Got the same issue - we need to merge this in - can you just return if component is destroyed/destroying, will make this PR much easier to review

xomaczar:
Ping @tyleryasaka

arschmitz:
Ran into this also

@xomaczar 100% agree early return is much better but just for future times with similar PR's that cant be so easily made better you can always add `?w=1`to url to ignore whitespace changes makes it much nicer to review 

https://github.com/miguelcobain/ember-paper/pull/811/files?w=1
gets it down to 2 lines lol

miguelcobain:
This fix was reworked a bit and is now on master. Thanks everyone.

xomaczar:
@arschmitz it's not about PR readability(it does help), it's more about code organization, maintenance, etc - I try to never wrap many code lines in if block - if I can help it

arschmitz:
@xomaczar i know thats why i said 
> 100% agree early return is much better but just for future times with similar PR's that cant be so easily made better

some pr's have large white spaces changes that just cant be avoided

xomaczar:
👍

PR #820, opened 2017-09-19
Add support for additional CSS class on paper-toast

pauln:
Due to `paper-toast` being tagless and wormholing the `paper-toast-inner` component, the `class` attribute needs to be passed through in order to allow additional CSS classes to be specified on the `md-toast` element.  Additionally, the `paperToaster` service should support a `toastClass` option (see AM's [$mdToast.simple()](https://material.angularjs.org/1.1.5/api/service/$mdToast#mdtoast-simple) documentation) do that toasts opened using the service can each have their own additional classes.

I've updated the `paperToaster` demo in a separate commit, so that you can choose whether to demo it or not.


xomaczar:
@miguelcobain thoughts? I also have a need for something like that?

miguelcobain:
This is perfect. Thank you and sorry for the delay.
Closes #829.

Issue #642, opened 2017-02-13
How can I reset searchText in autocomplete?


estshy:
Basically I need a select with search inside to allow user for adding new items to result collection from collection of possible options. Adding new item should cause search to be reseted. I though about using paper-autocomplete but unfortunately it doesn't work the way it should. Overwriting value passed as `searchText` to null doesn't do anything. Is there any official way of solving this issue?

miguelcobain:
@estshy how are "overriding" that value? I'd like to see your template, if possible.

estshy:
Sure. Example should show what I am trying to do

```
{{#each languages as |language| }}
  {{language.name}}    
{{/each}}

{{#paper-autocomplete
  label="Add another language"
  labelPath="name"
  searchField="name"
  options=options
  searchText=searchText
  onSelectionChange=(action 'addLanguage')
  as |lang select|}}
  {{lang.name}}
{{/paper-autocomplete}}

actions: {
  handleLanguageAdd(lang) {
    if (lang) {
      this.get('languages').pushObject(lang);
    }

    this.set('searchText', null);
  }
}
```

miguelcobain:
@estshy what happens if you `this.set('searchText', '');`?

pauln:
I ran into this while working on Chips, and the only workaround I was able to find was using jQuery to set the value to an empty string - once you've located the relevant trigger input, `input.val('');` will clear it.

If we can get an "official" way to reset/clear/set the searchText and have it actually clear the text in the trigger input, we should be able to use it to tidy up the Chips implementation slightly.

xomaczar:
Autocomplete keeps an internal state of recently searched text. When a user makes selection and then starts backspacing - md spec calls for search box to kee the recently search text minus deleted chars. However, give the latest use case - I believe it's a bug and we need to address it

estshy:
@pauln I've also tried dom manipulation but it doesn't work correctly. See @xomaczar response

pauln:
@estshy Have you tried calling `this.set('searchText', '');` followed by `input.val('');` (where `input` is the jQuery-wrapped trigger input)?  The current Chips implementation does this, and I've also done it with a standalone `paper-autocomplete` (outside of Chips).  If the combination of both doesn't do what you expect, describing what _does_ happen might help us figure out the best way to handle this situation :-)

estshy:
[...]

PR #824, opened 2017-09-21
fixed https://github.com/miguelcobain/ember-paper/issues/642 by clear…

markdicksonjr:
I wrote a test to reproduce https://github.com/miguelcobain/ember-paper/issues/642, then fixed it.

What was happening was: each time a key is pressed, it provides the value of the select to "v" in the setter of text (which will set _innerText) within paper-autocomplete-trigger.  When selected and searchText are set to something falsey, it still returns _innerText, the result of the key-presses:

```
text: computed('selected', 'searchText', '_innerText', {
    get() {
      let {
        selected,
        searchText,
        _innerText
      } = this.getProperties('selected', 'searchText', '_innerText');

      if (selected) {
        return this.getSelectedAsText();
      }
      return searchText ? searchText : _innerText;
    },
    set(_, v) {
      let { selected, searchText } = this.getProperties('selected', 'searchText');
      this.set('_innerText', v);

      // searchText should always win
      if (!selected && isPresent(searchText)) {
        return searchText;
      }

      return v;
    }
  }),
```

### DIFF
--- a/addon/components/paper-autocomplete-trigger.js
+++ b/addon/components/paper-autocomplete-trigger.js
@@ -64,6 +64,10 @@ export default Component.extend({
 
     if (oldSelect && oldSelect.isOpen && !select.isOpen && !loading && searchText) {
       this.set('text', this.getSelectedAsText());
+    } else if (oldSelect && select && oldSelect.selected && !select.selected) {
+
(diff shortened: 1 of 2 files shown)


miguelcobain:
Please reopen if still not solved.

PR #828, opened 2017-09-26
Tests added for paper chips, and fix made for lack of isPresent check

markdicksonjr:
Tests added for paper chips, and fix made for lack of isPresent check on autocomplete.actions when the input handles focus

### DIFF
--- a/addon/components/paper-chips.js
+++ b/addon/components/paper-chips.js
@@ -89,7 +89,7 @@ export default Component.extend({
       }
 
       // We don't want the autocomplete to open on focus - it'll open when the user starts typing.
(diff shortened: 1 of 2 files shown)


pauln:
This PR seems to predate the shift to module imports, so I've submitted a PR to @markdicksonjr to address the linting error from destructuring the Ember global.  That should result in these tests passing again; at that point, merging this PR would give us a starting point to flesh out more tests so that we would (hopefully) pick up issues such as #932 before cutting new releases in the future.

tpetrone:
@pauln Any chance to see this incorporated?

PR #832, opened 2017-09-28
Fix routable tabs example

steven-ferguson:
The example for routable tabs is not updating the `selected` property correctly. Steps to reproduce:

1) While on the tabs page, click `Nested Route`
2) Navigate away from the tabs page by clicking one of the other links in the sidebar
3) Navigate back to the tabs page by clicking the tabs link in the sidebar
4) `Nested Route` will still be highlighted even though you are on `demo.tabs.index`

This is happening because the `currentRouteName` on the controller is diverging from the routerService's `currentRouteName`. Making the `currentRouteName` property `readOnly` on the controller fixes the issue.

### DIFF
--- a/tests/dummy/app/controllers/demo/tabs.js
+++ b/tests/dummy/app/controllers/demo/tabs.js
@@ -15,7 +15,7 @@ export default Controller.extend({
 
   router: inject.service(),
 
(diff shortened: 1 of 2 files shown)


miguelcobain:
Good job here. Thanks. 👍

PR #834, opened 2017-10-02
paper-toaster: make duration and position configurable application wide

panthony:


miguelcobain:
Solved this in a slightly different way: https://github.com/miguelcobain/ember-paper/commit/88eca4d5a675a5e19a39ab635debd12dab2eb90c

Thanks! 👍

PR #842, opened 2017-10-12
Fix sidenav code snippet references

steven-ferguson:
The code snippets were overwriting each other because they were both named `sidenav`. I have added more descriptive names to avoid this.


miguelcobain:
Thanks. Nice catch.

PR #846, opened 2017-10-17
[FEATURE] Add onInvalid action to form.

Serabe:
Add `onInvalid` action, invoked when trying to submit an invalid form.

### DIFF
--- a/addon/components/paper-form.js
+++ b/addon/components/paper-form.js
@@ -48,6 +48,7 @@ export default Component.extend(ParentMixin, {
     onSubmit() {
       if (this.get('isInvalid')) {
         this.get('childComponents').setEach('isTouched', true);
+        this.sendAction('onInvalid');
       } else {
         this.sendAction('onSubmit');
         this.get('childComponents').setEach('isTouched', false);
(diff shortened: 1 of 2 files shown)


miguelcobain:
Very weird. I can merge this because it's clearly not related.

Regarding the PR, I love the idea, but I'm unsure about the naming.
`onInvalid` sounds like an action the is sent when the form becomes invalid, which is not the case.

What if we kept using `onSubmit` but we sent the validity of the form in the action?
People could then later treat the submit accordingly in their action.

Using your approach, I think we should rename the action. Perhaps `onSubmitInvalid`?

Let me know your thoughts. Thanks for the PR.

Serabe:
I would like to use `invalid` as the name of the event because [that is the standard name according to MDN](https://developer.mozilla.org/en-US/docs/Web/Events/invalid). If you don't like that option, `onSubmitInvalid` looks good too.

miguelcobain:
I was unaware of that event.
Makes sense to name this action accordingly since they attempt to do the same thing.
Thanks again!

Serabe:
Thanks!

Serabe:
@miguelcobain is there any planned release that includes this PR?

Thanks!

miguelcobain:
@Serabe `v1.0.0-beta.4` was published with this PR. Sorry for the delay.

Serabe:
Thanks!

PR #847, opened 2017-10-18
[New Feature] Allow specifying icon color

devotox:


miguelcobain:
I'm afraid this can lead to questions/requests like "if you have `color`, why don't you support background-color/margin/padding/any css property".

We could allow for a `style` attribute with you could use to add inline styles. Can you refactor?

devotox:
Yes that makes total sense. How would you want it implemented? Just combine
the size with the style attribute? I'm just thinking about duplicate styles
that would offset  the size attribute

pauln:
Sorry if I've missed a discussion elsewhere which justifies it, but is there actually a need for inline styles?  It's already possible to add custom CSS classes by specifying `class="foo bar"` on your `{{paper-icon}}`; you can then use CSS to set the colour (and plenty of other things).

If you're wanting to set an icon's colour dynamically at runtime (not from a preset palette), that seems like a sufficiently niche case that it might be better to create a custom component in your app which extends `paper-icon` and adds support for an additional `color` attribute.  I've just thrown together [a quick gist](https://gist.github.com/pauln/2d636fea06bd260f3b1471e74681449a) demonstrating a basic `paper-icon-colorful` component, which does just this.

devotox:
You may be right and I initially looked at just extending it as I have done
with the paper-input but I thought like size color is something big enough
that would be useful to more than just me and something someone wouldn't
want to make multiple classes for each possible color. But I've also done
this an extension of the class for my case so this can be closed

miguelcobain:
I think @pauln is right. Doing this for paper-icon would imply doing it for many other components as well.
Right now "class" is the way we custom style things.

PR #849, opened 2017-10-24
Allows you to use the html5 download attribute

devotox:
This especially works well with data URIs

### DIFF
--- a/addon/components/paper-button.js
+++ b/addon/components/paper-button.js
@@ -33,7 +33,8 @@ export default Component.extend(FocusableMixin, RippleMixin, ColorMixin, Proxiab
     'type',
     'href',
     'target',
-    'title'
+    'title',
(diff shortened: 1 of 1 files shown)


miguelcobain:
Thanks. 👍

devotox:
It's  my pleasure. Question... There is no way to change a tagName after
init is there?

miguelcobain:
There isn't. It's an ember restriction. You have some problem related with that?

devotox:
Not exactly I worked around it. I just noticed it when I was making this
PR. Since my href attribute came dynamically after render the component
remained a button I then tried didUpdateAttrs and realized it was not
possible but had never explicitly read that anywhere. Just wanted to
clarify my findings. Easy workaround was to set a default href and then
change it when I got the new value.

PR #850, opened 2017-10-25
Chips: Only fire focusIn/focusOut when whole component gains/loses focus

pauln:
Previously, moving focus between the chips and the input (for instance) would cause focusOut and focusIn to be triggered spuriously.  In part, this seems to have been caused by observing a computed property - the observer was firing when the computed property's dependent properties changed, instead of only when the computed value changed.  Other fixes include handling more transitions between internal elements and accounting for the autocomplete dropdown being wormholed when checking if the focus is moving to an autocomplete option.


pauln:
Rebased onto latest master.

Issue #853, opened 2017-11-01
paper-toaster is not compatible with ember-gestures


kbullaughey:
The `ember-gestures` addon for touch support adds events like `swipe`. However, glimmer doesn't allow tagless components to have event handlers and it performs a check in `Component#init` that results in an error that begins:

    You can not define a function that handles DOM events...

The check that is performed merely checks for the existence of methods the names of which match event names. So normally, without `ember-gestures`, `swipe` is not an event, and so the fact that `paper-toast` has a method `swipe` is not a problem. But once the event dispatcher is configured to support `swipe`, it's impossible to insert a tagless `paper-toast` component. 

It's not clear to me whether it's more sensible to try and fix `ember-paper` or `ember-gestures`, but given the popularity of both of these addons, it would be nice if they could work together.

PR #854, opened 2017-11-01
resolve swipe name conflict

kbullaughey:
Having a function that is named after an event on a tagless component is an error in glimmer. The paper-toast component has a method named `swipe`. Normally this isn't a problem because `swipe` is an official event registered with the event dispatcher. But when using `ember-gestures` there is a `swipe` event, making it impossible to use `paper-toast`. However, here, the tagless component has a method `swipe` out of coincidence. By renaming this method we solve the problem. This fixes issue #853.


miguelcobain:
Thanks. :+1:

PR #868, opened 2017-12-15
feat(ac): respect dropdownClass just like e-p-s

xomaczar:
Autocomplete should support dropdownClass attribute similar to e-p-s


miguelcobain:
👍

Issue #843, opened 2017-10-13
Performance suffers greatly due to "[Violation] Added non-passive event listener to a scroll-blocking 'touchmove' event."


fushi:
When using {{paper-checkbox}} and {{paper-switch}}, I get a console warning:

`[Violation] Added non-passive event listener to a scroll-blocking 'touchmove' event. Consider marking event handler as 'passive' to make the page more responsive. See https://www.chromestatus.com/feature/5745543795965952`

And performance tanks:
```
[Violation] 'setTimeout' handler took 5024ms
[Violation] Forced reflow while executing JavaScript took 2752ms
```

PR #870, opened 2017-12-17
Use native passive event listeners for ripple and tooltip

Subtletree:
To do: 
* Replace remaining jquery `on` / `off` with `addEventListener` / `removeEventListener`.
* Use `passive: true` where possible. (Anywhere where the handler does not call `preventDefault`)
* ~~Ripple mixin needs `unbindEvents`~~
  _Unnecessary as listeners are removed when element is destroyed in browsers newer than ie 7_

These two uses of `passive: true` remove 200 scroll-blocking violations in the test suite.

Related: https://github.com/miguelcobain/ember-paper/issues/843

### DIFF
--- a/addon/components/paper-tooltip.js
+++ b/addon/components/paper-tooltip.js
@@ -80,7 +80,9 @@ export default Component.extend({
     };
 
     anchorElement.addEventListener('focus', enterEventHandler);
-    anchorElement.addEventListener('touchstart', enterEventHandler);
+    anchorElement.addEventListener('touchstart', enterEventHandler, {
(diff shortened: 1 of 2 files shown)


panthony:
@Subtletree Hi, how's this PR going? Maybe we could split it so that we can at least merge the `passive: true` ?

Subtletree:
Think that's a good call to split it, I haven't looked at this since I posted it.

I've renamed this PR and will open an issue about removing the remaining jquery handlers.

@miguelcobain Your thoughts on merging this?

PR #874, opened 2017-12-27
Add checks to prevent calling methods on destroyed objects

panthony:
They are unfortunately no tests provided because this is kind of hard to test :/

They all come from errors reported into my app's [sentry](https://sentry.io).

- `Cannot read property 'style' of null`
> this.dropdownElement.style.transform = '';

- `undefined is not an object (evaluating 't.getBoundingClientRect')`
> var originBnds = this.copyRect(originator.getBoundingClientRect());

- `Cannot read property 'x' of undefined`
> this.setValueFromEvent(event.center.x);

- `Cannot read property 'get' of undefined`
> var inputElement = this.$('input').get(0);

This one since it comes from Hammer I don't know if it's a bug upstream or not I'm not familiar with Hammer at all.

- `Cannot read property 'find' of undefined`
> var focusTarget = this.$().find('.md-menu-focus-target');

### DIFF
--- a/addon/components/paper-input.js
+++ b/addon/components/paper-input.js
@@ -153,6 +153,9 @@ export default Component.extend(FocusableMixin, ColorMixin, ChildMixin, Validati
         }
         this.setValue(this.get('value'));
       });
+      if (this.isDestroying || this.isDestroyed) {
+        return;
+      }
       this.growTextarea();
       let inputElement = this.$('input').get(0);
       this.set('isNativeInvalid', inputElement && inputElement.validity && inputElement.validity.badInput);
--- a/addon/components/paper-menu-content-inner.js
+++ b/addon/components/paper-menu-content-inner.js
@@ -25,6 +25,9 @@ export default Component.extend(ParentMixin, {
 
   didInsertElement() {
     run.later(() => {
+      if (this.isDestroying || this.isDestroyed) {
+        return;
+      }
       let focusTarget = this.$().find('.md-menu-focus-target');
       if (!focusTarget.length) {
         focusTarget = this.get('enabledMenuItems.firstObject.element.firstElementChild');
--- a/addon/components/paper-select-content.js
+++ b/addon/components/paper-select-content.js
@@ -36,6 +36,9 @@ export default PaperMenuContent.extend({
   animateIn() {
     run.next(() => {
       run.scheduleOnce('afterRender', this, () => {
+        if (this.isDestroyed || this.isDestroying) {
+          return;
+        }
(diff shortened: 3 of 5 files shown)


panthony:
This is paper virtual repeat that is failing... does not look like it is because of this PR.

dancer-cat:
I guess I've encountered the same problem on closing paper-dialog. The following error is reproduced time after time:

```javascript
Uncaught TypeError: Cannot read property 'getBoundingClientRect' of undefined
    at Class.calculateZoomToOrigin (translate3d-mixin.js:107)
    at Class.willDestroyElement (translate3d-mixin.js:75)
    at Class.superWrapper (ember-utils.js:428)
    at Class.willDestroyElement (paper-dialog-inner.js:34)
    at Class.superWrapper [as willDestroyElement] (ember-utils.js:428)
    at Class.trigger (core_view.js:62)
    at Class.superWrapper [as trigger] (ember-utils.js:428)
    at ComponentStateBucket.destroy (ember-glimmer.js:3762)
    at SimpleBlockTracker.destroy (runtime.js:3261)
    at SimpleBlockTracker.destroy (runtime.js:3261)
}
```

**translate3d-mixin.js**

```javascript

 if (originator) {
        originator = Ember.$(originator).get(0);
//error occurs here:
        var originBnds = this.copyRect(originator.getBoundingClientRect());
```

Subtletree:
@dancer-cat I think that error might be because the selector you're passing to `origin` or `openFrom` is not being found in the dom.

That would be a good case for a development assertion with a more specific error.

dancer-cat:
@Subtletree The case is that I'm not working directly with DOM or selectors. I only work data bound to the form (paper-dialog) elements. The error occurs after an action attached to onclick button handler is executed, each 4th or 5th dialog opening. The save method is always executed successfully. 

,,,javascript
async  save(changeset){
    await changeset.validate();
    
    if (changeset.get("isInvalid"))
        return;
     
     const settings = this.settings;         
     later(() => {
        changeset.save();
        this.onSave(settings);
        //this is a condition to close dialog 
        this.set('settings', null);
     });
}
,,,

Subtletree:
@dancer-cat Hmm, that line where your error occurs:
```var originBnds = this.copyRect(originator.getBoundingClientRect());```
should only run when `paper-dialog` is passed either `origin`, `closeTo ` or `parent`. If you're not passing any of those then this could be a different bug?

dancer-cat:
@Subtletree You were right!!! I didn't notice changes done by my colleague. The reason was in wrong bound paper-dialog origin. Thank you very much for your help.

PR #875, opened 2017-12-27
Add "autofocus" to autocomplete-trigger input

raghuvaran:
Add autofocus flag to passThru hash to enable autofocus of autocomplete-trigger input


miguelcobain:
Thank you. 👍

PR #879, opened 2018-01-05
fix(progress-circular): render correctly in determinate mode

### LINKED ISSUE #878: paper-progress-circular's stroke-dashoffset is NaN
image

Hi, I've encountered (probably a bug) with progress-circular, I pass it a value either string, or number between 1-100, and the problem is that the circle is ALWAYS filled, no matter what, it is filled, and when I inspected it's path element, I saw that stroke is NaN.

So I was meddling with the component inside node_modules, and value is passed to it, and on the line 167 the value itself is already NaN.

```
{{paper-progress-circular diameter=100 value=progress}}
{{minutes-to-hours summarizedDays.worked}} / {{minutes-to-hours summarizedDays.target}}
<br> {{progress}}% {{yield}}
```
progress here is numerical, but I've  been trying to pass it a raw "20" for instance too.
```  
"ember-paper": "^1.0.0-beta.2",
"ember-cli": "^2.16.2",
"ember-data": "2.17.0",

ember is 2.17.0
```
I've checked progress-cilcular on different versions of ember i.e same one that 
http://miguelcobain.github.io/ember-paper/#/addons stands on, and ember paper 1.0.0.-beta.4 and 0.2.5

xomaczar:
Fixes #878

### DIFF
--- a/addon/components/paper-progress-circular.js
+++ b/addon/components/paper-progress-circular.js
@@ -59,7 +59,7 @@ export default Component.extend(ColorMixin, {
     return mode === MODE_DETERMINATE || mode === MODE_INDETERMINATE ? `md-mode-${mode}` : 'ng-hide';
   }),
 
-  isIndeterminate: equal('mode', MODE_INDETERMINATE),
+  isIndeterminate: equal('mode', MODE_INDETERMINATE).readOnly(),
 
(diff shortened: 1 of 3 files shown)


xomaczar:
@miguelcobain all tests pass now. I can squash if you prefer.

xomaczar:
Anything else you’d like to see on this one

PR #888, opened 2018-01-30
Add rel to paper-button attribute bindings

Subtletree:
If using `{{#paper-button href="..." target="_blank"}}` we need to pass through `rel="noopener noreferrer"` too.

https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/


miguelcobain:
Thanks!

PR #891, opened 2018-02-15
Allow higher versions of EPS and EBD

cibernox:


cibernox:
I didn't expect anything to break 😕 
I'll investigate it.

cibernox:
@miguelcobain I believe the failure in beta/canary is unrelated

miguelcobain:
Thanks for the PR. CI failures are indeed unrelated. 👍

PR #895, opened 2018-03-09
Fix paper-select content positioning when scrollable

### LINKED ISSUE #773: Select options appear on top of screen when too many options
When adding more than 5 options to the select component and selecting the ones at the bottom, the options box re render far top where the component is. The issue is present on the dummy app.

paper-select-issue

panthony:
This PRs fix #773.

The issue was that at the time where `scrollHeight` and `offsetHeight` were retrieved the container was still hidden.

Because of this both values where always 0.

Moving the statement below fixes the issue because the `computePosition` manually set `display:block`  on the container.

```js
    // Get the selectMenuRect *after* max-width is possibly set above
    containerNode.style.display = 'block';
    let selectMenuRect = selectNode.getBoundingClientRect();
    let centeredRect = getOffsetRect(centeredNode);

```


miguelcobain:
Thanks @panthony. Nice find!

PR #896, opened 2018-03-09
fix(paper-autocomplete): content positioning when body is relatively positioned

### LINKED ISSUE #851: Paper-autocomplete suggestion list is not fixed to the bottom of the autocomplete input
The suggestion list position (md-virtual-repeat-container) is not fixed to the bottom of the autocomplete input, as I continued to use paper-autocomplete multiple times down the page. you can see the behaviour in the below gif.

auto complete issue

panthony:
May fix #851 

I had a similar issue with `paper-autocomplete` that simply made it unusable on mobile device because the dropdown menu where hidden behind the keyboard and if you scroll the menu would just continue moving down out of reach.

Looking at the code I saw the following differences in `calculate-ac-position` with the most recent power select.

Since the base styles apply `position:relative;` on the body I gave it a try and it fixed my issue.

### DIFF
--- a/addon/utils/calculate-ac-position.js
+++ b/addon/utils/calculate-ac-position.js
@@ -90,6 +90,16 @@ export function calculateWormholedPosition(trigger, content, destination, { hori
 
   // Calculate vertical position
   let triggerTopWithScroll = triggerTop + scroll.top;
+
+  /**
+   * Fixes bug where the dropdown always stays on the same position on the screen when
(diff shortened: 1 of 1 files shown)


panthony:
Unfortunately this is not enough :sob:

It appeared fixed but it was because I left the `position:static` on the body as a test.

Not sure why this is relative in the first place though. And why it works on ember-paper documentation.

PR #898, opened 2018-03-11
Add new addon by betocantu93

betocantu93:
This is a component I've been using lately in some corber projects, it has work to do, but, I think people can benefit from what it is at this point.


Subtletree:
Addon looks awesome, might have to put it to use, cheers!

grodriguez85:
Excellent addon, I use Ember Paper on mobile apps and we definitly need more addons like this one.

PR #900, opened 2018-03-18
component:paper-radio-group select() should handle 0 properly

### LINKED ISSUE #899: paper-radio: arrow keys do not work properly in some cases
When a `group.radio` has a value of `0`, the `select` function of `component:paper-radio-group` does not work properly causing the arrow keys to behave oddly.

chbonser:
Resolves #899


miguelcobain:
CI failures are unrelated. Thanks for the PR. Nice catch.

PR #901, opened 2018-03-19
Removed null value for defaultHighlighted

Houndsto0th:
This property currently prevents the user from allowing a default highlight to occur as it does in power-select. By Removing this declaration a user can now pass in a value for `defaultHighlighted` on the component declaration.


miguelcobain:
@Houndsto0th if you pass `defaultHighlighted` to paper-autocomplete, it will overwrite that property. Isn't that the same?

Houndsto0th:
Actually, if you pass that property to autocomplete it won't overwrite that property (at least in my experimenting). If someone can get that to work I'll just close this up

miguelcobain:
Please reopen if something is not working as expected.

PR #908, opened 2018-04-14
Ember-css-transition upgrade broke sidenav

piotrpalek:
The 0.1.11v of ember-css-transitions seems to have removed it dependency on jquery, which also means the code below isn't working anymore (due to `parent()` being a jquery API)

This PR should fix it.

### DIFF
--- a/addon/components/paper-backdrop.js
+++ b/addon/components/paper-backdrop.js
@@ -29,7 +29,7 @@ export default Component.extend(TransitionMixin, {
   }),
 
   addDestroyedElementClone(original, clone) {
-    original.parent().append(clone);
+    original.parentNode.appendChild(clone);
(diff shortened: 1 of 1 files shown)


miguelcobain:
@piotrpalek nice catch!

I think we should use this instead to increase compatibility.
```js
original.parentNode.appendChild(clone);
```

Can you update the PR if you agree?

piotrpalek:
@miguelcobain np, I've updated the PR. By compatibility you mean browser support right? If so then yeah I agree :+1: 
I was confused for a bit and was researching the differences between `parentNode` and `parentElement` in terms of how they work :man_facepalming:

miguelcobain:
Thanks. 👍

PR #911, opened 2018-04-16
fix implicit creation of `validationErrorMessages` computed property in `validation-mixin`

wahhaj:
[This deprecation](https://deprecations-app-prod.herokuapp.com/deprecations/v3.x/#toc_ember-meta-descriptor-on-object) in the latest version of ember causes the `validationErrorMessages` computed property to never update after it's initialized. Using `defineProperty` fixes this issue.


miguelcobain:
Nice find! Thanks.

PR #912, opened 2018-04-24
Add title attribute to paper-item

raghuvaran:


raghuvaran:
Yes accessibility. Besides that, it's also handy to include a brief about the item that could visibily displayed as tooltip and also read by screen readers.

Issue #631, opened 2017-01-25
is virtual-repeat ready to be used?


v3ss0n:
I am trying to get virtual repeat to work. No errors but i can't get anything to actually display.
Is virtual repeat usable without demo classes on  demo site??
```
{{#paper-list}}
  {{#paper-virtual-repeat model
   as |visibleItems|}}
    {{#each visibleItems as |item|}}

          <div class="md-list-item-text flex">
            <h3>{{item.user}}</h3>
            <h4>{{item.message}}</h4>
          </div>

    {{/each}}
  {{/paper-virtual-repeat}}

{{/paper-list}}
```
![screenshot_20170125_231155](https://cloud.githubusercontent.com/assets/419606/22300253/47834a4c-e355-11e6-96b1-d07b4912a65f.png)

v3ss0n:
inside HTML there are elements there , but it is just not showing. Tried all the demos , reduced down to bare metal and still dosen't work. Give the same set of CSS as Demo and no. 
I am on ember 1.0.14.

v3ss0n:
Found the problem , it dosen't work with percent based layouts
work around is setting virtual-repeat to a class with defined height or width depending on direction :
for percent based , this works
```
.container-main {
  height: 100vh;
  max-height: 100%;
}
```

v3ss0n:
another problem 
Horizontal repate won't work unless virtual-repate classes are altered.

jessicabetonico:
Hi, i have the same problem. Has anyone managed to solve it?

miguelcobain:
I'm inclined to remove it. I feel like users can easily being their own virtual repeat component. There's nothing "material" in it.

Subtletree:
Yeah, using something dedicated to the task sounds better than building something specifically for ember paper

Check out https://github.com/runspired/smoke-and-mirrors

v3ss0n:
@miguelcobain  i agree with your decison , there are many other virtual repate libraries out there .

panthony:
I need something similar to this, shall I add smoke-and-mirrors (or other) instead of investing time in the virtual-repeat component ?

Is there a decision taken on that point ?

miguelcobain:
@panthony please don't use virtual-repeat. I'm inclined to drop it from ember-paper at some point, at least as a public component, certainly.

v3ss0n:
@panthony  please let me know how smoke and mirror works in e-paper . I haven't test it.

miguelcobain:
[...]

PR #914, opened 2018-04-25
Upgrade to virtual-each 0.6.0

bjornharrtell:


bjornharrtell:
Realize now that yarn.lock need an update for this. Will fix ASAP.

Subtletree:
There's #906 open too, removing virtual-repeat so might not need this one.

bjornharrtell:
Ah that seems nice, but I can't fully understand the changes in that PR.

My reason for doing this is to get rid of Babel 5 usage in ember-paper which I think is only because virtual-each 0.5.x at this point.

Subtletree:
Yeah I've been getting that deprecation too.

I think the idea is to let users handle `virtual-repeat`/ infinite-scroll themselves, using addons specifically designed for the task. https://github.com/miguelcobain/ember-paper/issues/631#issuecomment-288748440

bjornharrtell:
Sounds reasonable, but it looks like the PR is not complete. So I will attempt to get this passing.

bjornharrtell:
Unfortunately I now notice that there are several other deps that are out of date. AFAICT these are `ember-deferred-content`, `liquid-fire`, `ember-cp-validations`, `ember-transition-helper` and `ember-cli-fastboot`. Will attempt upgrading them in a separate PR.

PR #920, opened 2018-05-09
add 'paper-menu-content' example for 'dense'

panthony:
This PR simply add an exemple in the dummy app with the `dense:true` value.

Turned out I implemented a similar feature using custom css when option was there all along but undocumented.


miguelcobain:
Thank you @panthony!

PR #924, opened 2018-05-11
fix css issue when dense=true and content is scrollable

### LINKED ISSUE #922: paper-menu-item: css issue when scrollable with `dense=true`
When you use the `dense:true` option and the menu's content is scrollable then there is a display issue:

capture d ecran 2018-05-11 a 08 30 16

The `md-menu-item` does not respect the height of 32px.

Its default css is

```css
md-menu-content.md-dense md-menu-item {
    height: 32px;
    min-height: 0px;
}
```

Setting `min-height` to `32px` fixes the issue.

Should it be patched in `ember-paper`?

panthony:
Closes #922

### DIFF
--- a/app/styles/ember-paper.scss
+++ b/app/styles/ember-paper.scss
@@ -158,3 +158,4 @@
 
 // some overrides that are needed for ember-paper (avoid if possible)
 @import './overrides/paper-autocomplete';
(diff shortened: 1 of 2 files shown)


miguelcobain:
Thanks! 🎉

PR #933, opened 2018-05-24
ENHANCE_paper-select-documentation - Restoring example of paper-selec…

swgamerx:
Found and restored old example of a searchable paper-select. Added documentation on paper select fields.


miguelcobain:
@swgamerx Thanks. I'm not sure anymore why this example was removed, but it seems to work fine. 👍

PR #943, opened 2018-06-11
Fix setup of defaultAttrs for paper-virtual-repeat

### LINKED ISSUE #926: Error paper-chips autocomplete with version 1.0.0-beta.7 and ember 3
"Assertion Failed: You attempted to access `defaultAttrs.positionIndex` (on `<(unknown mixin):ember1386>`), but `defaultAttrs` is a computed property.↵↵Due to certain internal implementation details of Ember, the `defaultAttrs` property previously contained a private "descriptor" object, therefore `defaultAttrs.positionIndex` would have been `undefined`.↵↵This implementation detail has now changed and the "descriptor" object is no longer present at this location. Soon, accessing `defaultAttrs` on this object will return the computed property's current value (see RFC #281 for more details).↵↵If you are seeing this error, you are likely using an addon that relies on this now-defunct private implementation detail. If you can, identify the addon from the stack trace below and report this bug to the addon authors. If you feel stuck, the Ember Community Slack (https://ember-community-slackin.herokuapp.com/) may be able to offer some help.↵↵If you are an addon author and need help transitioning your code, please get in touch in the #dev-ember channel in the Ember Community Slack.

panthony:
Closes #926

The `defaultAttrs` was probably wrapped in a computed property because of the new "leaking object state" eslint error.

Since creating a computed property poses an issue here, let's simply setup it in `init()` lifecycle method.

### DIFF
--- a/addon/components/paper-virtual-repeat.js
+++ b/addon/components/paper-virtual-repeat.js
@@ -1,5 +1,6 @@
 import { mapBy } from '@ember/object/computed';
 import { run } from '@ember/runloop';
+import { assign } from '@ember/polyfills';
 import { observer, set, get, computed } from '@ember/object';
 import RSVP from 'rsvp';
 import { A as emberArray } from '@ember/array';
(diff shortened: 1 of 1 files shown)


miguelcobain:
Glad to finally have a fix for this! Thanks.

PR #944, opened 2018-06-11
Update tests using new ember test API

panthony:


panthony:
Once this PR is merged, the follow up could be the following codemod:

https://github.com/simonihmig/ember-test-helpers-codemod

In order to replace lot's of jQuery usage in tests by the new ember test helpers.

panthony:
@miguelcobain I know you said you cleared the cache but is it really a good idea to use "--skip-cleanup" from ember-try if the built files are cached?

> When running in a CI environment where changes are discarded you can skip resetting your environment back to its original state by specifying --skip-cleanup=true as an option to ember try. Warning: If you use this option and, without cleaning up, build and deploy as the result of a passing test suite, it will build with the last set of dependencies ember try was run with.

Still not sure how to fix this issue

panthony:
@miguelcobain There is something wrong with travis.

I updated the `paper-select-test` (d5f0e57) just to see if it would still complains about "wait" that I removed and it now complains about usage of jQuery that I just removed:

```
message: >
            integration/components/paper-select-test.js should pass ESLint
            
            87:9 - 'wait' is not defined. (no-undef)
            91:15 - '$' is not defined. (no-undef)
            91:15 - Do not use global `$` or `jQuery` (ember/no-global-jquery)
```

miguelcobain:
> @miguelcobain I know you said you cleared the cache but is it really a good idea to use "--skip-cleanup" from ember-try if the built files are cached?

--skip-cleanup is what the standard ember-cli blueprint for addons uses. It makes sense. We don't need to cleanup dependencies for a particular scenario once tests pass. That would reinstall the default dependencies and we wouldn't be using them anymore.

I think travis caches by scenario, I guess.

miguelcobain:
Thanks for this great update, @panthony !

PR #945, opened 2018-06-11
paper-progress-circular:  do not rely ember rAF/cAF from ember-css-transitions

### LINKED ISSUE #894: paper-progress-circular does not play well with ember-qunit 3.X
With `ember-qunit` a new way of writing integrations tests was added using `setupRenderingTest` and a new `render` helper.

```js
module('Integration | Component | component-name', function(hooks) {

  setupRenderingTest(hooks);

  test('it renders', async function(assert) {
    await render(hbs`{{component-name}}`);
  });
});
```

If `component-name` includes a `paper-progress-circular` then the `await render()` method will never return.

The theory would be because `paper-progress-circular` never stop adding tasks to the render queue.

panthony:
Closes #894 

See ticket for full explanation.


miguelcobain:
Thanks a lot @panthony !

PR #951, opened 2018-06-30
Add `ember-paper-modals-manager` description to `addon`-page

onechiporenko:
Hi, @miguelcobain. Thanks for your work on `ember-paper`! 
I've created an addon for modals https://github.com/onechiporenko/ember-paper-modals-manager. I hope it will be useful for others.


miguelcobain:
Interesting work you have there. Thanks!

Issue #918, opened 2018-05-03
Got build error after installing paper-ember and serving ember


aaronucsd:
after `yarn add -D ember-paper` and ember serve. Got this error

Build Error (AutoprefixerFilter) in assets/test-support.css:472:1

/Usersxxxx/projects/data-ui/pinot/txxxx/assets/test-support.css:472:1: Unknown word
}
//# sourceMappingURL=test-support.css.map
^

aaronucsd:
changing to below did not work.


$colors: (
  'red': $color-red,
  'pink': $color-pink,
  'purple': $color-purple,
  'deep-purple': $color-deep-purple,
  'indigo': $color-indigo,
  'blue': $color-blue,
  'light-blue': $color-light-blue,
  'cyan': $color-cyan,
  'teal': $color-teal,
  'green': $color-green,
  'light-green': $color-light-green,
  'lime': $color-lime,
  'yellow': $color-yellow,
  'amber': $color-amber,
  'orange': $color-orange,
  'deep-orange': $color-deep-orange,
  'brown': $color-brown,
  'blue-grey': $color-blue-grey,
  'grey': $color-grey,
  'shades': $shades,
  'foreground-light': $foreground-light,
  'foreground-dark': $foreground-dark,
  'background-light': $background-light,
);

wjk:
I've come up with a workaround for this issue, but I don't know how much it will break.

You need to apply the following patch to the file `app/styles/color-palette.scss`:

```diff
 // returns color shade given a color palette name or a palette map
 @function color($color, $type: null) {
   $curr_color: null;
   // color is a string case
   @if type-of($color) == 'string' and map-has-key($colors, $color) {
     $curr_color: map-get($colors, $color);
   } @else {
     // color is a map
     $curr_color: $color;
   }
 
+  @if type-of($curr_color) != 'map' {
+    @return $curr_color;
+  }
 
   // if no type was provided:
   //   - use 'default' key if present
   //   - use '500' from $default-color
   @if ($type == null) {
     @if map-has-key($curr_color, 'default') {
       $type: map-get($curr_color, 'default');
     } @else {
       $type: $default-color;
     }
   }
 
   @return map-get($curr_color, $type);
 }
 
 // returns the correct contrast color for the given color
 @function contrastColor($color, $type: null) {
   $curr_color: null;
   // color is a string case
   @if type-of($color) == 'string' and map-has-key($colors, $color) {
     $curr_color: map-get($colors, $color);
[...]

PR #952, opened 2018-07-02
Add workaround for Sass syntax error

wjk:
See #918 for context. Note that this PR is a workaround only. I do not know how much breakage it will cause if it is merged. On the other hand, ember-paper doesn't build or run _at all_ without this patch, so we can't get any worse than that, right? 😄

### DIFF
--- a/app/styles/color-palette.scss
+++ b/app/styles/color-palette.scss
@@ -496,6 +496,10 @@ $default-color: '500' !default;
     $curr_color: $color;
   }
 
+  @if type-of($curr_color) != 'map' {
+    @return $curr_color;
+  }
+
   // if no type was provided:
   //   - use 'default' key if present
   //   - use '500' from $default-color
@@ -519,6 +523,10 @@ $default-color: '500' !default;
   } @else {
     // color is a map
     $curr_color: $color;
+  }
+
+  @if type-of($curr_color) != 'map' {
+    @return $curr_color;
   }
 
   // if no type was provided:


miguelcobain:
This shouldn't be happening now in latest versions.
Please reopen if still relevant.

ike:
Still happening for me. I was pinned to `wjk:fix-sass-syntax-error` because of this. I just tried building off of `v1.0.0-beta.18`. Same error. 😢

miguelcobain:
@ike or @wjk can any of you provide me instructions to reproduce this issue?

I've installed latest ember-paper in a new app and got no issues.
It would be great if you could create a repo or provide instructions for me to recreate the issue, please.

wjk:
@miguelcobain I've only interacted with ember-paper through my company, and we have since decided to migrate off of it due to unrelated issues. Therefore, I am afraid I cannot offer any more assistance on this matter.

miguelcobain:
@wjk Do you mind sharing the reasons for your migration? We want to get better. :)
Thank you.

@ike A reproduction of this issue would be greatly appreciated.

wjk:
Quite simply, ember-paper was far too inflexible, both for our UI designer's tastes and for our sanity as developers. In particular, creating a custom color scheme to match our company's branding was nigh-impossible.

I was the one who first looked into ember-paper when a migration to it was announced, and reported the difficulty of making it work as we wanted it; my company's management then made the executive decision to abandon the migration and keep our existing hand-written framework. I am afraid I don't know any more than this, as I was not privy to those discussions.

miguelcobain:
@wjk I see. We're in a somewhat similar situation here.
Do you think any other material design addon/library would perform better in this situation?

In other words, I think the "problem" is material design itself. We can't have material design *and* our own custom design at the same time.

If a custom design jumps, suddenly ember-paper/material design starts being an issue because we constantly need to be overriding styles.

Thank you for your clarification.

ike:
Hey @miguelcobain -- thanks so much for the response! I appreciate it. 

I am not positive how to replicate the issue -- my gut is that it has something to do with other styles that may be conflicting with the ember-paper sass files. For instance, we also have Bootstrap 3 installed on this application, so perhaps there are conflicts between those?

Would it be best to try to replicate in a newly generated ember project?

miguelcobain:
@ike that would be perfect!

ike:
Awesome! Here's a repo that has the issue replicated. Let me know if it doesn't show up for you. Error should read:

```
Sass Syntax Error (SassCompiler) in app/styles/default-theme.scss:103:53

Error: argument `$color` of `red($color)` must be a color
        on line 103 of tmp/sass_compiler-input_base_path-9l1piwU6.tmp/app/styles/default-theme.scss, in function `red`
        from line 103 of tmp/sass_compiler-input_base_path-9l1piwU6.tmp/app/styles/default-theme.scss, in mixin `paper-theme-variables`
        from line 117 of tmp/sass_compiler-input_base_path-9l1piwU6.tmp/app/styles/default-theme.scss
        from line 64 of tmp/sass_compiler-input_base_path-9l1piwU6.tmp/app/styles/ember-paper.scss
        from line 3 of tmp/sass_compiler-input_base_path-9l1piwU6.tmp/app/styles/app.scss
>> #{$paletteName}-#{$hue}: #{red($color)}, #{green($color)}, #{blue($color)};

   ------------------------------------------^



Stack Trace and Error Report: /var/folders/pr/pb_1447s2k1780jm_mm271vm0000gn/T/error.dump.2dad8bfe080dd41db0d4186b2508cf7e.log
```

There are some `ember-cli-babel` deprication warnings, but as far as I can tell, they shouldn't be related to the bug? Of course, I could be wrong. 

Thank you SO MUCH for taking the time.

PR #959, opened 2018-07-19
Make paper-toast test friendly

gmurphey:
Timers were not being cleaned up when using `paper-toast` in a rendering test.


miguelcobain:
@gmurphey cancelling the timer wasn't enough to speed up the tests?

panthony:
Ins't there a less intrusive way to setup the duration for tests? In the service for instance as default configuration in test env?

miguelcobain:
@panthony we can't make the change just on the service because a paper-toast can be used without the service, and there are tests for that. So the short duration wouldn't be picked up.

paper-toaster service is already picking up options from the environment config: https://github.com/miguelcobain/ember-paper/blob/master/addon/services/paper-toaster.js#L49

We should do the same for paper-toast component options: https://github.com/miguelcobain/ember-paper/blob/master/addon/components/paper-toast.js#L22-L27

Then this would be a matter of using a config with a short duration just on the test env.

We can keep the timer cancellation on `willDestroyElement`, even though we properly guard in the handler using `this.isDestroyed`.

panthony:
@miguelcobain Didn't think of the usage of paper-toast directly, nevermind then :)

PR #964, opened 2018-07-24
make ripple touch events passive

xomaczar:
Improves scrolling and fixes warnings

### DIFF
--- a/addon/components/paper-tooltip.js
+++ b/addon/components/paper-tooltip.js
@@ -7,6 +7,7 @@ import { getOwner } from '@ember/application';
 import layout from '../templates/components/paper-tooltip';
 import $ from 'jquery';
 import getParent from 'ember-paper/utils/get-parent';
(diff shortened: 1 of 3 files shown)


Subtletree:
closes #870 

Good catch with the feature detection 👍.  Looks like we might need:
```
let options = supportsPassiveEventListeners ? { passive: true } : false;
```
instead of the empty object https://github.com/WICG/EventListenerOptions/blob/gh-pages/explainer.md#feature-detection

xomaczar:
@Subtletree thanks for the code review, sorry I did not see that there was an open PR handling similar use case. I've updated my PR with your suggestions.

xomaczar:
^ @miguelcobain

miguelcobain:
Indeed, the feature detection is needed to correctly use passive events.
Thank you both @xomaczar and @Subtletree.

Issue #791, opened 2017-08-21
paper-tab: disabled option partially work


panthony:
When you set `disabled=true` on a tab. It has a "disabled" look & feel but you are still able to click on it and it gets the selected state.

The `Basic Usage` should have one of its tab disabled to have an exemple of this option visible.

BennyAlex:
@miguelcobain @panthony 
When will this fixed? It's a really annyoing error

panthony:
@BennyAlex I may take a look at this eventually if nobody does but for now I'm happy with the following workaround:

```hbs
  {{#tabs.tab value=route disabled=disabled href=(if (not disabled) (href-to route))}}
     Label
  {{/tabs.tab}}
```

BennyAlex:
@panthony This will only work when using route-tabs, not when just using the index

BennyAlex:
@miguelcobain 
hey, are you able to release a bugfix?
The tabs working fine exept this bug :/

miguelcobain:
Closed via #971

PR #971, opened 2018-08-29
Fix paper tab disabled

MarcoUmpierrez:
Hi, I saw that this issue #791 hasn't been fixed yet. Here's a PR to fix it. I've added a test and an example of a disabled tab in the dummy app.

### DIFF
--- a/addon/components/paper-tab.js
+++ b/addon/components/paper-tab.js
@@ -11,7 +11,7 @@ export default Component.extend(ChildMixin, RippleMixin, FocusableMixin, {
   tagName: 'md-tab-item',
   classNames: ['md-tab'],
   classNameBindings: ['isSelected:md-active'],
-  attributeBindings: ['isSelected:aria-selected', 'href', 'style'],
+  attributeBindings: ['isSelected:aria-selected', 'style', 'maybeHref:href'],
 
   rippleContainerSelector: null,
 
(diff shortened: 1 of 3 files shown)


addon/components/paper-tab.js
```diff
@@ -57,7 +57,9 @@ export default Component.extend(ChildMixin, RippleMixin, FocusableMixin, {
   },
 
   click() {
-    this.sendAction('onClick', ...arguments);
-    this.sendAction('onSelect', this);
+    if(!this.get("disabled")) {
```
miguelcobain: @MarcoUmpierrez could you please change this line to `if (!this.get('disabled')) {` to conform with the rest of the project's coding style?

miguelcobain:
@MarcoUmpierrez this looks awesome. Thank you so much.

There's still something that isn't covered that shouldn't be too hard to do.

A tab can be an anchor if an href is passed. If a tab with an href is disabled, we need to disable a link. But how do we do that?
Well, a simple way is to just don't have any href.
So we can have an intermediate value that "maybe" holds the href if not disabled.

```js
attributeBindings: ['maybeHrefhref'],
maybeHref: computed('href', 'disabled', function() {
  if (this.get('href') && !this.get('disabled')) {
    return this.get('href');
  }
}),
```

Have a test that made sure that no href was rendered for a disabled tab would also be helpful.

MarcoUmpierrez:
@miguelcobain Added the changes you requested and a test to check if the href is set when the tab is disabled.

miguelcobain:
Thanks a lot @MarcoUmpierrez !

PR #973, opened 2018-08-31
3.4 Closure actions deprecation

betocantu93:
Deprecations Added in 3.4
Use closure actions instead of `sendAction`

https://emberjs.com/deprecations/v3.x#toc_ember-component-send-action


avdept:
Any progress on merging this?

miguelcobain:
@betocantu93 Thank you very much for this.
Would it be too much to ask if we could use [ember-invoke-action](https://github.com/martndemus/ember-invoke-action) addon?

The addon has a functional utility, just like you made here, and also has a mixin. I'd prefer the functional utility because it is better aligned with ES6 classes which don't support mixins. I'm planning to migrate ember-paper soon.

Using this addons promotes using shared solutions and it is already tested against many ember versions.
Let me know if you're ok to do it.

betocantu93:
@miguelcobain Interesting, nice addon. Excellent news about ember-paper migration, let us know if you need help with that - also I was planning an attempt to get ember-paper run without any jquery -  and sure, I can do the addon thing, I'll push to this branch I guess.

addon/components/paper-chips.js
```diff
@@ -170,7 +171,7 @@ export default Component.extend({
 
     noUnselected(old, event) {
       if (['Backspace', 'Delete', 'Del', 'ArrowLeft', 'Left', 'ArrowRight', 'Right'].includes(event.key)) {
-        this.sendAction('keyDown', event);
+        invokeAction(this, 'keyDown', item);
```
marcemira: this should be `event` instead of `item`

miguelcobain: Exactly. This is the error on travis.
Linting tests are useful after all, huh.

miguelcobain:
Thanks a lot for this!

PR #980, opened 2018-09-18
Paper-tabs Assertion Failed: You modified canPageBack twice in a single render

betocantu93:
Im getting this error
`Assertion Failed: You modified "canPageBack" twice in a single render`
in mobile chrome which prevents the smoothly transition to hidden tabs.

I tracked it down to `this.fixOffsetIfNeeded();` which is executed everytime the `_selectedTab` change which at least in my use case, it can be changed both programatically (i.e `selected=someProperty`) or by the end user when clicked on a tab or the next, previous arrows.

### DIFF
--- a/addon/components/paper-tabs.js
+++ b/addon/components/paper-tabs.js
@@ -7,7 +7,6 @@ import layout from '../templates/components/paper-tabs';
 import { ParentMixin } from 'ember-composability-tools';
 import ColorMixin from 'ember-paper/mixins/color-mixin';
 import { invokeAction } from 'ember-invoke-action';
-import { scheduleOnce } from '@ember/runloop';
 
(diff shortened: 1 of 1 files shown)


betocantu93:
I just noticed the observer `_selectedTabDidChange` is getting fired, twice... I need to track why, it might be a bug in my app code.

betocantu93:
The _selectedTab computed gets triggered N times each selected change, this leads to changing the variable twice in the single render, so changing the computed to watch selected, which in theory just changes once and finding in childs by value, will just execute the observer once.

betocantu93:
Your changes work in my app

miguelcobain:
Thanks a lot @betocantu93 !

PR #981, opened 2018-09-20
Correctly call 'paper-contrast-color'

gkappen:
Just a simple fix for the deprecated `contrastColor` function to correctly call the new `paper-contrast-color` function


miguelcobain:
This was a massive mistake on my part. Thank you for spotting it.

Issue #759, opened 2017-07-27
Paper-tabs initial render not working


betocantu93:
I kept testing the paper-tabs and found a new issue if I set the initial tab to one that is not currently visible, it doesn't scroll to the correct "tab page" so it's hidden.

```
{{#paper-card as |card| }}

	{{#paper-tabs
	 borderBottom=true
	 selected=consultation.activeTab
	 onChange=(action (mut consultation.activeTab)) as |tabs|}}
	 	{{#tabs.tab  value="consult_reason"}}
			Motivo de Consulta
	  {{/tabs.tab}}
		{{#tabs.tab  value="background"}}
		Antecedentes
	  {{/tabs.tab}}
		{{#tabs.tab  value="appys"}}
		Interrogatorio APyS
	  {{/tabs.tab}}
		{{#tabs.tab  value="physical_exploration"}}
		Exploración Física
	  {{/tabs.tab}}
		{{#tabs.tab  value="laboratory"}}
		Laboratorio y Gabinete
	  {{/tabs.tab}}
		{{#tabs.tab  value="diagnosis"}}
		Diagnostico
	  {{/tabs.tab}}
		{{#tabs.tab  value="treatment"}}
		Tratamiento
	  {{/tabs.tab}}

	{{/paper-tabs}}
	
	{{component consultation.componentName model=model errors=errors}}

{{/paper-card}}


```

Currently this happens if I initially render "treatment" tab

![image](https://user-images.githubusercontent.com/9092644/28694163-c10370c2-72ed-11e7-81c5-220805b90dc6.png)

And I expected this behaivor

![image](https://user-images.githubusercontent.com/9092644/28694177-cfbc7a78-72ed-11e7-9fed-024104ce697d.png)

Or even better to be in the "middle"

Issue #858, opened 2017-11-03
Paper tabs calculate wrong position for the inkBar on app loading


gmarquet:
go to http://miguelcobain.github.io/ember-paper/#/components/tabs (without cache enabled)
and click on the first active tab then the component should update the inkBar to the correct values.

The currentWidth of the _selectedTab seems to be wrong.
I suspect the change in width between before and after the loading of the font.

kkgomez2:
I also have this problem.

panthony:
Duplicate of #858 ?

PR #982, opened 2018-09-20
FixOffset when Selected tab is not visible on initial render

betocantu93:
`this.fixOffsetIfNeeded();` added to updateCanvasWidth, so if we resize or change orientation, we also get the offsetCorrection to the possible now hidden tabs, also, fixOffset on initial render, so if start with a selected hidden tab, (i.e routable) we fix the offset

One issue I couldn't fix so far is that somehow the `this.element.offsetLeft` calculated by each tab via updateDimensions() is broken on the initial render. It's like the `offsetLeft` is modified later when other tabs grow(?) in size or something like that... I get this... 

image

If we fix the offsetLeft, this solves #759

### DIFF
--- a/addon/components/paper-tabs.js
+++ b/addon/components/paper-tabs.js
@@ -75,6 +75,7 @@ export default Component.extend(ParentMixin, ColorMixin, {
     let updateCanvasWidth = () => {
       this.updateDimensions();
       this.updateStretchTabs();
+      this.fixOffsetIfNeeded();
     };
 
     window.addEventListener('resize', updateCanvasWidth);


miguelcobain:
Just to be clear, this PR fixes a different issue not related with #759, right?

betocantu93:
This PR fixes #759, I actually made this PR to fix it. But it needs a little bit of work with the misplaced inkbar on the first render.

miguelcobain:
@betocantu93 is this PR still needed after merging #1036?

betocantu93:
@miguelcobain hmm im using master to test this, and this issue is still there, and right now it seems the arrows for next or previous page of tabs are broken, I can't navigate to -> even there's more tabs. 

image

For demo purposes I added a couple of more tabs and tested it, as you can see in the picture the left arrow to next page is disabled, and there's like 5 more tabs.

Also look at the border bottom is misplaced.

panthony:
There is also this issue about initial rendering issue with paper-tabs: #858

panthony:
> this.fixOffsetIfNeeded(); added to updateCanvasWidth, so if we resize or change orientation, we also get the offsetCorrection to the possible now hidden tabs

@betocantu93 This particular change, which is already on master, completely breaks the pagination because:

- You click prev -> offset update -> tab partially hidden (because of the pagination) -> didRender trigger -> fixOffset -> tab partially hidden which change the offset to make it visible again

I did my best to make it work in most cases (ex: tabs wider than canvas, ...) with #1048 but "fixOffsetIfNeeded" just cannot be called within 'didRender' ..

miguelcobain:
Fixed by #1048

PR #986, opened 2018-10-15
Paper autocomplete text

jeetdas:
The `text` property should clear and not show the last searched (`searchText`) text if the selected property has been cleared.

### DIFF
--- a/addon/components/paper-autocomplete-trigger.js
+++ b/addon/components/paper-autocomplete-trigger.js
@@ -25,7 +25,7 @@ export default Component.extend({
     );
   }),
 
-  text: computed('select.selected}', function() {
+  text: computed('select.selected', function() {
     let selected = this.get('select.selected');
     if (selected) {
       return this.getSelectedAsText();
--- a/addon/components/paper-autocomplete-trigger.js
+++ b/addon/components/paper-autocomplete-trigger.js
@@ -25,12 +25,12 @@ export default Component.extend({
     );
   }),
 
-  text: computed('select.{searchText,selected}', function() {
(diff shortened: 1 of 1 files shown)


xomaczar:
API provided by `paper-autocomplete` allows for this scenario, however this not the default behavior defined in MD spec. Consider UX when a user using only a keyboard attempts to find and select a value from the list `[Bar, Baz, Foo]`:

1. User types `Ba`, the `paper-autocomplete` shows 2 options `[Bar, Baz]`
2. User presses &darr; (`arrow-down`) key, which highlights `Baz` option
3. User presses &crarr; (`enter`) key, `Baz` option gets selected, cursor is still active in trigger
4. User decides to select a different option, and presses a backspace key. 
5. The state should be `selected=null, searchText='Ba'` and two options `[Bar, Baz]` are shown in the dropdown.

jeetdas:
@xomaczar Good point, I didn't think through the user-experience. This came about because I was trying to clear the `paper-autocomplete` text and wasn't sure why the `searchText` was being returned. I was using something else to highlight/bold the searchText and didn't consider the UX part. I'm still not sure what to do if I want to clear the field after a user has clicked submit or something like that.

jeetdas:
Not exactly sure how to do that, could you elaborate on it or provide an example?

jeetdas:
@xomaczar ^

xomaczar:
@jeetdas 
https://codesandbox.io/s/github/xomaczar/ember-new-output/tree/example-ember-paper-reset-search/

jeetdas:
@xomaczar Thank you!

PR #988, opened 2018-10-15
paper-dialog no jquery

betocantu93:
Hello, I refactored jquery from the component and from the component integration tests, it seems to be working ok. I have three tests failing though.

1. `test('should only prevent scrolling behind scoped modal')`
2. `test('pressing escape triggers close action`)` this was failing even without the refactor.
3. `test('opening gives focus')`

### DIFF
--- a/addon/components/paper-dialog-inner.js
+++ b/addon/components/paper-dialog-inner.js
@@ -24,7 +24,7 @@ export default Component.extend(Translate3dMixin, {
         let focusableButtons = this.element.querySelectorAll('md-dialog-actions button');
         toFocus = focusableButtons[focusableButtons.length - 1];
       }
-      if(toFocus) {
+      if (toFocus) {
         toFocus.focus();
       }
     }
@@ -41,18 +41,18 @@ export default Component.extend(Translate3dMixin, {
     this.checkContentOverflow();
     // content overflow might change depending on load of images inside dialog.
     let imageElements = this.element.querySelectorAll('img');
-    this._checkContentOverflowOnLoad = run.bind(this, this.checkContentOverflow) 
-    imageElements.forEach( image => {
-      image.addEventListener('load', this._checkContentOverflowOnLoad)
+    this._checkContentOverflowOnLoad = run.bind(this, this.checkContentOverflow)
+    imageElements.forEach((image) => {
+      image.addEventListener('load', this._checkContentOverflowOnLoad);
     });
-  
+
   },
 
   willDestroyElement() {
     this._super(...arguments);
     let imageElements = this.element.querySelectorAll('img');
-    imageElements.forEach( image => {
-      image.removeEventListener('load', this._checkContentOverflowOnLoad)
+    imageElements.forEach((image) => {
+      image.removeEventListener('load', this._checkContentOverflowOnLoad);
     });
     this._checkContentOverflowOnLoad = null;
   },
--- a/addon/components/paper-dialog-inner.js
+++ b/addon/components/paper-dialog-inner.js
@@ -41,8 +41,9 @@ export default Component.extend(Translate3dMixin, {
     this.checkContentOverflow();
     // content overflow might change depending on load of images inside dialog.
     let imageElements = this.element.querySelectorAll('img');
+    this._checkContentOverflowOnLoad = run.bind(this, this.checkContentOverflow) 
     imageElements.forEach( image => {
-      image.addEventListener('load', run.bind(this, this.checkContentOverflow))
+      image.addEventListener('load', this._checkContentOverflowOnLoad)
     });
   
   },
@@ -51,8 +52,9 @@ export default Component.extend(Translate3dMixin, {
     this._super(...arguments);
     let imageElements = this.element.querySelectorAll('img');
     imageElements.forEach( image => {
-      image.removeEventListener('load', run.bind(this, this.checkContentOverflow))
+      image.removeEventListener('load', this._checkContentOverflowOnLoad)
     });
+    this._checkContentOverflowOnLoad = null;
   },
 
   checkContentOverflow() {
--- a/addon/components/paper-dialog.js
+++ b/addon/components/paper-dialog.js
@@ -1,8 +1,6 @@
 /**
  * @module ember-paper
  */
-import { inject as service } from '@ember/service';
-
 import { or } from '@ember/object/computed';
 import Component from '@ember/component';
 import { computed } from '@ember/object';
@@ -63,15 +61,13 @@ export default Component.extend({
     return document.querySelector(this.get('destinationId'));
   }),
 
-  constants: service(),
-
   didInsertElement() {
     this._super(...arguments);
     if (this.get('escapeToClose')) {
 
       this._destinationEle = document.querySelector(this.get('destinationId'));
       this._onKeyDown = (e) => {
-        if (e.keyCode === this.get('constants.KEYCODE.ESCAPE') && this.get('onClose')) {
+        if (e.keyCode === 27 && this.get('onClose')) {
           invokeAction(this, 'onClose');
         }
       }
@@ -82,8 +78,9 @@ export default Component.extend({
 
   willDestroyElement() {
     this._super(...arguments);
-    if (this.get('escapeToClose')) {
+    if (this.get('escapeToClose') && this._destinationEle) {
       this._destinationEle.removeEventListener('keydown', this._onKeyDown);
+      this._onKeyDown = null;
     }
   },
 
--- a/addon/components/paper-dialog.js
+++ b/addon/components/paper-dialog.js
@@ -41,7 +41,7 @@ export default Component.extend({
       return '#ember-testing';
     }
     let parent = this.get('defaultedParent');
-    let parentEle = document.querySelector(parent);
+    let parentEle = typeof parent === "string" ? document.querySelector(parent) : parent;
     // If the parentEle isn't found, assume that it is an id, but that the DOM doesn't
     // exist yet. This only happens during integration tests or if entire application
     // route is a dialog.
--- a/addon/components/paper-dialog.js
+++ b/addon/components/paper-dialog.js
@@ -4,7 +4,6 @@
 import { inject as service } from '@ember/service';
 
 import { or } from '@ember/object/computed';
-import $ from 'jquery';
 import Component from '@ember/component';
 import { computed } from '@ember/object';
 import { getOwner } from '@ember/application';
@@ -42,20 +41,21 @@ export default Component.extend({
       return '#ember-testing';
     }
     let parent = this.get('defaultedParent');
-    let $parent = $(parent);
-    // If the parent isn't found, assume that it is an id, but that the DOM doesn't
+    let parentEle = document.querySelector(parent);
+    // If the parentEle isn't found, assume that it is an id, but that the DOM doesn't
     // exist yet. This only happens during integration tests or if entire application
     // route is a dialog.
-    if ($parent.length === 0 && parent.charAt(0) === '#') {
+    if (!parentEle && parent.charAt(0) === '#') {
       return `#${parent.substring(1)}`;
     } else {
-      let id = $parent.attr('id');
+      let id = parentEle.getAttribute('id');
       if (!id) {
         id = `${this.elementId}-parent`;
-        $parent.get(0).id = id;
+        parentEle.setAttribute('id', id);
       }
       return `#${id}`;
     }
+
   }),
 
   // Find the element referenced by destinationId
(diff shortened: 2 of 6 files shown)


tests/integration/components/paper-dialog-test.js
```diff
@@ -1,8 +1,7 @@
 import { module, test } from 'qunit';
 import { setupRenderingTest } from 'ember-qunit';
-import { render, settled } from '@ember/test-helpers';
-import hbs from 'htmlbars-inline-precompile';
-import $ from 'jquery';
+import { render, settled, find, findAll, click } from '@ember/test-helpers';
+import hbs from 'htmlbars-inline-precompile';ç
```
panthony: Trailing `ç`

tests/integration/components/paper-dialog-test.js
```diff
@@ -35,7 +35,7 @@ module('Integration | Component | paper dialog', function(hooks) {
       {{/paper-dialog}}
     `);
 
-    let dialogContent = $('md-dialog').html().trim();
+    let dialogContent = find('md-dialog').innerHTML.trim();
```
panthony: With `qunit-dom` (already a dependency) you can write `assert.dom('md-dialog').asText('Lorem ipsum.')`

tests/integration/components/paper-dialog-test.js
```diff
@@ -57,9 +57,9 @@ module('Integration | Component | paper dialog', function(hooks) {
       {{/paper-dialog}}
     `);
 
-    assert.ok(!$('#paper-wormhole md-dialog').length, 'did not render in default');
-    assert.ok($('#sagittarius-a md-dialog').length, 'rendered in parent');
-    assert.ok($('#ember-testing md-dialog').length, 'still rendered in ember-testing');
+    assert.ok(!find('#paper-wormhole md-dialog'), 'did not render in default');
+    assert.ok(find('#sagittarius-a md-dialog'), 'rendered in parent');
+    assert.ok(find('#ember-testing md-dialog'), 'still rendered in ember-testing');
```
panthony: You can do `assert.dom(selector).exists()`

tests/integration/components/paper-dialog-test.js
```diff
@@ -77,7 +77,7 @@ module('Integration | Component | paper dialog', function(hooks) {
       {{paper-dialog}}
     `);
 
-    assert.ok($('md-backdrop').hasClass('md-opaque'), 'backdrop is opaque');
+    assert.ok(find('md-backdrop').classList.contains('md-opaque'), 'backdrop is opaque');
```
panthony: You can do `assert.dom('md-backdrop').hasClass('md-opaque')`

tests/integration/components/paper-dialog-test.js
```diff
@@ -171,18 +171,18 @@ module('Integration | Component | paper dialog', function(hooks) {
       {{/paper-dialog}}
     `);
 
-    assert.ok($('md-dialog').length, 'dialog is showing');
+    assert.ok(find('md-dialog'), 'dialog is showing');
 
-    $('#the-button').mousedown().mouseup().click();
-    assert.ok($('md-dialog').length, 'dialog is still showing');
+    click(find('#the-button'));
```
panthony: Click can take the selector directly, no need to call 'find'

tests/integration/components/paper-dialog-test.js
```diff
@@ -229,19 +230,19 @@ module('Integration | Component | paper dialog', function(hooks) {
       </button>
     `);
 
-    $('#theorigin').focus();
-    assert.equal(document.activeElement, $('#theorigin').get(0));
-    $('#theorigin').click();
+    focus('#theorigin')
+    assert.equal(document.activeElement, find('#theorigin'));
```
panthony: `assert.dom('#theorigin').isFocused()`

betocantu93:
@panthony I refactored to use qunit-dom and fix the broken tests

tests/integration/components/paper-dialog-test.js
```diff
@@ -145,9 +140,9 @@ module('Integration | Component | paper dialog', function(hooks) {
       {{/if}}
     `);
 
-    assert.ok($('md-dialog').length, 'dialog is showing');
+    assert.dom('md-dialog').exists();
 
-    $('.md-dialog-container').mousedown().mouseup().click();
+    click('.md-dialog-container');
```
panthony: `await click()`

tests/integration/components/paper-dialog-test.js
```diff
@@ -156,10 +151,10 @@ module('Integration | Component | paper dialog', function(hooks) {
       {{paper-dialog}}
     `);
 
-    assert.ok($('md-dialog').length, 'dialog is showing');
+    assert.dom('md-dialog').exists();
 
-    $('.md-dialog-container').mousedown().mouseup().click();
-    assert.ok($('md-dialog').length, 'dialog is still showing');
+    click('.md-dialog-container');
```
panthony: `await click()`

tests/integration/components/paper-dialog-test.js
```diff
@@ -171,18 +166,17 @@ module('Integration | Component | paper dialog', function(hooks) {
       {{/paper-dialog}}
     `);
 
-    assert.ok($('md-dialog').length, 'dialog is showing');
+    assert.dom('md-dialog').exists();
 
-    $('#the-button').mousedown().mouseup().click();
-    assert.ok($('md-dialog').length, 'dialog is still showing');
+    click('#the-button');
```
panthony: `await click()`

tests/integration/components/paper-dialog-test.js
```diff
@@ -229,39 +222,40 @@ module('Integration | Component | paper dialog', function(hooks) {
       </button>
     `);
 
-    $('#theorigin').focus();
-    assert.equal(document.activeElement, $('#theorigin').get(0));
-    $('#theorigin').click();
+    await focus('#theorigin')
+    assert.dom('#theorigin').isFocused();
+    click('#theorigin')
```
panthony: `await click()`

tests/integration/components/paper-dialog-test.js
```diff
@@ -201,15 +195,14 @@ module('Integration | Component | paper dialog', function(hooks) {
...
     `);
 
-    assert.ok($('md-dialog'), 'dialog is showing');
-
-    let event = new $.Event('keydown');
-    event.keyCode = 27;
-    $('md-dialog').trigger(event);
+    assert.dom('md-dialog').exists();
+  
+    triggerKeyEvent('md-dialog', 'keydown', 27);
```
panthony: `await triggerKeyEvent()`

addon/components/paper-dialog.js
```diff
@@ -42,20 +41,21 @@ export default Component.extend({
       return '#ember-testing';
     }
     let parent = this.get('defaultedParent');
-    let $parent = $(parent);
-    // If the parent isn't found, assume that it is an id, but that the DOM doesn't
+    let parentEle = document.querySelector(parent);
```
panthony: Without jQuery the proper  way is to test if 'parent' is a string or an element.

You could do:

```js
if (typeOf(parent) === 'string') {
  // according to the docs, it's either an element or a selector, no need to test for '#'...
  return parent;
} else {
  //
} 

```

betocantu93: What about the jQuery objects? can I just assume strings or element and update docs?

addon/components/paper-dialog-inner.js
```diff
@@ -18,32 +18,41 @@ export default Component.extend(Translate3dMixin, {
...
+  
   },
 
   willDestroyElement() {
     this._super(...arguments);
-    let images = this.$().find('img');
-    images.off(`load.${this.elementId}`);
+    let imageElements = this.element.querySelectorAll('img');
+    imageElements.forEach( image => {
+      image.removeEventListener('load', run.bind(this, this.checkContentOverflow))
```
miguelcobain: I think that this `removeEventListener` is not working as expected, because `run.bind` returns a new function, different from the one used on the previous `window.addEventListener`.

I think we need to do the same trick that you did on #995, which is basically:

```js
this._checkContentOverflow = run.bind(this, this. checkContentOverflow);
imageElements.forEach((image) => {
  image.addEventListener('load', this._checkContentOverflow);
});
```

and later

```js
imageElements.forEach((image) => {
  image.removeEventListener('load', this, this._checkContentOverflow);
});
```

addon/components/paper-dialog.js
```diff
@@ -42,20 +41,21 @@ export default Component.extend({
       return '#ember-testing';
     }
     let parent = this.get('defaultedParent');
-    let $parent = $(parent);
-    // If the parent isn't found, assume that it is an id, but that the DOM doesn't
+    let parentEle = typeof parent === "string" ? document.querySelector(parent) : parent;
```
miguelcobain: use single quotes in js files.

addon/components/paper-dialog.js
```diff
@@ -42,20 +41,21 @@ export default Component.extend({
...
+    // If the parentEle isn't found, assume that it is an id, but that the DOM doesn't
     // exist yet. This only happens during integration tests or if entire application
     // route is a dialog.
-    if ($parent.length === 0 && parent.charAt(0) === '#') {
+    if (!parentEle && parent.charAt(0) === '#') {
```
miguelcobain: This creates an impossible condition.
i.e if `!parentEle` then `parent.charAt(0) === '#'` can never be true as well.

I that what we want is:

```js
if (typeof parent === 'string' && parent.charAt(0) === '#') {
```

addon/components/paper-dialog.js
```diff
@@ -68,18 +68,22 @@ export default Component.extend({
   didInsertElement() {
     this._super(...arguments);
     if (this.get('escapeToClose')) {
-      $(this.get('destinationId')).on(`keydown.${this.elementId}`, (e) => {
+
+      this._destinationEle = document.querySelector(this.destinationId);
```
miguelcobain: I think we should use `this.get('destinationId')` since `destinationId` is a computed property.

addon/mixins/translate3d-mixin.js
```diff
@@ -73,22 +73,36 @@ export default Mixin.create({
    */
   willDestroyElement() {
     this._super(...arguments);
+  
+    let config = getOwner(this).resolveRegistration('config:environment');
+
+    let containerClone = this.element.parentElement.cloneNode(true);
```
miguelcobain: Use `parentNode` instead, as it is being used in other places and to increase compatibility a little bit.

addon/mixins/translate3d-mixin.js
```diff
@@ -73,22 +73,36 @@ export default Mixin.create({
...
+    
+    if (config.environment === 'test' && !this.get('parent')) {
+      parent = '#ember-testing';
+    }
 
-    let containerClone = this.$().parent().clone();
-    let dialogClone = containerClone.find('md-dialog');
-    $(this.get('defaultedParent')).parent().append(containerClone);
-
+    document.querySelector(parent).parentElement.appendChild(containerClone);
```
miguelcobain: parentElement -> parentNode

addon/mixins/translate3d-mixin.js
```diff
@@ -73,22 +73,36 @@ export default Mixin.create({
...
 
+    let origin = typeof this.get('origin') === "string"
+                    ? document.querySelector(this.get('origin'))
+                    : this.get('origin');
+
     nextTick().then(() => {
-      dialogClone.removeClass('md-transition-in');
-      dialogClone.addClass('md-transition-out');
-      dialogClone.attr('style', toStyle);
+      dialogClone.classList.remove('md-transition-in')
```
miguelcobain: Needs `;`

tests/integration/components/paper-dialog-test.js
```diff
@@ -57,36 +53,35 @@ module('Integration | Component | paper dialog', function(hooks) {
       {{/paper-dialog}}
     `);
 
-    assert.ok(!$('#paper-wormhole md-dialog').length, 'did not render in default');
-    assert.ok($('#sagittarius-a md-dialog').length, 'rendered in parent');
-    assert.ok($('#ember-testing md-dialog').length, 'still rendered in ember-testing');
+    assert.dom('#paper-wormhole md-dialog').doesNotExist()
+    assert.dom('#sagittarius-a md-dialog').exists()
+    assert.dom('#ember-testing md-dialog').exists()
```
miguelcobain: Needs `;`

tests/integration/components/paper-dialog-test.js
```diff
@@ -57,36 +53,35 @@ module('Integration | Component | paper dialog', function(hooks) {
...
 
   test('should only prevent scrolling behind scoped modal', async function(assert) {
     await render(hbs`
       <div id="sagittarius-a"></div>
       {{paper-dialog parent="#sagittarius-a"}}
     `);
-
-    assert.equal($('md-backdrop').css('position'), 'absolute', 'backdrop is absolute');
+  
+    assert.equal(window.getComputedStyle(find('md-backdrop')).getPropertyValue("position"), 'absolute', 'backdrop is absolute');
```
miguelcobain: single quotes in js files.

PR #992, opened 2018-10-28
paper-input no jquery

betocantu93:
I'm not quite sure about the setValue, if I didn't validated the value, it placed undefined in the inputs...
```javascript
setValue(value) {
    if (this.element.querySelector('input, textarea').value !== value && value) {
      this.element.querySelector('input, textarea').value = value;
    }
  },
```

### DIFF
--- a/addon/components/paper-input.js
+++ b/addon/components/paper-input.js
@@ -4,7 +4,6 @@
 import { or, bool, and } from '@ember/object/computed';
 
 import Component from '@ember/component';
-import $ from 'jquery';
 import { computed } from '@ember/object';
 import { isEmpty } from '@ember/utils';
 import { run } from '@ember/runloop';
@@ -90,7 +89,7 @@ export default Component.extend(FocusableMixin, ColorMixin, ChildMixin, Validati
   didInsertElement() {
     this._super(...arguments);
     if (this.get('textarea')) {
-      $(window).on(`resize.${this.elementId}`, run.bind(this, this.growTextarea));
+      window.addEventListener('resize', run.bind(this, this.growTextarea));
     }
   },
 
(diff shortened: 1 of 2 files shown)


addon/components/paper-input.js
```diff
@@ -104,55 +103,62 @@ export default Component.extend(FocusableMixin, ColorMixin, ChildMixin, Validati
   willDestroyElement() {
     this._super(...arguments);
     if (this.get('textarea')) {
-      $(window).off(`resize.${this.elementId}`);
+      window.removeEventListener('resize', run.bind(this, this.growTextarea));
```
miguelcobain: I think that this `removeEventListener` is not working as expected, because `run.bind` returns a new function, different from the one used on the previous `window.addEventListener`.

I think we need to do the same trick that you did on #995, which is basically:

```js
this._growTextarea = run.bind(this, this.growTextarea);
window.addEventListener('resize', this._growTextarea);
```

and later

```js
window.removeEventListener('resize', this._growTextarea);
```

addon/components/paper-input.js
```diff
@@ -104,55 +103,62 @@ export default Component.extend(FocusableMixin, ColorMixin, ChildMixin, Validati
...
         let rowsToSet = Math.min(proposedHeight, maxRows);
-        inputElement
-          .css('height', `${this.lineHeight * rowsToSet}px`)
-          .attr('rows', rowsToSet)
-          .toggleClass('md-textarea-scrollable', proposedHeight >= maxRows);
+        
+        inputElement.style.height = `${this.lineHeight * rowsToSet}px`;
+        inputElement.setAttribute('rows', rowsToSet);
+        
+        if(proposedHeight >= maxRows) {
```
miguelcobain: Need space after `if`.

tests/integration/components/paper-input-test.js
```diff
@@ -46,8 +45,9 @@ module('Integration | Component | paper input', function(hooks) {
 
     await render(hbs`{{paper-input iconComponent="custom-icon" icon="person" onChange=dummyOnChange}}`);
 
-    assert.equal(this.$('md-input-container md-icon').length, 0, 'default icon component is not rendered');
-    assert.equal(this.$('md-input-container .custom-icon').length, 1, 'custom icon component rendered');
+    assert.dom('md-input-container md-icon').doesNotExist()
+    assert.dom('md-input-container .custom-icon').exists()
```
miguelcobain: This should be:

```
assert.dom('md-input-container .custom-icon').exists({ count: 1 }, 'custom icon component rendered');
```

- try to keep assertion messages
- use `exists: 1` to make sure only one item is rendered
- keep semicolon `;`

PR #993, opened 2018-10-30
paper-menu no jquery

betocantu93:
Refactor tests to use qunit-dom and components from using jquery


addon/components/paper-menu-content-inner.js
```diff
@@ -54,12 +54,12 @@ export default Component.extend(ParentMixin, {
   },
 
   focusMenuItem(e, direction) {
-    let currentItem = this.$(e.target).closest('md-menu-item');
+    let currentItem = e.target.closest('md-menu-item');
```
miguelcobain: unfortunately, we can't use `closest` as that is not supported in IE11.
We're trying to keep IE11 support because ember also does.

With that being said, I wouldn't mind if we added a polyfill to our dependencies: https://github.com/jonathantneal/closest

Then we `app.import()` the file in our `index.js` and we can start using closest anywhere.

tests/integration/components/paper-menu-test.js
```diff
@@ -98,17 +54,15 @@ module('Integration | Component | paper menu', function(hooks) {
       {{/menu.content}}
     {{/paper-menu}}`);
 
-    return settled().then(() => {
-      clickTrigger();
+    return settled().then(async () => {
```
miguelcobain: we can use `await settled();` and get rid of this annoying nesting. :)

tests/integration/components/paper-menu-test.js
```diff
@@ -165,32 +119,29 @@ module('Integration | Component | paper menu', function(hooks) {
...
-        let e = new $.Event('keydown');
-        let menu = $('md-menu-content');
-        e.which = 40;
-        e.target = menu[0].firstElementChild;
+        let selectors = findAll('md-menu-item')
+        assert.dom(selectors[0].firstElementChild).hasClass('md-focused')
 
-        $(menu[0].firstElementChild).trigger(e);
+        let menu = findAll('md-menu-content');
+        await triggerKeyEvent(menu[0].firstElementChild, "keydown", 40)
```
miguelcobain: Use single quotes in js. `'keydown'`

addon/components/paper-menu-content.js
```diff
@@ -99,14 +111,14 @@ export default ContentComponent.extend({
     let parentElement = this.get('renderInPlace') ? dropdownElement.parentElement.parentElement : dropdownElement.parentElement;
     let clone = dropdownElement.cloneNode(true);
     clone.id = `${clone.id}--clone`;
-    let $clone = $(clone);
+    let _clone = clone;
```
miguelcobain: I think we can just remove the `_clone` variable and just use `clone`.

betocantu93:
@miguelcobain I updated the PR to use the addon, but I get some conflicts with the yarn.lock, can you give me a hand here?

PR #994, opened 2018-10-30
paper-select no jquery

betocantu93:
Refactor jquery from paper-select

### DIFF
--- a/addon/components/paper-select-content.js
+++ b/addon/components/paper-select-content.js
@@ -46,16 +46,15 @@ export default PaperMenuContent.extend({
     let parentElement = this.get('renderInPlace') ? dropdownElement.parentElement.parentElement : dropdownElement.parentElement;
     let clone = dropdownElement.cloneNode(true);
     clone.id = `${clone.id}--clone`;
-    let _clone = clone;
     parentElement.appendChild(clone);
     
     clone.children[0].children[0].scrollTop = dropdownElement.children[0].children[0].scrollTop;
     window.requestAnimationFrame(() => {
       if (!this.get('isDestroyed')) {
         this.set('isActive', false);
-        _clone.classList.add('md-leave');
+        clone.classList.add('md-leave');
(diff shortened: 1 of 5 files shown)


addon/components/paper-select-content.js
```diff
@@ -47,15 +46,16 @@ export default PaperMenuContent.extend({
     let parentElement = this.get('renderInPlace') ? dropdownElement.parentElement.parentElement : dropdownElement.parentElement;
     let clone = dropdownElement.cloneNode(true);
     clone.id = `${clone.id}--clone`;
-    let $clone = $(clone);
+    let _clone = clone;
```
miguelcobain: We can just use the `clone` variable and remove `_clone`.

addon/components/paper-select-menu.js
```diff
@@ -31,19 +30,19 @@ export default PaperMenu.extend({
   triggerComponent: 'paper-select-menu-trigger',
 
   calculatePosition(trigger, content, destination, { dropdown }) {
-    let $dropdown = $(content);
+    let _dropdown = content;
```
miguelcobain: We can remove the `_dropdown` variable and use `content` directly.

addon/components/paper-select-menu.js
```diff
@@ -31,19 +30,19 @@ export default PaperMenu.extend({
...
+      selectEl: _dropdown.querySelector('md-select-menu'),
+      contentEl: _dropdown.querySelector('md-content')
     };
 
-    let containerNode = $dropdown.get(0);
-    let targetNode = opts.target[0].firstElementChild; // target the label
-    let parentNode = opts.parent.get(0);
-    let selectNode = opts.selectEl.get(0);
-    let contentNode = opts.contentEl.get(0);
+    let containerNode = _dropdown;
```
miguelcobain: `containerNode` can also be replaced by `content` as well, I think.

PR #995, opened 2018-10-30
Paper sidenav no jquery

betocantu93:
Remove jquery from tests and component

### DIFF
--- a/addon/components/paper-sidenav-inner.js
+++ b/addon/components/paper-sidenav-inner.js
@@ -3,10 +3,8 @@
  */
 /* globals FastBoot */
 import { inject as service } from '@ember/service';
-
 import Component from '@ember/component';
 import { computed } from '@ember/object';
-import $ from 'jquery';
 import { run } from '@ember/runloop';
 import TransitionMixin from 'ember-css-transitions/mixins/transition-mixin';
 import { invokeAction } from 'ember-invoke-action';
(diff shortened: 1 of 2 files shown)


tests/integration/components/paper-sidenav-test.js
```diff
@@ -80,12 +81,12 @@ module('Integration | Component | paper sidenav', function(hooks) {
 
     await render(hbs`{{paper-sidenav open=isOpen lockedOpen=false}}`);
 
-    assert.ok(this.$('md-sidenav').hasClass('md-closed'));
+    assert.dom('md-sidenav').hasClass('md-closed');
 
     this.set('isOpen', true);
 
     return settled().then(() => {
```
miguelcobain: We can use 
```js
await settled();
assert.dom('md-sidenav').doesNotHaveClass('md-closed');
```

instead of the promise then.

Can you convert that, please?

Thank for all your efforts on removing jquery. It's really valuable!

addon/components/paper-sidenav-inner.js
```diff
@@ -49,7 +47,8 @@ export default Component.extend(TransitionMixin, {
 
   didInsertElement() {
     this._super(...arguments);
-    $(window).on(`resize.${this.elementId}`, run.bind(this, 'updateLockedOpen'));
+    this._updateOnResize = run.bind(this, 'updateLockedOpen')
```
miguelcobain: Need `;` here.

miguelcobain: Can also use `run.bind(this, this.updateLockedOpen)` instead of the "magic" string.

PR #996, opened 2018-10-31
respect autoprefixer enable option defined by the host app

xomaczar:
If host app disabled autoprefixer (in dev mode for example), this will disable ember-paper's `autoprefixing` dropping the rebuild time by `~500ms`.


xomaczar:
cc @miguelcobain

xomaczar:
Addressed in #997

PR #1001, opened 2018-11-01
util no-jquery

betocantu93:
Refactor jquery from util service, I'm not actually sure if it's done... I don't know how to test it.


miguelcobain:
I've managed to remove the util service from the codebase, so this is not needed any longer.
Thanks.

PR #1008, opened 2018-11-01
paper-virtual-repeat no jquery

betocantu93:
Refactor jquery from paper-virtual-repeat-scroller.js, paper-virtual-repeat.js and tests.


addon/components/paper-virtual-repeat.js
```diff
@@ -63,20 +63,25 @@ const VirtualRepeatComponent = VirtualEachComponent.extend({
     let style = this.get('positionStyle');
 
     if (height !== null && !isNaN(height)) {
-      style += ` height: ${height}px;`;
+      style += ` height: em{height}px;`;
```
miguelcobain: This `em` doesn't belong here, right?

betocantu93: Indeed, typo, my bad

PR #1009, opened 2018-11-01
base-focusable no jquery

betocantu93:
I couldn't run this tests, but I just refactored them, it seems ok.


miguelcobain:
You couldn't run the tests because this file didn't end in `-test.js`.
Anyway, I removed it because this was a test for a component that doesn't exist anymore. We had a base-focusable component, but it started being a mixin some time ago.

Thanks.

PR #1014, opened 2018-11-02
paper-item no jquery

betocantu93:
Refactor jquery from tests


miguelcobain:
I fixed the semicolons and squashed the commits. Thank you!

PR #1019, opened 2018-11-02
package.json no jquery

betocantu93:
Change `ember-ajax` devDependency for `ember-fetch`.


miguelcobain:
We actually don't use any ajax stuff, so no need to include this dependency.

PR #1020, opened 2018-11-02
Blacklist components

betocantu93:
Use broccoli-funnel ember-bootstrap approach to reduce the build size with whitelist, blacklist approach.


betocantu93:
Closed in favor of #1020

PR #1021, opened 2018-11-02
Blacklist whitelist components

betocantu93:
The past PR was corrupted, this is a clean PR.

whitelist/blacklist components using ember-bootstrap approach.

For whitelist, all the component dependencies will also be included.


betocantu93:
@miguelcobain yep, that's a good idea!

PR #1023, opened 2018-11-14
Added missing slashes in closing tags of typography demo template

muesli:
I assume the missing slashes in the closing tags were just an oversight and not intentional.


miguelcobain:
Thank you. 👍

PR #1024, opened 2018-11-19
Associate input field with corresponding info messages for better a11y

praskovia-root:
Adding aria-describedby attribute to input fields with reference to occasionally appearing info text element below the related field causes screen readers to read out input fields together with corresponding info text (mainly error notification, e.g. failed validation). This change matches [1.3.1 Info and Relationships](https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships.html) success criterion of WCAG 2.
Without this connection the info underneath the input field is being read out separately and for screen reader users it may not be clear which input element this info belongs to.

### DIFF
--- a/addon/templates/components/paper-input.hbs
+++ b/addon/templates/components/paper-input.hbs
@@ -8,6 +8,7 @@
 
 {{#if textarea}}
   <textarea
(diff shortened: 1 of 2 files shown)


miguelcobain:
Looks good and more a11y PRs are also welcome!

PR #1026, opened 2018-11-20
Improve paper-icon a11y

praskovia-root:
Icon's and related text element's meaning usually overlap. Which is why reading out icons' label is often redundant. Developers should be able to prevent screen reader from reading out paper-icons using aria-hidden attribute which is set to false by default. This change aims [WCAG2 Success Criterion 1.1.1: Non-text Content](https://www.w3.org/WAI/WCAG21/Understanding/non-text-content.html)


miguelcobain:
Thank you. 👍

PR #1029, opened 2018-11-27
Allow to add title attribute to paper-input

mikoscz:
It allows us to add `title` attribute to the input field - improvement for screen readers.


miguelcobain:
Thank you. 👍

PR #1030, opened 2018-11-28
fix searchable paper-select in dummy app

wahhaj:
previous implementation made it impossible to select an item from the list.


miguelcobain:
Thanks a lot. 👍

PR #1031, opened 2018-11-28
Make checkbox accessible

mikoscz:
I've added support for
- [x] aria-checked
- [x] aria-labelledby
- [x] aria-label
and the `role="checkbox"`

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/checkbox_role

@miguelcobain Should I also add `type="checkbox"`? I see that the angular's version has it but I am not sure.

### DIFF
--- a/addon/components/paper-checkbox.js
+++ b/addon/components/paper-checkbox.js
@@ -3,6 +3,7 @@
  */
 import { inject as service } from '@ember/service';
 
+import { computed } from '@ember/object';
 import { not, and } from '@ember/object/computed';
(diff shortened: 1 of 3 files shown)


miguelcobain:
According to MDN, `type` is only used on `<button>, <input>, <command>, <embed>, <object>, <script>, <source>, <style>, <menu>` elements.

https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes

The PR LGTM! Thanks.

mikoscz:
Thanks!

PR #1038, opened 2019-01-02
paper-tabs: correct 'paginationStyle' updated twice in single render

panthony:
Two fixes in this PR:

- ensure `newOffset` is never undefined
- fix `paginationStyle updated twice in single render`

In my case, the later only occurred in tests. Not sure if due to the "speed" of the tests or the viewport that is much smaller and trigger more often the pagination

### DIFF
--- a/addon/components/paper-tabs.js
+++ b/addon/components/paper-tabs.js
@@ -35,7 +35,8 @@ export default Component.extend(ParentMixin, ColorMixin, {
 
     this.setMovingRight();
 
-    this.fixOffsetIfNeeded();
+    this.scheduleAfterRender(() =>
+      this.fixOffsetIfNeeded());
 
     this.set('_previousSelectedTab', selectedTab);
   }),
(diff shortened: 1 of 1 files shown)


panthony:
Still have issues with pagination & infinite rendering... Urgh this is a nightmare!

panthony:
The problem is that there is too many cascading 'updateXXX' being called.

For exemple the simple fact of clicking on another tab will change the value of `_selectedTab` multiple time and trigger all the 'updateXXXXX' functions.

Changing it will call 'updateXXX' which will call 'didRender' which will call again 'updateXXX' ...

betocantu93:
@panthony Some time ago, @miguelcobain and I did track the infinite rendering issue, in this PR #980 but later it was reverted cuz of the following issue 

> ... It's causing a bug where we can't use isSelected directly on a paper-tab. ...
... the isSelected property is very important for use cases where we need the is-active helper to decide which tab is active based on the current route. ...

panthony:
@betocantu93 I was thinking of maybe using an observer on the isSelected property which would use decounce to update the selectedTab.

Things I noticed too:
  - didRender calls updateCanvasWidth that calls 'updateDimensions' on each tab but whereas it's already done by the 'didRender' of the tab and from what I could see, didRender on child components is always called BEFORE the parent component.

PR #1044, opened 2019-01-11
ensure 'parentNode' is not NULL in translate3d

### LINKED ISSUE #1040: translate-3d-mixin: trying to manipulate DOM on destroyed component
The following line:

> containerClone.parentNode.removeChild(containerClone);

Failed because `e.parentNode` was null.

Simply missing a check on isDestroyed && isDestroying before trying to do anything in the `run.next` callback.

panthony:
Fixes #1040

### DIFF
--- a/addon/mixins/translate3d-mixin.js
+++ b/addon/mixins/translate3d-mixin.js
@@ -96,7 +96,9 @@ export default Mixin.create({
       dialogClone.style.cssText = toStyle;
       nextTick().then(() => {
         run.later(() => {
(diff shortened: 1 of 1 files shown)


miguelcobain:
I don't see any problem with adding these checks.
But I'm wondering how is it possible for parent to not exist.
The default parent should always be there. Maybe this happens when you specify a custom parent?
Did this error pop up in a test?

xtagon:
Thanks so much for this fix! My test suite was failing randomly/intermittently and I was having a hard time tracking it down. This was it.

PR #1048, opened 2019-01-12
Trying to address paper-tabs issues

### LINKED ISSUE #759: Paper-tabs initial render not working

### LINKED ISSUE #858: Paper tabs calculate wrong position for the inkBar on app loading

### LINKED ISSUE #893: paper-tabs: setting center=true squashes tabs / breaks pagination
If tabs are rendered with `center=true` on a small screen, they get squashed up to fit within the visible space, preventing pagination from kicking in:
localhost_4222_ iphone 7

This can be seen by adding `center: true` to the `demo/tabs.js` controller in the dummy app and loading it in a narrow browser (i.e. on a mobile device or using mobile view in dev tools).

I'm guessing that this hasn't been noticed yet because if `center` is toggled after page load (such as by ticking the `center` box in the docs) or you open the page in a larger window and then resize to mobile size, the tabs appear full size and paginated, as expected:
localhost_4222_ iphone 7 1

panthony:
Hopefully:

- Superseed #982 
- Fixes #858
- Fixes #759 
- Fixes #893 

I'm trying to take a fresh start at paper tabs offsets handling as they are may issues at the moment.

If they are things that I'm missing feel free to tell me.

Amongs the hopefully fixed issues:

- infinite rendering
- movingRight that is not properly set
- currentOffset that may not be set
- inkBar that may be rendered without proper left/right offset
- lot's of (I think) unnecessary 'updateXXX' calls

Issues yet to be fixed:

- when you paginate, it triggers a re-render that eventually calls 'fixOffset' which 'cancel' the pagination because it caused the selected tab to be partially hidden :sob:

**edit**:

To correct this 'fixOffset' issue that breaks the pagination I now call it in only 2 cases:

- upon first render to ensure the selected tab is visible
- upon selected tab update to put the whole tab in focus

I'm not calling it anymore in 'didRender' which means that if the tab's label is updated while selected it might become partially hidden but this is way WAY minor IMO.

**edit²**:

Unfortunately I do not have time for this right away but I think we should add some tests with a wrapper around "paper-tabs" with a small max width, to ensure that we can still paginate around.

### DIFF
--- a/addon/components/paper-tab.js
+++ b/addon/components/paper-tab.js
@@ -47,6 +47,7 @@ export default Component.extend(ChildMixin, RippleMixin, FocusableMixin, {
     // this is the initial tab width
     // it is used to calculate if we need pagination or not
     this.set('width', width);
+    this.set('left', this.element.offsetLeft);
   },
 
   didRender() {
--- a/addon/components/paper-tab.js
+++ b/addon/components/paper-tab.js
@@ -41,15 +41,6 @@ export default Component.extend(ChildMixin, RippleMixin, FocusableMixin, {
     }
   },
 
-  didInsertElement() {
-    this._super(...arguments);
-    let width = this.element.offsetWidth;
-    // this is the initial tab width
-    // it is used to calculate if we need pagination or not
-    this.set('width', width);
-    this.set('left', this.element.offsetLeft);
-  },
-
   didRender() {
     this._super(...arguments);
     this.updateDimensions();
--- a/addon/components/paper-tab.js
+++ b/addon/components/paper-tab.js
@@ -41,12 +41,7 @@ export default Component.extend(ChildMixin, RippleMixin, FocusableMixin, {
     }
   },
 
-  didRender() {
-    this._super(...arguments);
-    this.updateDimensions();
-  },
-
-  // this method is also called by the parent
+  // this method is called by the parent
   updateDimensions() {
     // this is the true current width
     // it is used to calculate the ink bar position & pagination offset
--- a/addon/components/paper-tab.js
+++ b/addon/components/paper-tab.js
@@ -57,11 +57,12 @@ export default Component.extend(ChildMixin, RippleMixin, FocusableMixin, {
 
   // this method is also called by the parent
   updateDimensions() {
-    let left = this.element.offsetLeft;
     // this is the true current width
-    // it is used to calculate the ink bar position
-    let currentWidth = this.element.offsetWidth;
-    this.setProperties({ left, currentWidth });
+    // it is used to calculate the ink bar position & pagination offset
+    this.setProperties({
+      left: this.element.offsetLeft,
+      width: this.element.offsetWidth
+    });
   },
 
   click() {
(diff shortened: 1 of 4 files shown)


panthony:
@betocantu93 @miguelcobain Do you have an idea how we could fix the following point:

- we have a selected tab
- we click "prev", the currentOffset is updated in such a way that selected tab is partially hidden
- didRender is triggered
- fixOffset restore the currentOffset in such a way that cancel the prev action.

No sure how we could "cleanely" differenciate when we should try to fix the current offset and when not.

panthony:
Another annoying issue:

- tab 1 is selected
- you click on tab 2
- didRender is called with _selectedTab still on tab1
- didRender is called (twice) with _selectedTab on tab2

The problem is that, although you clicked on tab2, the 'fixOffset' will be triggered with tab1 as selected tab, meaning that it will try to correct the offset for this tab.

chbonser:
@panthony I pulled you updated into a project and it resolved an issue for me where the next/previous buttons were disabled when they shouldn't have been. I'll do some more testing, but I see no issues so far.

betocantu93:
@panthony sorry for the delay, I also pulled your tag and it works, gonna keep testing my use cases, and will let you know, thanks for the effort

pauln:
I spotted this PR after submitting my own to fix the pagination issues.  I can confirm that this fixes #1063 for me, as well as the other issues listed in the OP.

miguelcobain:
Thanks a lot @panthony. Sorry for taking so long to review. This looks great!

PR #1056, opened 2019-01-24
add little github icon in demo toolbar

panthony:
The idea is to have a more visible link between the demo website and the github repository.

capture d ecran 2019-01-24 a 09 51 22


miguelcobain:
Thank you.

PR #1068, opened 2019-02-21
Fix tab pagination controls

### LINKED ISSUE #1063: paper-tabs: arrows to scroll on overflowing tabs are always disabled
Bisected to https://github.com/miguelcobain/ember-paper/commit/23073e806cb50391cae55b8cf36853cfeed292b0

pauln:
Fixes #1063.

There were two main issues:
- `fixOffsetIfNeeded()` had been added into `updateCanvasWidth`, so it was being called on every render.  This means that it ran after the `nextPage` action, and effectively reverted the pagination offset change because it doesn't know/care about pagination (only the selected tab)
- `fixOffsetIfNeeded()` was setting `this.currentOffset` to `undefined` in some situations, which then prevented pagination from working as it's used in calculations

In order to address these issues, I've removed `fixOffsetIfNeeded()` from `updateCanvasWidth` (it seems to be being called directly where needed already) and defaulted `newOffset` to `currentOffset` in `fixOffsetIfNeeded()`.  I've done some manual testing in addition to running the automated tests, and all seems well.

### DIFF
--- a/addon/components/paper-tabs.js
+++ b/addon/components/paper-tabs.js
@@ -74,7 +74,6 @@ export default Component.extend(ParentMixin, ColorMixin, {
     let updateCanvasWidth = () => {
       this.updateDimensions();
       this.updateStretchTabs();
-      this.fixOffsetIfNeeded();
     };
 
     window.addEventListener('resize', updateCanvasWidth);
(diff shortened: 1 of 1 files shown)


pauln:
Closing in favour of #1048, which I noticed after submitting this PR - and have since tested and confirmed that it fixes (for me, at least) both #1063 and #893, which was next on my hit list.

Issue #925, opened 2018-05-14
paper-select causes the app to become unresponsive


chbonser:
Occasionally (5 or 10% of the time) `paper-select` causes my app to become unresponsive.

When this occurs the `#ember-basic-dropdown-wormhole` contains both an `md-backdrop` element and a div for the dropdown content `#ember-basic-dropdown-content-...`.

The content div returns the following from `getBoundingClientRect()`, making it invisible and unclickable:
```
{
  "x": 0,
  "y": 0,
  "width": 0,
  "height": 0,
  "top": 0,
  "right": 0,
  "bottom": 0,
  "left": 0
}
```

The `md-backdrop` element masks the rest of the app and _does not respond to clicks_ making the app unresponsive. There are no errors in the console.

The only other unusual thing about this use case is that the drop down exists in a modal (which animates open) which has tabs (with animations between tabs). Could the animation of the parent context be causing this bug?

### Related (?) Bugs

This was observed with both Ember 3.0.0 and 3.1 applications. ~~I'm also using `paper-select` with a 3.0 application and this behavior does not occur~~

I do sometimes experience two other issues which I'll write up separately:
- sometimes the options list appears a few hundred px above the expected location
- sometimes the user has to click multiple times to make a selection

panthony:
I did experienced the "click multiple times to make a selection". Like I had to click exactly on the label's text and not anywhere else.

For the options list that appears above the expected location it may have been recently fixed (https://github.com/miguelcobain/ember-paper/pull/895).

chbonser:
@panthony Thanks! Nice find on #895. I'll cross that off my list of potentially related things.

I'm still digging into my original issue. And I've made an additional observation... sometimes the `paper-dialog` content (and/or `md-backdrop`) are left in the dom after the `paper-dialog` has been removed. It seems clear that something about my use case breaks how ember-paper removes elements from the wormholes.

chbonser:
To be doubly sure, I setup a debugger to trigger only if the 'timeout' event returned by the `later` call is about to be cleared. I recreated the issue and the debugger is not tripped. Something else is canceling (skipping?) the later call.

chbonser:
When I can get the app into the error state there are hundreds of timers in [backburner](https://github.com/BackburnerJS/backburner.js/blob/fed33e7208ff2b39a7161481c64e1a0f2b73851c/lib/index.ts#L207)'s `_timers` prop and many have error like this one...

```
[...]

PR #1075, opened 2019-03-21
Fix backdrop-bug

ruzz311:
When using multiple components that have an invisible background (like paper-menu) paper-backdrop seems to remain on the page and capture clicks.  After looking through issues I found issue #925 where @panthony mentioned a work around but never offered it as a PR.  I'm doing that for him because this issue is annoying.


panthony:
I never offered a PR because it was not really a fix but more of a work around.

There is a true issue upstream that should be fixed 👍 

That being said, a transition effect on an invisible div is wasted resources.

ruzz311:
Thats fair @panthony - I'm closing.

PR #1076, opened 2019-04-04
Fix paper-autocomplete value being overriden when changed

### LINKED ISSUE #1043: paper-autocomplete searchText is overriden on input
Hi,

I'm using the latest v1.0.0-beta21 and I have an issue with autocomplete.

Whenever I type something in the autocomplete, it doesn't show me what I type. The second character override the first one so the search is always using 1 character ...

Does anyone got that one ?

Using Ember 3.7 (and 3.6 before).

xtagon:
Fixes #1043

### DIFF
--- a/addon/components/paper-input.js
+++ b/addon/components/paper-input.js
@@ -162,7 +162,7 @@ export default Component.extend(FocusableMixin, ColorMixin, ChildMixin, Validati
     // normalize falsy values to empty string
     value = isEmpty(value) ? '' : value;
 
-    if (this.element.querySelector('input, textarea').value !== value) {
+    if (this.element.querySelector('input, textarea').value !== value && value) {
       this.element.querySelector('input, textarea').value = value;
     }
   },


xtagon:
I tried to add a new test for this scenario but wasn't able to get it to reproduce in a test. I did verify in my own app that it's no longer affected after this change, and the existing tests passed.

Thanks! 👍

xtagon:
Anything holding this back from being merged?

miguelcobain:
@xtagon sorry for taking so long to merge this. Thank you!

miguelcobain:
@xtagon can you please retest if this bug persists on beta.26?

Basically this fix introduced a new bug. We were barking at the wrong door. The bug was somewhere else.

xtagon:
Good catch, I will have a chance to test it in the next few days.

xtagon:
@miguelcobain beta.26 looks good to me

PR #1081, opened 2019-05-15
Default validators messages

betocantu93:
This lets us set default error messages in a global config


betocantu93:
closed in favor of #1093

PR #1084, opened 2019-05-18
Add guard against missing original.parentNode

backspace:
This is a naïve fix for a problem I was experiencing in tests after updating Ember Paper, you can see an example here:
https://travis-ci.org/backspace/prison-rideshare-ui/builds/530523899#L578

It seems somewhat akin to what’s discussed in #925, but I’m definitely at the limit of my understanding of what’s happening. I only encountered this problem in tests.

Let me know if I can provide any details from the application that might help solve this differently.

### DIFF
--- a/addon/components/paper-backdrop.js
+++ b/addon/components/paper-backdrop.js
@@ -29,7 +29,9 @@ export default Component.extend(TransitionMixin, {
   }),
 
   addDestroyedElementClone(original, clone) {
-    original.parentNode.appendChild(clone);
+    if (original.parentNode) {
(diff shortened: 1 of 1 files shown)


backspace:
I should have linked to the [tests for the following commit](https://travis-ci.org/backspace/prison-rideshare-ui/builds/530527303) when I changed to this fork and the `appendChild` message goes from 65 instances to 0.

It looks like the test failures are existing and unrelated to this PR 😞

miguelcobain:
Seems like during tests, the parent could be destroyed, and there's no need to add the cloned element for animation. Thanks.

Issue #1087, opened 2019-05-31
Tooltips for {{paper-input}}


Oreoz:
Hi there!

I'm trying to implement a tooltip feature for the paper-input component. I'm not able to get it to render when placed inside of the `label` tag. When I move it out of the `label` tag it renders without any problems.

Is there any reason for this? I've tried playing around with various css properties without any success.

![image](https://user-images.githubusercontent.com/3945623/58709317-e823e180-8387-11e9-8ddc-21ead238d426.png)

```hbs
{{#if label}}
  <label for={{inputElementId}} class={{if required "md-required"}}>
    {{label}}

    {{#if tooltip}}
      {{#paper-icon "info"}}
        {{#paper-tooltip position="right"}}{{tooltip}}{{/paper-tooltip}}
      {{/paper-icon}}
    {{/if}}
  </label>
{{/if}}
```

Cheers!

woohoou:
Try to wrap label + paper-icon inside the paper-tooltip.

Shadowere:
The issue with the tooltip not rendering when inside the label tag was because of `pointer-events:none;` on the label. Fixed using this:
```css
// Tooltip with input
md-input-container label md-icon {
  pointer-events: auto;
}
```

PR #1088, opened 2019-06-04
Add input with tooltip

Shadowere:
#1087 
Adds input with tooltip

- Adds tooltips to input;
- Adds input with tooltip to demo;
- Adds tests for input with tooltip;

### DIFF
--- a/addon/templates/components/paper-input.hbs
+++ b/addon/templates/components/paper-input.hbs
@@ -1,5 +1,14 @@
 {{#if label}}
-  <label for={{inputElementId}} class={{if required "md-required"}}>{{label}}</label>
+  <label for={{inputElementId}} class={{if required "md-required"}}>
+    {{label}}
+    {{#if tooltip}}
+      {{#paper-icon "info"}}
+        {{#paper-tooltip position=tooltipPosition}}
+          {{tooltip}}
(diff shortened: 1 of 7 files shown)


miguelcobain:
All of this would be solved if we had named blocks instead of a single block. This way we could customize the label with whatever content we want, including an icon + tooltip.

Since named blocks are not a reality, my suggestion is that we accept a `labelComponent` argument, which paper-input would render if truthy.
Example:

```hbs
{{paper-input
  labelComponent=(component "tooltip-label" text="myLabel" tooltip="My tooltip")
  value=myValue
  onChange=(action (mut myValue))
}}
```

In the `paper-input.hbs` template, we would do something like:

```hbs
{{#if labelComponent}}
  {{component labelComponent inputElementId=inputElementId required=required}}
{{else if label}}
  <label for={{inputElementId}} class={{if required "md-required"}}>{{label}}</label>
{{/if}}
```

In this case, you could have your own `tooltip-label` component in which you could do all sorts of crazy stuff.

I can't merge this PR the way it is now because it's too specific. It assumes that people want an info icon, etc.

Let me know what you think.

Shadowere:
I agree that providing a custom labelComponent will be cleaner and more versatile. I'll change it up soon enough

PR #1089, opened 2019-06-05
Prevent submitting a form when clicking clear in paper-autocomplete

neaket360pi:
Problem:
When using the paper-autocomplete inside of a form, its clear button is triggering the forms submit action.  This is called because the button defaults to type="submit" inside of a form.

Changes:
- update the paper-reset-button type to button to avoid the default of
  type="submit" when inside a form
- add tests for paper-reset-button and paper-autocomplete components


miguelcobain:
💯 !
Thank you.

neaket360pi:
Thank you for merging.  😄

PR #1090, opened 2019-06-13
Missing dependency for paper-autocomplete

betocantu93:
I noticed this dep was missing for paper-autocomplete... `paper-autocomplete-trigger-container`


miguelcobain:
Nice catch. Thanks.

Issue #1092, opened 2019-07-17
1.0.0-beta.25 no longer clears input from text-field/area


larryh:
Hi,

First things first: thanks for the great addon! I know it must be tough working on this (and maybe other addons) in addition to your "day job", so it is greatly appreciated.

I have a couple of textfield and textarea controls where I clear the input after a form submission in preparation for another input/submission. This has worked fine for years with ember-paper.

However, the most recent update seems to have broken this capability. Shortly after upgrading to beta.25 I ran my tests and the ones that ensured " " was in the form field after hitting "Submit" failed.

Sure enough, the text remained in the fields. I have attached screenshots of the textarea, as it is easier to see what is going on.

The first picture shows the initial state of the textarea; the second, after entering text; the third, after clicking "Submit".
 
You can barely see the " Text * " label underneath the text that I entered after clicking "Submit". But the text should be cleared, and the " Text * " label clearly visible so the user can enter another comment.

Regarding the textfield: The "Submit" button cleared out the text, but pressing the "Enter" key did not.

Backing off to 1.0.0-beat.24 fixed the problems for both the textfield and textarea.

![textarea_initial](https://user-images.githubusercontent.com/89646/61377656-4427c400-a869-11e9-8f98-452944c2ab56.png)

![textarea_with_input](https://user-images.githubusercontent.com/89646/61377665-4853e180-a869-11e9-8df3-ea5b5bfa62ea.png)

![textarea_submit](https://user-images.githubusercontent.com/89646/61377679-4db12c00-a869-11e9-8a36-e63dc3458417.png)

backspace:
I’m finding this too, in two different varieties.

The first is somewhat like @larryh’s situation: when a user changes between radio buttons, the related fields are supposed to clear. [Here](https://travis-ci.org/backspace/prison-rideshare-ui/builds/561518352#L689-L708)’s the failing test.

A GIF of it when it works:

![report-clear-working](https://user-images.githubusercontent.com/43280/61592821-bff17b80-ab9d-11e9-8d00-7c31f3438230.gif)

And what happens with `1.0.0-beta.25`:

![report-clear-broken](https://user-images.githubusercontent.com/43280/61592825-cc75d400-ab9d-11e9-8739-e22677b4534b.gif)

Another is where I’ve added a clear button to a search field. [Here](https://travis-ci.org/backspace/prison-rideshare-ui/builds/561518352#L719-L738)’s the failing test.

[...]

PR #1095, opened 2019-07-21
Add test for clearing paper-input

backspace:
This shows the bug being described in #1092. Reverting
#1076 does fix it, but then presumably #1043 would return 😞


backspace:
[Here](https://travis-ci.org/miguelcobain/ember-paper/jobs/561733352#L1086-L1090)’s the failure, as expected.

miguelcobain:
Thank you for the test. 👍

backspace:
You’re welcome, thanks for the fix! 💞

PR #1102, opened 2019-09-26
[CLEAN] Fix delay in transition at scale speed dial

betocantu93:
Clean PR for #1072 credits to @woohoou

Speed dial has wrong calculation when animation is "Scale".

The calculation of the delay is negative and it prevents click actions because the close animation start and the negative delay buttons disappear immediately so the release button event/action never trigger.


alanlalonde:
@betocantu93 please merge from master and try running Travis build again.

betocantu93:
@alanlalonde 👍

Issue #1106, opened 2019-11-04
Tests failing - Prevents TravisCI builds & Pull request merging


alanlalonde:
Over the course of the past three months, tests began failing in canary and are now failing in the main build.

> not ok 437 Chrome 78.0 - [64 ms] - Integration | Component | paper progress circular: renders correctly with explicit value and diameter
    ---
        actual: >
            Element path[fill="none"][style="stroke-width: 2.5px"][stroke-dasharray="70.68583470577035"][d="M12.5,1.25A11.25,11.25 0 1 1 1.25,12.5A11.25,11.25 0 0 1 12.5,1.25"] does not have attribute "transform"
        expected: >
            Element path[fill="none"][style="stroke-width: 2.5px"][stroke-dasharray="70.68583470577035"][d="M12.5,1.25A11.25,11.25 0 1 1 1.25,12.5A11.25,11.25 0 0 1 12.5,1.25"] has attribute "transform" with value "rotate(0 12.5 12.5)"
        stack: >
                at DOMAssertions.hasAttribute (http://localhost:7357/assets/test-support.js:9744:20)
                at Object._callee4$ (http://localhost:7357/assets/tests.js:6637:34)
                at tryCatch (http://localhost:7357/assets/vendor.js:406:40)
                at Generator.invoke [as _invoke] (http://localhost:7357/assets/vendor.js:680:22)
                at Generator.prototype.<computed> [as next] (http://localhost:7357/assets/vendor.js:439:21)
                at asyncGeneratorStep (http://localhost:7357/assets/tests.js:6500:105)
        message: >
            rotated halfway
        negative: >
            false
        browser log: |
    ...

> not ok 568 Chrome 78.0 - [70 ms] - Integration | Component | paper-virtual-repeat: should reposition items when scrolled so that there is still only enough items to fit + 3
    ---
        actual: >
            Element .md-virtual-repeat-offsetter has attribute "style" with value "transform: translateY(0px);"
        expected: >
            Element .md-virtual-repeat-offsetter has attribute "style" with value "transform: translateY(30px);"
        stack: >
                at DOMAssertions.hasAttribute (http://localhost:7357/assets/test-support.js:9744:20)
                at http://localhost:7357/assets/tests.js:10910:62
                at invoke (http://localhost:7357/assets/vendor.js:65446:16)
                at Queue.flush (http://localhost:7357/assets/vendor.js:65336:13)
[...]

PR #1113, opened 2019-11-08
Trying to fix the travis build

alex9nadon:
Add waitUntil before trying to get the selector _transform_ for some reasons, the transform is updated after.


alanlalonde:
Fixes #1113

alanlalonde:
Fixes #1106

PR #1118, opened 2019-12-27
update whitelist/blacklist to support pods

betocantu93:
Adds support for blacklist/whitelist with pod components.

Use only whitelist or blacklist, NOT both.

As a reminder, whitelist is an array of components which will be included in the build, as well as all of their dependencies...

i.e
```ts
['ember-paper']: {
  whitelist: ['paper-autcomplete']
}

``` 
Would only include into the build:
```ts
[
      'paper-autocomplete',
      'paper-autocomplete/ebd-trigger',
      'paper-autocomplete/eps-trigger',
      'paper-autocomplete/ebd-content',
      'paper-autocomplete/options',
      'paper-autocomplete/no-matches-message',
      'paper-autocomplete/highlight'
]
```

If nested, the "path" must be specified 
```ts
['ember-paper']: { 
      whitelist: [
        'paper-autocomplete/ebd-trigger'
      ]
}
```

Blacklist is harder to get right, use with caution, as it would only exclude this particular component.
```ts
['ember-paper']: { 
      blacklist: [
        'paper-autocomplete/ebd-trigger'
      ]
}
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -20,21 +20,24 @@ const componentDependencies = {
       'components/autocomplete/autocomplete-theme.scss'
     ],
     dependencies: [
-      'paper-autocomplete-trigger-container',
-      'paper-autocomplete-trigger',
-      'paper-autocomplete-options',
-      'paper-autocomplete-highlight',
-      'paper-autocomplete-content'
+      'paper-autocomplete/ebd-trigger',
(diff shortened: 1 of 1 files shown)


miguelcobain:
Thanks a lot @betocantu93 !

PR #1120, opened 2020-01-13
Embroider compatibility

tempo22:
Hi!

As I was setting up a new app using ember-paper and embroider but i got several error at build time.
This was the necessary fix to make i work.

The 2 changes on comments position were necessary to fix this syntax error: 
https://github.com/glimmerjs/glimmer-vm/blob/6a4c85370d24640f31812036f5e75f8ffff06df7/packages/@glimmer/syntax/lib/parser/handlebars-node-visitors.ts#L216

The deleted file seem to be a forgotten file from a component renaming.


miguelcobain:
Thank you! :+1:

PR #1121, opened 2020-02-02
remove unnecesary "fn" helper

betocantu93:
The new changes to `paper-menu` `paper-select` `paper-chips` and `paper-autocomplete` had unnecessary usage of `fn` helper, the recommended way in octane is just

```ts
import { action } from '@ember/object';
export default class SomeComponent extends Component {
 
  @action
  open() {
    //do stuff that needs this component's context
   this.set('isOpen', true);
  }

 calculatePosition(e) {
    //this function doesn't need component's context
    e.getBoundingClientRect(); 
  }

 @action
  async close(opt) {
   //This function needs component's context and receives an argument
 
    if(opt.id === this.selectedOpt.id) { //whatever
      return await this.onClose();
    }
  }
}
```

```hbs

<button {{on "click" this.open}}>Open Stuff </button>

{{#if this.isOpen}}
   <div {{did-insert this.calculatePosition}} ...attributes>
     <button {{on "click" (fn this.close @opt)}}>Close</button>
   </div>
{{/if}}

```

### DIFF
--- a/addon/components/paper-autocomplete/ebd-content/template.hbs
+++ b/addon/components/paper-autocomplete/ebd-content/template.hbs
@@ -15,7 +15,7 @@
   @otherStyles={{this.customStyles}}
 
   @htmlTag="md-virtual-repeat-container"
-  @shouldReposition={{fn this.shouldReposition}}
+  @shouldReposition={{this.shouldReposition}}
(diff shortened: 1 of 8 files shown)


miguelcobain:
Thank you. This makes sense to me now. 👍

PR #1123, opened 2020-02-13
remove @class

betocantu93:
`...attributes` pass down the class from the invokation


miguelcobain:
I'm keeping the class because I'm aiming for backwards compatibility.
This way `{{#paper-select class="something"}}` would still work.

betocantu93:
@miguelcobain oh I see, talking about backwards compatibility I also think we should use `this.variable` on templates for this matter (even if they are conceptually args), since in 3.16 accessing variables in templates via `@` do mean _arguments_, for me it was a pain to refactor because I had at least a 5 EP components extended which relied in setting defaults (i.e label) if none provided, because of this, I would strongly suggest to make a major release of this changes (when using `@` notation)

PR #1131, opened 2020-02-29
[regression] added child-mixin

betocantu93:
Noticed paper-select doesn't implement child mixin anymore, this is a regression.


miguelcobain:
Thanks @betocantu93 . Good catch. Hoping we can get rid of the Mixins soon, but meanwhile we definitely don't want to break things.

Issue #1134, opened 2020-03-16
Getting mouseEnter/Leave/Move event handler deprecation logs


Shajansheriff:
In ember-cli 3.15, getting deprecation logs when i use paper-input.

As per [ember deprecations doc](https://deprecations.emberjs.com/v3.x/#toc_component-mouseenter-leave-move ) mouse events are deprecated.

PR #1135, opened 2020-03-16
Fix[1134] mouseEnter/Leave/Move deprecations by adding event listeners

Shajansheriff:
Fix [1134](https://github.com/miguelcobain/ember-paper/issues/1134)

### DIFF
--- a/addon/components/paper-input.js
+++ b/addon/components/paper-input.js
@@ -4,7 +4,7 @@
 import { or, bool, and } from '@ember/object/computed';
 
 import Component from '@ember/component';
-import { computed } from '@ember/object';
+import { computed, set } from '@ember/object';
 import { isEmpty } from '@ember/utils';
 import { run } from '@ember/runloop';
 import { assert } from '@ember/debug';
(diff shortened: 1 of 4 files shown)


betocantu93:
interesting, I didn't realized you could write native class for mixins!

Shajansheriff:
@betocantu93 We have to replace mixins with native class or with some other patterns. Mixins are going to be removed, however we need to have a roadmap for refactor, because we are using quite a handful of mixins in our addon.

Shajansheriff:
@miguelcobain Do you have any idea why those two specific tests are failing? It is passing in my local.

miguelcobain:
@Shajansheriff interesting workaround for the deprecation. That didn't cross my mind.
The failing tests are unrelated, as you probably guess.
Those tests are flaky, and I suspect are related to using ember vertical collection. Usually they pass after retrying them some times.
Since it's related I'll merge this even if they don't pass.

alex9nadon:
@miguelcobain  Hi, did you plan to merge and create a new release?

PR #882, opened 2018-01-17
mouseUp --> touchEnd for paper-reset-button

betocantu93:
mouseUp don't work in mobile

PR #1137, opened 2020-03-20
allowclear for mobile device

alex9nadon:


betocantu93:
Indeed, it doesn't work, I think its the same as #882

PR #1141, opened 2020-03-25
Update paper-button to native class syntax and move it to pod structure.

Shajansheriff:
Refactor focusin & focusout at focusablemixin to event listeners.


Shajansheriff:
@miguelcobain Do let me know this approach is good enough to gradually migrate other components as well. Or shall we refactor them to tagless component and move all the decorators bindings to template? Even though there are ember-decorators to support native class syntax for ember-component, Ember is strongly advising to start using tagless component and modifiers in template for event handling. What is your suggestion?

I also see a small problem in migrating paper-button to tagless, there is href logic to create button as anchor tag in init hook, which might be an obstacle. Any workaround?

miguelcobain:
@Shajansheriff indeed, the idea is to use tagless components (with the decorators `@tagName('')`.

You bring up a good question regarding the href logic in paper-button.
The way to solve that is to use the [ember-element-helper](https://github.com/tildeio/ember-element-helper) addon like:

```hbs
{{#let (element (if @href "a" "button")) as |Tag|}}
  <Tag ...attributes>{{yield}}</Tag>
{{/let}}
```

Don't forget to include the ember-element-helper in the dependencies and not devDependencies of `package.json`.

Doing this will force you to migrate away from `ColorMixin` and `FocusableMixin`.
Those two should be easy to refactor using normal templating.

The `ProxiableMixin` however might require a bit more testing, but I think shouldn't be too hard as it is only used on the `<PaperItem>` component.

Let me know if you need any further clarification, and thank you for your work!

PR #1143, opened 2020-04-02
Fix [1142] remaining mouseEnter/mouseLeave deprecations by adding eve…

fpauser:
…nt listeners

### DIFF
--- a/addon/components/paper-item.js
+++ b/addon/components/paper-item.js
@@ -17,6 +17,9 @@ export default Component.extend(ParentMixin, {
   layout,
   tagName: 'md-list-item',
 
+  _mouseEnterHandler: undefined,
+  _mouseLeaveHandler: undefined,
+
   // Ripple Overrides
   // disable ripple when we have a primary action or when we don't have a proxied component
   noink: computed('hasPrimaryAction', 'hasProxiedComponent', function() {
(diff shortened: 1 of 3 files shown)


addon/components/paper-item.js
```diff
@@ -49,6 +52,26 @@ export default Component.extend(ParentMixin, {
...
+  },
+
+  willDestroyElement() {
+    this._super(...arguments);
+
+    this.element.removeEventListener('mouseenter', this._mouseEnterHandler);
+    this.element.removeEventListener('mouseleave', this._mouseLeaveHandler);
+
+    this._mouseEnterHandler = undefined;
+    this._mouseLeaveHandler = undefined;
```
Shajansheriff, line 72: @fpauser Why do we have to define a handler property to store the action and then clear it in will destroy ? Instead we can directly assign the actions to event listener and remove it in will destroy hook.

fpauser: Its needed because `bind()` returns a new function bound to `this` with each call. To have `removeEventListener` remove our handler we have to give it the exact same handler function that was given to `addEventHandler`.

```js
this.handleMouseEnter.bind(this) !== this.handleMouseEnter.bind(this)
```

fpauser: Otherwise we'd produce leaks.

Shajansheriff: That is a good insight. Thanks

fpauser:
Can we have a fresh release with the fix?

miguelcobain:
Thanks, @fpauser!
A new release is coming shortly.

PR #1144, opened 2020-04-29
fix paper menu positioning

### LINKED ISSUE #1140: Regression: paper-menu always goes down, even if it's near the bottom of the screen
On beta.23 and earlier, `paper-menu`s near the bottom of the page looked like this:
Screen Shot 2020-03-23 at 4 28 35 PM

On beta.27 and later, it looked like this:
Screen Shot 2020-03-23 at 4 31 42 PM

I wasn't able to test betas 24-26 due to #1119, but beta 27's menu refactor seems to be the most likely culprit anyway.

blmundie:
closes #1140

### DIFF
--- a/addon/components/paper-menu/component.js
+++ b/addon/components/paper-menu/component.js
@@ -1,21 +1,8 @@
+import { layout, tagName } from '@ember-decorators/component';
 import Component from '@ember/component';
+import { action, computed } from '@ember/object';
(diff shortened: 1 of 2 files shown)


miguelcobain:
@blmundie thanks for this. Indeed, this was a regression that was introduced. Thanks for taking the time to find it and cleaning things up.

PR #1147, opened 2020-05-11
Use click instead of mouseUp to clear autocomplete (allowClear=true)

### LINKED ISSUE #1138: autocomplete allowclean does not work on mobile device
mouseUp does not work on mobile device

alex9nadon:
Click works on both mobile and desktop.

Fixes #1138 & #882


alex9nadon:
MouseUp and TouchEnd allow users to changes his mind by releasing out of the reset button.
But it, dose'nt work's on both mobile & desktop.
I ended up using click instead, at lease it work's on both.

miguelcobain:
Thank you @alex9nadon

Issue #1158, opened 2020-07-18
Paper Select not showing error message


wukongrita:
Hi @miguelcobain 
Why is the paper select not showing the error message? It just display the warning colour...

![image](https://user-images.githubusercontent.com/25361005/87863868-9f175e00-c98a-11ea-9487-d5f478640f63.png)

Ember Paper version 1.0.0-beta.31

Works when using the 1.0.0-beta.26 version...

wukongrita:
I added these lines at https://github.com/miguelcobain/ember-paper/blob/master/addon/components/paper-select/template.hbs#L35

```
  {{#unless hideAllMessages}}
    <div class="md-errors-spacer"></div>

    {{#if isInvalidAndTouched}}
      <div class="md-input-messages-animation md-auto-hide">
        <div class="paper-input-error ng-enter ng-enter-active md-input-message-animation" style="opacity: 1; margin-top: 0">
          {{@errors}}
        </div>
      </div>
    {{/if}}
  {{/unless}}
```

Seems there is missing on displaying error messages for paper select...

PR #1164, opened 2020-09-08
Fix PaperSelect error message

wukongrita:
Hi @miguelcobain ,
This will add the error message to show on PaperSelect component ( issue #1158 )
So the error message will appear at the bottom of te PaperSelect, like so:
image


miguelcobain:
Thank you, @wukongrita.
The problem with this is that it just renders the passed in errors.

You should use `this.validationErrorMessages` instead, that contains the manually passed in errors as well as other custom validation errors.
This was indeed a regression. Here is how it looked like before the refactor:
https://github.com/miguelcobain/ember-paper/blob/v1.0.0-beta.1/addon/templates/components/paper-select.hbs#L51-L67

Can you please update this PR?

wukongrita:
Something like adding:
```
.
.
.
    @noMatchesMessageComponent="paper-select/no-matches-message"
    @searchMessageComponent="paper-select/search-message"
    @onClose={{this.close}}
    @onOpen={{this.open}}
    @onFocus={{this.focus}}
    @onBlur={{this.blur}}
    as |opt term|>
    {{yield opt term}}
  </PowerSelect>

  {{#unless hideAllMessages}}
    <div class="md-errors-spacer">
      {{#if maxlength}}
        <div class="md-char-counter">{{renderCharCount}}</div>
      {{/if}}
    </div>
    {{#if isInvalidAndTouched}}
      <div class="md-input-messages-animation md-auto-hide" ng-messages>
        {{#each validationErrorMessages as |error index|}}
          <div id="error-{{inputElementId}}-{{index}}"
          class="paper-input-error ng-enter ng-enter-active" ng-message>
            {{error.message}}
          </div>
        {{/each}}
      </div>
    {{/if}}
  {{/unless}}

</md-input-container>
```

at the `addon/components/paper-select/template.hbs` ?

wukongrita:
> Thank you, @wukongrita.
> The problem with this is that it just renders the passed in errors.
> 
> You should use `this.validationErrorMessages` instead, that contains the manually passed in errors as well as other custom validation errors.
> This was indeed a regression. Here is how it looked like before the refactor:
> https://github.com/miguelcobain/ember-paper/blob/v1.0.0-beta.1/addon/templates/components/paper-select.hbs#L51-L67
> 
> Can you please update this PR?

@miguelcobain , is the updated PR ( https://github.com/miguelcobain/ember-paper/pull/1164/commits/dd42c15a21e6acc4ca06dbccf5b742d7bba538ae ) is what you meant?

wukongrita:
@miguelcobain any other things that missing from the updated PR?

addon/components/paper-select/template.hbs
```diff
@@ -32,4 +32,21 @@
...
   </PowerSelect>
 
-</md-input-container>
\ No newline at end of file
+  {{#unless this.hideAllMessages}}
+    <div class="md-errors-spacer" id={{concat this.inputElementId "-char-count"}}>
+      {{#if @maxlength}}
+        <div class="md-char-counter">{{this.renderCharCount}}</div>
+      {{/if}}
+    </div>
```
miguelcobain, line 40: I just realised that we can remove this whole char count div.
PaperSelect is not a control you can type characters in and `renderCharCount` is not even defined.

wukongrita:
@miguelcobain done, removed the charCount div section...

Issue #1166, opened 2020-09-22
Paper select errors when menu is closed on ember > 3.17


LucasHill:
When you open a <PaperSelect ... /> and then close the menu, you'll see this error show up in the console.

`
component.js:65 Uncaught (in promise) TypeError: Cannot read property 'appendChild' of null
    at PaperMenuContent.animateOut (component.js:65)
    at Object.destroyModifier (will-destroy.js:68)
    at CustomModifierState.destroy (index.js:6906)
    at StringDestroyableDestructor.[DID_DROP] (util.js:290)
    at util.js:201
    at Set.forEach (<anonymous>)
    at didDestroyAssociated (util.js:200)
    at SimpleDestructor.[DID_DROP] (util.js:314)
    at util.js:201
    at Set.forEach (<anonymous>)
`

The issue is from ember-power-select and you can see the relevant issues at:
https://github.com/cibernox/ember-power-select/issues/1338 (fixed here)
https://github.com/cibernox/ember-basic-dropdown/issues/540

The fix was released in ember-power-select 4.0.

I was able to verify the issues goes away by forcing ember-power-select to resolve to 4.0.4.

I see that ember-power-select requires ember 3.13 which is a year old, but this seems worth it as we would be totally blocked from upgrading without forcing the resolution.

hergaiety:
Can confirm we're seeing this too trying to upgrade to Ember 3.17.x

```
        stack: >
            TypeError: Cannot read property 'appendChild' of null
                at PaperSelectEbdContent._callee$ (http://localhost:7357/assets/vendor.js:209457:33)
                at tryCatch (http://localhost:7357/assets/vendor.js:5014:40)
                at Generator.invoke [as _invoke] (http://localhost:7357/assets/vendor.js:5288:22)
                at Generator.prototype.<computed> [as next] (http://localhost:7357/assets/vendor.js:5047:21)
                at asyncGeneratorStep (http://localhost:7357/assets/vendor.js:209370:105)
                at _next (http://localhost:7357/assets/vendor.js:209372:196)
                at http://localhost:7357/assets/vendor.js:209372:366
                at new Promise (<anonymous>)
                at PaperSelectEbdContent.<anonymous> (http://localhost:7357/assets/vendor.js:209372:99)
                at PaperSelectEbdContent.animateOut (http://localhost:7357/assets/vendor.js:209483:30)
        message: >
            Cannot read property 'appendChild' of null
```

LucasHill:
[...]

Issue #1151, opened 2020-05-27
"Cannot read property 'appendChild' of null" when running integration test with PaperMenu


calvinps:
I have a component that renders a PaperMenu component. After running an integration test for my component in which I render the PaperMenu component and click on the trigger, my test suite fails. The source of the issue is that the `animateOut` function in PaperMenu::Content the element doesn't have a "parentElement" and the function crashes at the end of the test. I've tried unsuccessfully to find a way to destroy the PaperMenu gracefully. 

Ember version: 3.18
Ember Paper version: 1.0.0-beta.30

My component:
```
<PaperMenu @isOpen={{this.isOpen}} @position="target-right target" as |menu|>
  <menu.trigger>
    <PaperButton @class="menu-btn" @onClick={{null}} @iconButton={{true}}>
      <ProfilePicture @user={{this.user}} @class="user-icon" @height={{42}} />
    </PaperButton>
  </menu.trigger>
  <menu.content as |content|>
    {{#each @items as |item|}}
      <content.menu-item @onClick={{action "selected" item}}>
        <p class="tool-bar-menu-item-title">
          {{item.title}}
        </p>
        {{#if item.icon}}
          <PaperIcon @icon={{item.icon}} @class="md-menu-align-target" />
        {{/if}}
      </content.menu-item>
    {{/each}}
  </menu.content>
</PaperMenu>

```

My test:

```
import { module, test } from "qunit";
import { setupRenderingTest } from "ember-qunit";
import {
  click,
  findAll,
  render,
  settled
} from "@ember/test-helpers";
import hbs from "htmlbars-inline-precompile";

module("Integration | Component | tool-bar/menu", function (hooks) {
  setupRenderingTest(hooks);

  hooks.beforeEach(function () {
    this.user = this.owner.lookup("service:store").createRecord("user", {
      profilePicture:
        "https://xxxxx.s3.amazonaws.com/media/190304/conversions/blob-thumb.jpg",
      fullName: "John Doe",
    });
  });

  test("it renders", async function (assert) {
    this.set("items", [
      {
        title: "Home",
        transitionTo: "get-started",
      },
      {
        title: "Settings",
        transitionTo: "settings",
      },
      {
        title: "Logout",
        transitionTo: "logout",
      },
    ]);

    await render(
      hbs`<ToolBar::Menu @items={{this.items}} @user={{this.user}} />`
    );

    assert.equal(
      this.element.querySelector("img").getAttribute("src"),
[...]

Issue #1139, opened 2020-03-21
paper-select : parentElement is null error after using


larryh:
Hi,

First of all: thanks for all of the work you folks do on Ember Paper!

I just upgraded ember from 3.15 to 3.17, and ember-paper from .beta.26 to .beta.29 and am having a problem with ember-select.

The drop-down has a list of states, and when you select a state you get a list of the politicians in that state. It starts out ok, and when you select a state you get the correct list of pols, but in so doing you get a `parentElement is null` error (which you can see in the console) and the app's entire UI is frozen.

Also note that in the last screenshot, the selected state, e.g. Massachusetts, does not appear in the select-box, instead the "Select A State" placeholder has "moved up" above the control and stayed there.

I have attached screenshots of the process (in order, i.e. 1_xxx, 2_yyy, 3_zzz). Here is the code that I use, which has worked for a long time - even with the new "Octane Way" of doing things.

Can someone shed some light on this? Obviously I had to back down to .beta.26 so I can use the app. 

Actually, I guess I should try moving to .27 and then .28 to see exactly where this happens. I will do that and update this issue with my findings. In the meantime, here is the code that I use (it's a component):

`  <PaperSelect 
data-test-select-pols-state
@label="State"
@placeholder="Select a State"
@options={{this.states}}
@selected={{this.selectedState}}
@onChange={{fn this.filter "state"}}
 class="column-control" as |state| >
    {{state.name}} 
<\/PaperSelect>

`

![1_InitialState](https://user-images.githubusercontent.com/89646/77232472-1e2c8f00-6b6f-11ea-9158-7dfc7c4283d5.png)
![2_States](https://user-images.githubusercontent.com/89646/77232473-1ec52580-6b6f-11ea-9012-e007fbe1b418.png)
![3_AfterSelection](https://user-images.githubusercontent.com/89646/77232474-1ec52580-6b6f-11ea-82aa-a61b721fef5a.png)

larryh:
Hello again,

Turns out the problem was introduced in .beta.27. Hope this helps.

Gaelan:
Running into this with `paper-menu` as well.

ctwoolsey:
[...]

PR #1175, opened 2020-11-13
Better handling basic-dropdown parent for testing env (#1166)

romgere:
This try to fix `Cannot read property 'appendChild' of null` issue on `paper-menu` & `paper-select` for ember > 3.17 when testing.

I think related issues are : 
- https://github.com/miguelcobain/ember-paper/issues/1139
- https://github.com/miguelcobain/ember-paper/issues/1151
- https://github.com/miguelcobain/ember-paper/issues/1166

The code of https://github.com/miguelcobain/ember-paper/blob/b7f13ad6137e456c53afa1bfda8ede0855e1a2d1/addon/utils/ebd-get-parent.js is inspired by what `ember-basic-drop-down` do : https://github.com/cibernox/ember-basic-dropdown/blob/850c227c0a58148056d55d41aa0e5d88656b8165/addon/components/basic-dropdown.js#L273-L290

In addition to this PR, maybe I miss something, but : 

From what I understand here, the reason `animateOut` exits on `paper-select` & `paper-menu` is for animation & to handle `isActive` :thinking: .

Plus, on paper-select, to keep the scroll position into the drop-down on animate out... (https://github.com/miguelcobain/ember-paper/blob/master/addon/components/paper-select/ebd-content/component.js#L71)

So we duplicate what `ember-basic-dropdown` already does (https://github.com/cibernox/ember-basic-dropdown/blob/v2.0.15/addon/components/basic-dropdown-content.js#L172-L184) "just" to handle `isActive` & keep scroll... So we have to fix on `ember-paper` side an issue already fixed / wich does not exist in `ember-basic-dropdown`, right ?

Wouldn't it be easier to only handle `isActive` by our own in `paper-select` & `paper-menu` with `did-insert` & `will-destroy`, as it is already done, make the scroll behavior "native" in `ember-basic-dropdown` (seems legit) & rely on `ember-basic-dropdown` for animation part :thinking:

### DIFF
--- a/addon/components/paper-menu/content/component.js
+++ b/addon/components/paper-menu/content/component.js
@@ -8,6 +8,9 @@ import { tagName, layout } from '@ember-decorators/component';
 
 import { ESCAPE, LEFT_ARROW, UP_ARROW, RIGHT_ARROW, DOWN_ARROW } from 'ember-paper/utils/key-constants';
 
+import { getOwner } from '@ember/application';
+import ebdGetParent from 'ember-paper/utils/ebd-get-parent';
+
 function waitForAnimations(element, callback) {
   let computedStyle = window.getComputedStyle(element);
   if (computedStyle.transitionDuration && computedStyle.transitionDuration !== '0s') {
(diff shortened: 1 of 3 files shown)


romgere:
> rely on ember-basic-dropdown for animation part
I made some tests & it seems more complicated than expected... `ember-basic-dropdown` rely on CSS `animation` but actually material use CSS `transition` for select & menu animation :disappointed: 

I have 0 XP in css `transition`/`animation` & don't know if we can override material css for this part or even if it worth it :man_shrugging: 

Maybe someone more confident with ember-paper / EBD & css can give some pros/cons about this.

I stop this PR here & wait for some feedback.

PR #1176, opened 2020-11-28
Patching ebd-content to workaround issue 1166 

adamjson64:
This same patch was applied in paper-menu. I am on Ember 3.22 and this was the only way to get the paper-select component to work.


LucasHillDex:
@miguelcobain could we get this in if it's acceptable? I think lots of people are blocked on upgrading ember if using emper-paper.

LucasHillDex:
Does ember-paper need to get adopted? For such a widely used addon it feels like its dying a bit.

PR #1179, opened 2021-02-09
Update ember-basic-dropdown to remove deprecation warning

ohcibi:
The warning was caused by an outdated version of ember-maybe-in-element.


bjornharrtell:
Another reason to merge this is that ember-basic-dropdown 2.x is broken on ember-source 3.25.0 and forward. See https://github.com/cibernox/ember-basic-dropdown/issues/578#issuecomment-806063900.

bertdeblock:
`ember-basic-dropdown` should remain under `dependencies` though? [It's used directly here](https://github.com/miguelcobain/ember-paper/blob/master/addon/components/paper-menu/template.hbs#L1).

bjornharrtell:
@bertdeblock I would think it should work and is more appropriate as a dev dependency? (Used in build not directly in runtime)

bertdeblock:
Don't think so? If it would work listed under `devDependencies` that would probably be because `ember-paper` also depends on `ember-power-select`, which in turn depends on `ember-basic-dropdown` (and there it's listed under normal `dependencies` as well). `ember-paper` uses the `<BasicDropdown>` component directly so keeping it under `dependencies` is the right thing.

bertdeblock:
I think you are confusing it with listing dependencies under `devDependencies` for a _project_?

bjornharrtell:
@bertdeblock ah yes you are of course correct, my mistake. And I think it does not make much sense to update only one of these related components. I've started on https://github.com/miguelcobain/ember-paper/pull/1182 but it's not a trivial upgrade.

bjornharrtell:
Discussed with @miguelcobain and since ember-paper is essentially in maintenance only mode it is better to solve this issue with a patch for ember-basic-dropdown v2 branch instead of attemping a large refactoring in ember-paper.

PR #1217, opened 2022-03-05
remove Ember 3.28 from ember-try scenarios

mansona:
Hey I just noticed that Ember 3.28 is showing up on the ember-try scenarios (but I didn't see a PR that added that 🤔)

Just a minor thing: we probably don't need that since the current version of Ember in the repo is 3.28 so the normal `Tests` will be using that version.

Happy to leave it there too but it's just a tiny bit of an efficiency to remove it 👍

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,7 +56,6 @@ jobs:
         try-scenario:
           - ember-lts-3.20
           - ember-lts-3.24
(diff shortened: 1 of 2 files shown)


Subtletree:
Thanks @mansona my mistake.

> but I didn't see a PR that added that 🤔

In future I will create a PR for you to review.

PR #1240, opened 2023-05-29
fix: Ember is not defined error

wozny1989:
# Tasks:
 - [x] Fix error `Ember is not defined` - import Ember during library version register

### DIFF
--- a/index.js
+++ b/index.js
@@ -2,7 +2,6 @@
 
 const path = require('path');
 const resolve = require('resolve');
-const version = require('./package.json').version;
 const { MergeTrees } = require('broccoli-merge-trees');
 const writeFile = require('broccoli-file-creator');
 const Funnel = require('broccoli-funnel');
@@ -407,7 +406,6 @@ module.exports = {
 
     this.emberPaperOptions = Object.assign({}, app.options['ember-paper']);
 
-    app.import('vendor/ember-paper/register-version.js');
     app.import('vendor/hammerjs/hammer.js');
     app.import('vendor/propagating-hammerjs/propagating.js');
   },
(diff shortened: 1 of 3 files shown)


Rxbsxn (approved):

index.js
```diff
@@ -464,7 +464,7 @@ module.exports = {
 
     let versionTree = writeFile(
       'ember-paper/register-version.js',
-      `Ember.libraries.register('Ember Paper', '${version}');`
+      `const Ember = require('ember').default;Ember.libraries.register('Ember Paper', '${version}');`
```
mansona, line 467: I think I would prefer just to remove this functionality 🤔 I don't think we need to define the Ember-Paper version in the client side for any good reason

package.json
```diff
@@ -13,13 +13,14 @@
     "start": "ember serve",
     "test": "npm-run-all lint test:*",
     "test:ember": "ember test",
-    "test:ember-compatibility": "ember try:each"
+    "test:ember-compatibility": "ember try:each",
+    "postinstall": "patch-package"
```
mansona, line 17: we really can't be adding a patch-package post install to this 🙈

package.json
```diff
@@ -42,14 +43,15 @@
...
+    "patch-package": "^6.5.1",
     "polyfill-nodelist-foreach": "^1.1.2",
     "propagating-hammerjs": "^2.0.1",
     "resolve": "^1.13.1",
     "sass": "^1.23.7",
     "tinycolor2": "^1.4.1"
   },
   "devDependencies": {
-    "@ember/jquery": "^1.1.0",
+    "@ember/jquery": "2.0.0",
```
mansona, line 54: we should probably be removing `@ember/jquery`. Is there anything useful that it's actually being used for?

mansona (changes requested):
So I don't think we should probably do these changes 🙈 I understand that you're trying to fix something here but I would rather a) just plainly remove the functionality b) fix it upstream or c) upgrade to more standard ways of doing it.

I hope all that makes sense?

wozny1989:
@mansona You're right, I just removed functionality with register version if is no super required 👍

mansona (approved):
looks great 🎉

PR #1261, opened 2024-10-15
build(deps): upgrades to `ember-composability-tools@~1.0.3`.

matthewhartstonge:
Helps to removes some deprecations

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -23,7 +23,7 @@
         "ember-cli-htmlbars": "^5.7.1",
         "ember-cli-polyfill-importer": "^0.0.4",
         "ember-cli-sass": "^10.0.1",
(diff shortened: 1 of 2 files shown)


matthewhartstonge:
Ah, interesting...

We can't just bump our way to victory? 1.1 jumps to ember 4 support.

What issues does it cause?

matthewhartstonge:
As an FYI I'm starting to hammer things into a fork just in the need to move fast as conflicts are and will become a bane, so I'm not sure how things are going to go in terms of divergence from upstream.

Do you have any suggestions? Dependencies will obviously be the worst.

I'm trying my best to keep things from not diverging as much as possible.

Maybe atomic PRs that merge into one another? Enables me to go back and fix things as I go?

PR #1290, opened 2024-10-29
ci: adds no deprecations workflow

matthewhartstonge:
TSIA!

Removes deprecations from the dummy app where it was directly accessing the application controller router:
- https://deprecations.emberjs.com/id/application-controller-router-properties

### DIFF
--- a/config/ember-try.js
+++ b/config/ember-try.js
@@ -88,6 +88,15 @@ module.exports = async function () {
           },
         },
       },
+      {
+        name: 'ember-release-no-deprecations',
+        npm: {
+          devDependencies: {
+            'ember-source': await getChannelURL('release'),
+            'ember-deprecation-error': '*',
+          },
+        },
+      },
     ],
   };
 };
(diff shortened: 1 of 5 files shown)


SergeAstapov (approved):

config/ember-try.js
```diff
@@ -80,6 +80,14 @@ module.exports = async function () {
       },
       embroiderSafe(),
       embroiderOptimized(),
+      {
+        name: 'no-deprecations',
```
mansona, line 84: so this should be added to the github action workflow (probably in the allowed to fail section 😂) so we actually run it with CI

Also I tend to also add an ember-release-no-deprecations one too that lets us know if we're getting new deprecations that we would need to deal with https://github.com/mansona/ember-cli-notifications/blob/main/test-app/config/ember-try.js#L133-L141 but that's probably getting ahead of ourselves 🙈

matthewhartstonge: Have added 👍

PR #1295, opened 2024-10-31
feat: remove mixins [wip]

matthewhartstonge:
experimenting with paper-button

### DIFF
--- a/addon/components/-focusable.js
+++ b/addon/components/-focusable.js
@@ -1,0 +1,174 @@
+/**
+ * @module ember-paper
+ */
+import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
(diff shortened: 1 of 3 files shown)


addon/components/-focusable.js
```diff
@@ -0,0 +1,175 @@
...
+
+  toggle = false;
+
+  // Only render the "focused" state if the element gains focus due to
+  // keyboard navigation.
+  get focusOnlyOnKey() {
+    return this.args.focusOnlyOnKey ? this.args.focusOnlyOnKey : false;
+  }
+
+  @action registerListeners(element) {
```
matthewhartstonge, line 45: this _feels_ like an insane amount of listeners... I think a lot of it has to do with animating ripple effects

PR #1296, opened 2024-11-01
Add percy

mansona:


mansona:
@matthewhartstonge I would like to get your opinion on this one. I noticed when I added percy that it wasn't a full-screen snapshot. It turns out that a few decisions were made in the past that prevented that. 

This PR fixes it but it is actually a breaking change 🙈 I'm wondering if you have any thoughts on what we should do here?

matthewhartstonge (commented):
I've run a test with percy @ https://github.com/adopted-ember-addons/ember-paper/pull/1297

I've simplified it to only lock if in the dummy app on the catalog rather than introduce the breaking change.

Does that satisfy backwards compat?

PR #1297, opened 2024-11-05
percy: only lock in the dummy app on the catalog route.

matthewhartstonge:
Refer: [#1296 ](https://github.com/adopted-ember-addons/ember-paper/pull/1296#pullrequestreview-2414336136)


matthewhartstonge:
closed in favour of #1296

PR #1299, opened 2024-11-08
feat(addon/components): converts paper-grid to glimmer components.

matthewhartstonge:
Note: Instead of using ember runloop, this does rely on `requestAnimationFrame` for debouncing. (breaking change?)


addon/components/paper-grid-list.js
```diff
@@ -171,26 +252,28 @@ export default class PaperGridList extends Component.extend(ParentMixin) {
 
   // Calculates tile positions
   _setTileLayout() {
-    let tiles = this.orderedTiles();
+    let tiles = this.orderedTiles;
     let layoutInfo = gridLayout(this.currentCols, tiles);
 
-    tiles.forEach((tile, i) => tile.set('position', layoutInfo.positions[i]));
+    tiles.forEach((tile, i) => {
+      tile.position = layoutInfo.positions[i];
```
matthewhartstonge, line 259: This feels bad and I feel like we should invert the control here, i.e. the child provides an update position func that can be called by the parent

hero-davidm (approved):
all looks to be in order 👍

PR #1304, opened 2024-11-13
feat(addon/components/paper-item): converts to a glimmer component.

matthewhartstonge:
note: two tests fail:
- `Integration | Component | paper item: single action checkboxes should react to checkbox clicks`
- `Integration | Component | paper item: Item checkbox with secondary action and no primary action is toggled by checkbox click`

This appears due to to `click` only bubbling once under classic components, whereas the proxy call is being made under the glimmer parent, then the click handler is being called on the checkbox classic component.

Hoping this will come out in the wash once paper-checkbox is migrated to glimmer.


matthewhartstonge:
Failing tests fixed by #1305!

hero-davidm (approved):
LGTM

PR #1312, opened 2024-11-26
feat(addon/components/paper-input): converts to a glimmer component.

matthewhartstonge:
This was fun...

I've opted to migrate from the ValidationMixin by using an injectable `Validation` class.
There are a couple more components that require this migration (paper-autocomplete and paper-select) so I'll make changes as required, or maybe make unique ones if required...

I'm pretty sure all places work on validating an input component, so we should be good 🎉

(The Validator now allows me to go back and fix the failing paper-form tests once migrating the other two dependants)

### DIFF
--- a/addon/components/paper-input.hbs
+++ b/addon/components/paper-input.hbs
@@ -1,3 +1,4 @@
+{{! template-lint-disable no-down-event-binding no-positive-tabindex }}
 <md-input-container
   class='md-default-theme
     {{~if this.hasValue " md-input-has-value"}}


hero-davidm (approved):

matthewhartstonge:
Theres a cyclic dependency between `paper-input` #1312, `paper-select` #1314 and `paper-form` #1300 in relation to tests.

These tests all come out in the wash once `paper-form` is upgraded to the new style of tracking/adopting child components rounding out the removal of `ember-composable-components`.

PR #1314, opened 2024-12-09
feat(addon/components/paper-select-*): converts to glimmer components.

matthewhartstonge:


matthewhartstonge:
Theres a cyclic dependency between `paper-input` #1312, `paper-select` #1314 and `paper-form` #1300 in relation to tests.

These tests all come out in the wash once `paper-form` is upgraded to the new style of tracking/adopting child components rounding out the removal of `ember-composable-components`.

